PR #1, opened 2019-11-13
Euro-calliope inflow and README updates

brynpickering:
Two updates in here:

1. More explicit directions to data in the README
2. A method to deal with spotty hydro inflow data

In the second case, it's clear from the data that instances of zero inflow are erroneous (when the data in timesteps either side is >x10^4). I've dealt with this by rescaling capacities based on the percentage of timesteps in which inflow is non-zero, when calculating the capacity share. It's possibly a bit of a hack, but no more so than the existing method ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,11 +19,11 @@ conda activate euro-calliope
 
 4. Further, you need all data files that cannot be retrieved automatically:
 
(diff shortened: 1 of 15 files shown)


brynpickering:
Closing in favour of two separate PRs

PR #4, opened 2019-11-18
git: to https:

brynpickering:
Does what it says on the tin. GitHub's [current recommendation](https://help.github.com/en/github/using-git/which-remote-url-should-i-use) is to use the 'https' URL for cloning repositories. But, besides that, it's a lot easier to work with on a cluster compute node. using the 'git:' URL certainly didn't work for me (either a credential issue, or some blocking of SSH on those nodes).


timtroendle:
I never had an issue there, but if this makes it more broadly compatible, that should be good. Thanks!

PR #5, opened 2019-11-21
update README to clarify download locations

brynpickering:
Update README to be more explicit about download locations


timtroendle:
Thanks!

PR #7, opened 2019-11-22
Update hydro data path

brynpickering:
It takes a a reasonable amount of time to get the hydro runoff data from Copernicus, so it might be better to store annual data locally, such that switching between years and rerunning the snakemake workflow doesn't require the data to be freshly downloaded. This also deals with the current issue that running for a different year requires manually deleting the 'europe-cutout' directory, since it's protected.


timtroendle:
Yes that seems reasonable, thanks!

PR #27, opened 2020-06-08
Update to hydro power database release 05

FebinKa:


timtroendle:
Thanks for the commit. Have a look at https://github.com/timtroendle/euro-calliope/commit/6ab001a49e154f0575dcc6f9c59e7bf69b6097db where I updated from version 03 to 04. You also need to update line 38. Further, we should run the workflow at least once as a test to see if things work. If you can't, let me know and I'll give it a try.

FebinKa:
Thanks for the comment, I've updated the code but I can't run the workflow.

timtroendle:
I'm very sorry for stalling this @FebinKa. I am closing this PR as the current version is v07 and it will be introduced in #43.

PR #40, opened 2020-11-03
Update IRENA hydro generation data

timtroendle:
This adds most recent hydro generation data of IRENA.


brynpickering:
I'm sorry to have missed this PR @timtroendle. Are the changes still valid?

timtroendle:
No problem. Yes, they are still valid.

notebooks/preprocess-irena-data.ipynb
```diff
@@ -0,0 +1,139 @@
+{
```
brynpickering, line 1: Why can this not be done inside the workflow? It accesses an Excel and processes it, which seems like something that doesn't need to be separate. It just requires the Excel to be pre-downloaded by the user into `data`

notebooks/preprocess-irena-data.ipynb
```diff
@@ -0,0 +1,139 @@
...
+    "        return None"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "COUNTRY_IDS = [\n",
```
brynpickering, line 47: Should it actively access the config and get the scope of countries and work from that? Then you know expplicitly if you have data gaps where certain countries are missing

notebooks/preprocess-irena-data.ipynb
```diff
@@ -0,0 +1,139 @@
...
+    "df = (\n",
+    "    pd\n",
+    "    .read_excel(\"../data/IRENA_RE_Statistics_July2020extract.xlsx\", index_col=[0, 1, 2])\n",
+    "    .rename_axis(index=[\"country_code\", \"indicator\", \"technology\"])\n",
+    "    .rename_axis(columns=\"year\")\n",
+    "    .stack()\n",
+    "    .unstack(\"indicator\")\n",
+    "    .rename(index=country_id, level=\"country_code\")\n",
+    "    .reset_index()\n",
+    "    .dropna(\"index\", subset=[\"country_code\"])\n",
```
brynpickering, line 87: You really just want to drop any NaNs, so you could avoid resetting the index and dropping and just do `dropna()` on the df?

src/hydro/inflow_mwh.py
```diff
@@ -14,12 +14,16 @@ def determine_energy_inflow(path_to_stations_with_water_inflow, path_to_generati
...
-        index_col=[0, 1],
-        names=["country_code", "year", "generation"],
-        header=0
-    )["generation"].to_xarray().sel(year=year).to_series() * 1000 # from GWh to MWh
+    return (
+        pd
+        .read_csv(path_to_generation, index_col=[0, 1])
+        .loc[:, "generation_gwh"]
+        .rename("generation")
+        .to_xarray()
```
brynpickering, line 22: do you really need to convert to xarray to get this information (same question for the notebook)? Use `.xs(year, level='year')` to get the relevant year from the MultiIndex.

It would also seem worthwhile to keep the data units in the name moving forward (now  'generation_mwh' )

brynpickering (changes requested):
The code snippet in the notebook is so short that I would remove the notebook and process the excel directly in the workflow. The excel is in 'data' either way. The code can be simplified by keeping the data as a MultiIndex series the whole time and using `xs(...)` when accessing single index level items (even across several index levels) and `.loc[pd.IndexSlice[...]]` when accessing multiple index level items.

notebooks/preprocess-irena-data.ipynb
timtroendle: Yes, but there are two reasons against doing that. First, I don't want another manual download here. We should get rid of manual downloads instead of introducing more. Second, this data is not downloadable from their website. It has been sent to me. This notebook is for us to ease updating the data in the future.

notebooks/preprocess-irena-data.ipynb
timtroendle: Actually no, because this is not in-the-loop for reasons I explained above.

src/hydro/inflow_mwh.py
timtroendle: Yes, but it actually is: the unit is the variable name, see line 10. This dataframe is only internal anyways.

timtroendle: Can do entirely within pandas, ok.

notebooks/preprocess-irena-data.ipynb
timtroendle: Actually, I don't remember why I did this back then. Will have a look again if it can be shortened.

src/hydro/inflow_mwh.py
brynpickering: Fair enough, then don't bother renaming it?

timtroendle: I rename it to avoid a line like this: `annual_national_generation_mwh.loc["generation_gwh"]`

notebooks/preprocess-irena-data.ipynb
timtroendle: Hmm yeah, dropna only works on data, not on the index. I could iterrupt here and reset the index based on a na-cleaned index, but don't think that's much cleaner.

notebooks/preprocess-irena-data.ipynb
timtroendle: I removed the country ids and keep all countries.

brynpickering (approved):

Issue #41, opened 2021-02-01
OPSD load data sources


brynpickering:
OPSD is the source for load data in Euro-Calliope. It combines data published by ENTSOE and by some individual TSOs, and combines all these sources in the 2019 dataset we are using.

The current method for selecting which load data source to use in Euro-Calliope is aimed at selecting the ENTSOE transparency platform data in reference of other sources (inc. 'power statistics'). This method has issues in two ways:

1. The code orders the data by source, in reverse alphabetical order. It then removes duplicates, with the expectation that the transparency data comes before the power statistics data. From an inspection of the CSV, this actually places 'load forecast' data first, meaning that the transparency platform data isn't chosen in many instances.
2. OPSD suggests that power statistics data is preferable. In [their documentation](https://nbviewer.jupyter.org/github/Open-Power-System-Data/datapackage_timeseries/blob/2020-10-06/main.ipynb) they state:

> The two sources differ Values on PS (~500 TWh annaually in Germany) are usually slightly higher than on the TP (~490 TWh). The reason probably lies with different reporting deadlines: Values on the TP have to be reported "no later than one hour after the end of the operating period". For the PS, the data is published with a delay of up to 3 months, which might allow for more accurate metering. For a comparison of the two sources see Hirth, et al. (2018).

Based on these two points, I would a. choose the power statistics from the OPSD load data, and b. explicitly define an order of gap filling from the remaining data sources. The order I propose is:
1. `actual_entsoe_power_statistics`
2. `actual_entsoe_transparency`
3. `actual_tso`
4. `actual_net_consumption_tso`

NOTE: as of the 2020 dataset, it seems (perhaps erroneously?) that only the transparency platform data is available from the OPSD timeseries dataset. [The documentation for the dataset](https://nbviewer.jupyter.org/github/Open-Power-System-Data/datapackage_timeseries/blob/2020-10-06/main.ipynb) still suggests there should be multiple sources available.

brynpickering:
Using this proposed method (see #42), there is a reasonable discrepancy in hourly data. This table shows the statistics for the ratio between hourly data, using the old method vs the new method.

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>min</th>
      <th>median</th>
      <th>mean</th>
      <th>max</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>ALB</th>
      <td>1.000000</td>
      <td>1.000000</td>
[...]

PR #3, opened 2019-11-18
Add gap filling of national electricity demand from other years

brynpickering:
Currently, ALB demand data is filled in for any year prior to 2017 using the 2017 data. However, there are other countries with missing data when considering years prior to 2016 (e.g. 2007, see log below).

So, this is an update on the ENTSOE data loading and processing function, to fill missing data more intelligently than a single hard-coded case. Certainly, how the data is filled in is up for discussion. At the moment, data from different years are stitched together, if there isn't enough data in the nearest available year. E.g. if you want to fill 2006 data but only have 9 months of data from 2007, the remaining 3 months of data will come from either 2005 or 2008, depending on which can provide that data.

Given the possible correlation between demand and weather, and intra-annual demand correlation, combining different years of data is never ideal. But it is still probably preferable to no data at all, so long as it is clear how the data has been combined. So, I've also included a more verbose log output. 

<details><summary><b>2007 'electricity_load_national.log' with PR</b></summary>
<p>

```
Country FIN has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country CHE has 81 missing load values, a working dataset was constructed from year(s) 2008. 81 missing timesteps will be filled from nearby values
Country LVA has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country GBR has 8760 missing load values, a working dataset was constructed from year(s) 2010,2011. 8760 missing timesteps will be filled from nearby values
Country IRL has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country ISL has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country CYP has 8760 missing load values, a working dataset was constructed from year(s) 2013. 8760 missing timesteps will be filled from nearby values
Country TUR has 8760 missing load values, a working dataset was constructed from year(s) 2016. 8760 missing timesteps will be filled from nearby values
Country LTU has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country DNK has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8760 missing timesteps will be filled from nearby values
Country NOR has 8760 missing load values, a working dataset was constructed from year(s) 2010. 8752 missing timesteps will be filled from nearby values
[...]

PR #42, opened 2021-02-01
Entsoe data priority

brynpickering:
Change order of priority of OPSD load data sources and explicitly define the order. See #41 for more information.


src/national_load.py
```diff
@@ -49,12 +54,29 @@ def read_albania(path_to_raw_load, number_rows_valid, other_ts):
...
+        reported "no later than one hour after the end of the operating period".
+        For the PS, the data is published with a delay of up to 3 months,
+        which might allow for more accurate metering.
+        For a comparison of the two sources see Hirth, et al. (2018).
+    See https://nbviewer.jupyter.org/github/Open-Power-System-Data/datapackage_timeseries/blob/2020-10-06/main.ipynb for more info.
+    """
+    load_by_attribute = (
+        load
+        .set_index(["region", "utc_timestamp", "attribute"])
+        ["data"]
```
timtroendle, line 72: ``.loc[:, "data"]`` would be slightly nicer here, as it's more explicit.

timtroendle (approved):
Looks good to me! However, as mentioned over in #41: as entso-e is dropping PS, should we rely on transparency data for future compatibility?

brynpickering:
@timtroendle in light of the discussion in #41, maybe this would be better dealt with by having the order preference in the config, so a user can decide if they want to drop PS (I would include it preferentially still, based on its seemingly higher accuracy).

timtroendle:
Fine with me.

timtroendle (changes requested):
Changes are fine but need to be added in the changelog.

brynpickering:
Once #45 is in, I'll also update this PR to include a schema entry for the priority list.

brynpickering:
@timtroendle bump

scripts/national_load.py
```diff
@@ -49,12 +49,30 @@ def read_albania(path_to_raw_load, number_rows_valid, other_ts):
...
 
-def remove_entsoe_power_statistic_data_where_possible(load):
-    sorted_load = load.sort_values(
-        "attribute",
-        ascending=False
-    ) # will end with entsoe-transparency ahead of entsoe-power-statistics
-    return sorted_load.drop_duplicates(["region", "utc_timestamp"], keep="first")
+def select_statistics_by_source_priority(load, entsoe_priority):
+    """
+    Choosing `entsoe_power_statistics` as main source since OPSD states:
```
timtroendle, line 54: This comment here is weird. You are not selecting here, but in the config.

scripts/national_load.py
```diff
@@ -49,12 +49,30 @@ def read_albania(path_to_raw_load, number_rows_valid, other_ts):
...
+    load_by_attribute = (
+        load
+        .set_index(["region", "utc_timestamp", "attribute"])
+        .loc[:, "data"]
+        .unstack("attribute")
+    )
+    load_top_priority = load_by_attribute[entsoe_priority[0]]
+    if len(entsoe_priority) > 1:
+        for source in entsoe_priority[1:]:
+            load_top_priority = load_top_priority.fillna(load_by_attribute[source])
```
timtroendle, line 73: I am not so sure about this. Should we really mix data sources? I'd rather not. When there are only a few hours to fill, I think an interpolation is the best you can do.

Mixing data sources between countries could be fine I'd say. So if "Albania" is not within your top priority data source, go look Albania up in the next on the list.

timtroendle (changes requested):
I am not convinced by the approach of mixing data sources. I don't think we should do that.

scripts/national_load.py
brynpickering: well, the comment was from before it moved to config ;)

scripts/national_load.py
brynpickering: Hmm, over one or two hours I agree that interpolation might be better. The majority of missing source data for a particular country is for the entire year. But what about those few cases over several days? In #3 There is gap filling for more than a few days using previous years' data. I would say that this gap filling should preferentially come from alternative data sources for the same year, then from different years, then by interpolation.

Maybe I merge in #3 and then incorporate this into the gap filling process? Go with filling data for more than a few hours from alternative data sources, then from other years, then leave the last few hours to be filled in by interpolation?

timtroendle: I find other data sources problematic because they may literally represent a different measure. Using other years can be problematic as well: what if 15 April was not a workday in that other year, but a weekend? Or a holiday?

I'm fine having all these mechanism in, but the default config should be very conservative. Then, one could do all sorts of gap filling things, but one would need to do that explicitly.

brynpickering: I agree that it isn't ideal, but that is kind of where we get to when using incomplete datasets. The different sources don't necessarily represent a different measure. More often they are just a different method of quantification of the same measure. Those sources that would be different are low down in the priority list (net consumption vs gross generation). Nevertheless, their profiles would be accurate, just their magnitudes would be slightly different (and by slightly, we're talking on the order of <10% since they just represent the differences caused by transmission losses).

If April 15th were a holiday one year and not the next, the profile and the magnitude would be slightly different, but the error would be *much* less than linearly interpolating from the demand at 23:00 on the 14th and 00:00 on the 16th. 

If we were being conservative, the current workflow should possibly not run: using 2017 Albanian data puts it out of sync with its neighbours who are all using 2016 data. For me the most conservative is to fail and then ask the user to allow gap filling mechanisms. In order of preference for least deviation from what 'actual' demand might be, I would go with: 

1. go down the priority list of data sources, replacing the whole year of data if that is necessary, provided the next options down the list has more data than the one it is replacing,
2. fill from other years where large gaps still exist, and
3. fill by interpolation.

timtroendle: Yes, that sounds good to me. Fail, and then let the user decide. And be verbose with manipulations in the log messages, so that the user can afterwards analyse the gap filling.

brynpickering:
@timtroendle I think I would merge this and #3 in (they both run on the minimal workflow) and then start another PR that merges the two sets of functionalities into one that makes the gap filling more user-driven.

timtroendle:
Good for me. I opened #66 to keep track of these changes. You can also use #66 for the changelog.

timtroendle (approved):
Fine with merging this in. Still, I'd prefer to switch this mechanism off by default before merging.

brynpickering:
Since such a switch hasn't ever really existed, I'll introduce it in the next PR.

timtroendle:
You could simply remove the other data sources from the config. ;-) I'm just a little concerned because I am working from `develop` at the moment and I'd like to keep it clean because of that (and I think we should always keep it clean). But here as well, if you promise to add the _real_ switch soon, just merge and go on.

brynpickering:
> You could simply remove the other data sources from the config.

OK, that's done!

timtroendle:
Thanks!

PR #44, opened 2021-04-01
Update atlite and geo packages

timtroendle:


envs/hydro.yaml
```diff
@@ -3,21 +3,16 @@ channels:
     - conda-forge
 dependencies:
     - python=3.7
```
brynpickering, line 5: Does it make sense to remain consistent between sub-environments with the python version?

src/hydro/runoff.py
```diff
@@ -1,72 +1,34 @@
...
 
-X_RANGE = slice(-10.75, 34.75)
-Y_RANGE = slice(71.25, 34.5)
 
-
-def runoff(path_to_cutout, year):
+def runoff(path_to_cutout, year, x_min, x_max, y_min, y_max):
     """Retrieve runoff data from ERA5 as atlite cutout."""
+    logging.basicConfig(level=logging.INFO)
     path_to_cutout = Path(path_to_cutout)
```
brynpickering, line 11: As with inflow_m3.py, you probably don't need this any more

src/hydro/runoff.py
```diff
@@ -1,72 +1,34 @@
...
-
-    monkeypatch_atlite_to_download_runoff_only()
+    x_range = slice(x_min, x_max)
```
brynpickering, line 12: does `slice` include the last element, or not?

src/hydro/runoff.py
```diff
@@ -1,72 +1,34 @@
 """Retrieve runoff data from ERA5 as atlite cutout."""
+import logging
 from pathlib import Path
 
 import atlite
-import xarray as xr
 
-X_RANGE = slice(-10.75, 34.75)
```
brynpickering, line 7: nice

brynpickering (commented):
Looks like a nice cleanup (on atlite's side of things, as much as anything!). I will assume there's been no breaking changes from the version updates ;)

envs/hydro.yaml
timtroendle: Good point. Will update everything to 3.8.

src/hydro/runoff.py
timtroendle: Not sure if that's a "slice" feature or an "atlite" feature, but it does.

timtroendle:
Ready for you to review again @brynpickering. I've added a changelog and will use that for the other pull requests as well.

brynpickering (approved):

PR #43, opened 2021-02-09
Update JRC Hydro Database version

brynpickering:
From V4 to V7, and update the method extracting data from the zip to be version hash agnostic.

Issue #19, opened 2020-05-28
Update to hydro power database release 07


timtroendle:
https://github.com/energy-modelling-toolkit/hydro-power-database/releases/tag/v7

timtroendle:
See also #8 .

timtroendle:
The dataset as introduced in #48 does not work: it contains unknown country codes. First, it contains entries for Kosovo. Second, it contains EU country code instead of ISO 3166 country code.

brynpickering:
😔

timtroendle:
I'm fixing it right now. Still, this is the third issue in less than 24h that could have been caught by CI. We should try to tackle #56 in same way.

timtroendle:
There is another issue with 07: there are 4 duplicated indices in the dataset. See https://github.com/energy-modelling-toolkit/hydro-power-database/releases/tag/v8.

PR #48, opened 2021-04-07
update-jrc-hydro-from-develop

brynpickering:
Only updates JRC hydro database version, fixing #19 and superseding #43

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -11,7 +11,6 @@ data-sources:
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
     potentials: https://zenodo.org/record/3533038/files/possibility-for-electricity-autarky.zip
-
 root-directory: . # point to the root directory if working directory is not root directory
 scaling-factors: # values are tuned for models with a few hours resolution and one year duration
     power: 0.00001 # from MW(h) to 100 GW(h)
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -7,10 +7,11 @@ data-sources:
     national-phs-storage-capacities: data/pumped-hydro/storage-capacities-gwh.csv
     capacity-factors: https://zenodo.org/record/3899687/files/{filename}?download=1
     gadm: https://biogeo.ucdavis.edu/data/gadm3.6/gpkg/gadm36_{country_code}_gpkg.zip
-    hydro-stations: https://zenodo.org/record/3371462/files/energy-modelling-toolkit/hydro-power-database-v4.zip?download=1
+    hydro-stations: https://zenodo.org/record/4289229/files/energy-modelling-toolkit/hydro-power-database-v7.zip?download=1
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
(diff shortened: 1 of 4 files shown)


scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -34,25 +20,6 @@ def preprocess_stations(path_to_stations, path_to_basins, buffer_size, path_to_o
     )
 
 
-def add_romanian_phs(stations):
```
brynpickering, line 37: According to the comparison table I give in #42, the Romanian PHS capacity is now given in the JRC database, so no need to manually add it from Geth.

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -34,25 +20,6 @@ def preprocess_stations(path_to_stations, path_to_basins, buffer_size, path_to_o
...
-    return pd.concat(
-        [stations, ROMANIAN_PHS],
-        axis=0,
-        sort=False
-    )
-
-
-def remove_index_duplicates(stations):
```
brynpickering, line 45: This issue is now closed with the v7 release

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -3,26 +3,12 @@
 from shapely.geometry import Point
 
 MAXIMUM_NUMBER_OF_DROPPED_STATIONS = 5
```
brynpickering, line 5: Can you explain what this is? should it be in the config?

config/default.yaml
```diff
@@ -7,10 +7,11 @@ data-sources:
     national-phs-storage-capacities: data/pumped-hydro/storage-capacities-gwh.csv
     capacity-factors: https://zenodo.org/record/3899687/files/{filename}?download=1
     gadm: https://biogeo.ucdavis.edu/data/gadm3.6/gpkg/gadm36_{country_code}_gpkg.zip
-    hydro-stations: https://zenodo.org/record/3371462/files/energy-modelling-toolkit/hydro-power-database-v4.zip?download=1
+    hydro-stations: https://zenodo.org/record/4289229/files/energy-modelling-toolkit/hydro-power-database-v7.zip?download=1
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
     potentials: https://zenodo.org/record/3533038/files/possibility-for-electricity-autarky.zip
+
```
brynpickering, line 14: erroneous whitespace, will remove

scripts/hydro/preprocess_hydro_stations.py
timtroendle: Well as you see in the assert statement below it limits the number of stations that can be removed from the dataset. Stations without installed capacity are removed.

This could move into the config, but as really it is a detail (and a low number), I'd be fine with keeping it here.

timtroendle (commented):
Looks good to me. Can you add the change to the changelog?

scripts/hydro/preprocess_hydro_stations.py
brynpickering: OK, I asked only because you were keen on the OPSD data gap filling threshold in #3 to be moved to the config. This seems to be a similar style of threshold...

timtroendle: Sure, but the important difference is: one was written by me, the other by you. 😎

I think there is a small difference between the two, but again -- I am not against moving this into config as well.

timtroendle (approved):

scripts/hydro/preprocess_hydro_stations.py
brynpickering: Happy not to move it, since that's less work for me!

PR #49, opened 2021-04-07
Make scaling PHS with Geth et al. optional

brynpickering:
Supersedes part of #43, where discussion can be found on the impact.

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -44,6 +44,7 @@ parameters:
         potential-year: "2020"
         cost-year: "2030"
     biofuel-efficiency: 0.45
+    scale-phs: false
 sea-connections:
     continental:
     national: # Source: https://www.entsoe.eu/data/map/
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -44,7 +44,7 @@ parameters:
         potential-year: "2020"
         cost-year: "2030"
     biofuel-efficiency: 0.45
-    scale-phs: false
+    scale-phs-according-to-geth-et-al: false
 sea-connections:
     continental:
     national: # Source: https://www.entsoe.eu/data/map/
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -53,3 +53,9 @@ properties:
     root-directory:
         type: string
         description: Path to the root directory of euro-calliope containing scripts and template folders.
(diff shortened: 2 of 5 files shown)


config/schema.yaml
```diff
@@ -53,3 +53,9 @@ properties:
     root-directory:
         type: string
         description: Path to the root directory of euro-calliope containing scripts and template folders.
+    parameters:
+        type: object
+        properties:
+            scale-phs:
+                type: boolean
+                description: Whether or not to use pumped hydro capacities derived from Geth et al. (2015) to scale pumped hydro capacities givven by the JRC hydro database
```
timtroendle, line 61: You removed the newline here... and I opened #50 :D

config/default.yaml
```diff
@@ -44,6 +44,7 @@ parameters:
         potential-year: "2020"
         cost-year: "2030"
     biofuel-efficiency: 0.45
+    scale-phs: false
```
timtroendle, line 47: A little bit more information here would be good. Why not: `scale-phs-according-to-geth-et-al: false`

timtroendle: In the code, the short version is fine. Here, I'd go for something more self-explanatory.

scripts/hydro.py
```diff
@@ -72,8 +75,8 @@ def read_national_phs_storage_capacities(path_to_data, locations):
 def scale_phs_storage_capacities(hphs, locations, national_storage_capacities):
     """Scale PHS storage capacities to match (Geth et al., 2015).
 
-    Storage capacities of pumped hydro within the JRC data base seem too high. Thus, we are scaling
-    them here, so that the national numbers of (Geth et al., 2015) are fulfilled.
+    Storage capacities of pumped hydro within the JRC database may seem too high.
+    Thus, they can be optionally scaled here, so that the national numbers of (Geth et al., 2015) are fulfilled.
```
timtroendle, line 79: Being picky: they aren't _optionally_ scaled in this function. They are compulsory scaled. 🤓

timtroendle (changes requested):
Looks good. But please add this to the changelog. This is a major change.

timtroendle (approved):

PR #51, opened 2021-04-07
Add automatic download of hydro basins database

### LINKED ISSUE #34: Automatically derive HydroBASINS data
Currently, the data must be downloaded manually. Add a rule to download the data automatically.

timtroendle:
Fixes #34.


rules/hydro.smk
```diff
@@ -31,6 +41,13 @@ rule download_stations_database:
         "curl -sLo {output} '{params.url}'"
 
 
+rule basins_database:
+    message: "Unzip basins database."
+    input: rules.download_basins_database.output
+    output: "build/data/basins/hybas_eu_lev07_v1c.shp"
+    shell: "unzip {input} -d ./build/data/basins/"
```
brynpickering, line 48: Is the shapefile the only thing inside the .zip? If not, I would specify in the `unzip` command that you just want the shapefile, such that we minimise random files ending up in `build`

brynpickering (approved):
I don't like the idea of depending on dropbox for a file, but otherwise looks fine to me!

rules/hydro.smk
timtroendle: "Shapefiles" are actually a [combination of several files --  three of them mandatory (including the `.shp` file)](https://en.wikipedia.org/wiki/Shapefile).

I don't like that. That fact alone is a good reason for me to avoid shapefiles where possible.

timtroendle:
I agree with the Dropbox link, but that's the only public download they provide.

PR #53, opened 2021-04-07
Update to remove conflicts associated with py3.8

brynpickering:
On updating dependencies in #44, defaults.yaml was updated to python 3.8. The other packages in the env needed to also be updated to avoid conflicts. This updates them to match package versions in other envs.


timtroendle:
Argh, sorry for that. CI, where are you?

timtroendle (approved):

PR #58, opened 2021-04-08
Fix issues with JRC hydro stations database v07

### LINKED ISSUE #19: Update to hydro power database release 07

timtroendle:
Fixes #19.


timtroendle:
@brynpickering can you put this high on your priority list? It blocks #60 and my own work moving forward. Sorry for being pushy!

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -44,6 +71,20 @@ def new_coords(station, buffer_size, hydrobasins):
     return point.buffer(buffer_size).intersection(closest_basin.geometry).representative_point()
 
 
+def eu_country_code_to_iso3(eu_country_code):
```
brynpickering, line 74: Could you place this in a general util directory, since it come up elsewhere (`scripts/biofuels.py`, `scripts/shapes/nuts.py`).

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -1,39 +1,66 @@
...
-    assert stations[~is_in_basin].apply(station_in_any_basin(hydrobasins), axis="columns").all()
+    stations = ensure_stations_in_basins(stations, hydrobasins, buffer_size)
     stations.to_csv(
         path_to_output,
         header=True,
         index=True
     )
 
 
+def resolve_index_duplicates(stations):
```
brynpickering, line 24: this would be a clean, one-line alternative: `df.index = df.index.where(~df.index.duplicated(), df.index + '_b')` (from [here](https://stackoverflow.com/questions/43095955/rename-duplicated-index-values-pandas-dataframe))

brynpickering (changes requested):
Damn these datasets having so many issues...

timtroendle:
Thanks a lot for the fast review, @brynpickering !

scripts/hydro/preprocess_hydro_stations.py
timtroendle: Absolutely agree, but I don't know how to do that properly. First, all these files are scripts, not Python modules, and there is no Python package that one could import. Second, the Snakemake rule would need to be rerun if that central file was changed. That's (currently) only possible by adding the central file to the input list of the rule. That's possible, but error-prone: you can easily forget to add this file there.

For those two reasons, I gave up imports in my Snakemake workflows. If you have a good idea to solve these two issues, I am happy to improve this.

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -1,39 +1,66 @@
...
+import pycountry
 
 MAXIMUM_NUMBER_OF_DROPPED_STATIONS = 5
 
 
 def preprocess_stations(path_to_stations, path_to_basins, buffer_size, path_to_output):
     stations = pd.read_csv(path_to_stations, index_col=0)
-    assert stations.installed_capacity_MW.isna().sum() <= MAXIMUM_NUMBER_OF_DROPPED_STATIONS
-    stations.dropna(subset=["installed_capacity_MW"], inplace=True)
+    stations = resolve_index_duplicates(stations)
```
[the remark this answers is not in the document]
timtroendle: I see your point but I disagree. This is partly a matter of style, but partly there are also reasons.

1. This way, the main function is simply a list of manipulations of the dataset. Each line is a human-readable description of the manipulation.
2. This way, each function has exactly one task. No task is hiding within functions.
3. It avoid the necessity for inline comments. Take the drop of Kosovo entries: without the function name (and without an inline comment) you would not know what's going on (I wouldn't!).

Point 2 and 3 are mentioned in "The Clean Code" ([summary 1](https://medium.com/mindorks/how-to-write-clean-code-lessons-learnt-from-the-clean-code-robert-c-martin-9ffc7aef870c), [summary 2](https://www.freecodecamp.org/news/clean-coding-for-beginners/)).

> Functions should do one thing. They should do it well. They should do it only. – Clean Code

> Do not leave code in comments

I understand your concerns and agree that one-line-functions are a borderline case. Still, keeping the main function as a list of human-readable manipulation steps makes sense to me -- even if these steps are one-line functions.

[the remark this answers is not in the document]
timtroendle: Plus, there are a few more manipulation steps incoming within #60 🙈.

scripts/hydro/preprocess_hydro_stations.py
brynpickering: how about you have to point to dependencies within the workflow to be able to use them (instead of it being 'optional' to have them as an input)? E.g. if you want to access helper functions then you have to specify e.g.:

```
inputs:
    util="scripts/utils.py"
```
And in the python file:
```
spec = importlib.util.spec_from_file_location("foo", snakemake.inputs.util)
util = importlib.util.module_from_spec(spec)
spec.loader.exec_module(util)
...
util.eu_country_code_to_iso3(...)
```
This is more verbose but explicitly requires a user to go via snakemake (and its tracking of file changes) to import from other files in the workflow.

If it's too much, I'd say that still having everything as util functions and importing them is better than writing the same function N times. If the function needs updating, it seems much less error-prone to have to re-run specific rules manually than it is to go about changing every function that is doing the same thing.

timtroendle: Ok, that seems to work. A little verbose, as you say, but still better than re-re-redefining the function.

scripts/hydro/preprocess_hydro_stations.py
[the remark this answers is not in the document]
brynpickering: "The Clean Code" seems a bit like the bible, it all depends on which passages you choose to quote. From summary 1:

> Minimize the number of entities such as classes, methods, functions, and the like.
> Contains no duplication

I would argue that one line functions are certainly not looking to minimise the number of entities, and they are also creating line duplication (you use more lines to call the function and have the result returned than the lines of useful code doing anything).
 
Also, "Do not leave code in comments" is about leaving redundant code commented out, not about leaving comments in the code!

Anyway, we could battle over this forever, let's leave it for now.

scripts/hydro/preprocess_hydro_stations.py
brynpickering: will you implement the change in this PR or later?

timtroendle: Did it just now within this PR.

scripts/hydro/preprocess_hydro_stations.py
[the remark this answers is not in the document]
timtroendle: I don't think there is a contradiction, but rather a trade-off in function length. I agree with you that this PR leans towards the too-short-end of that trade-off.

Let's have a look again at this in #60!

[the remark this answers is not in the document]
timtroendle: Oh, but regarding comments: no, that's not what I mean. Read the section about comments here:
https://medium.com/mindorks/how-to-write-clean-code-lessons-learnt-from-the-clean-code-robert-c-martin-9ffc7aef870c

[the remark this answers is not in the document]
brynpickering: Sorry for the accusation! The 'do not leave code in comments' quote is from a subheading in https://www.freecodecamp.org/news/clean-coding-for-beginners/ regarding commenting out code. I agree that minimising the number of comments is a good thing, but where comments reduce code verbosity rather than add to it I disagree with the idea that including them is a 'blunder'.

scripts/shapes/nuts.py
```diff
@@ -8,6 +9,12 @@
 import pandas as pd
 import pycountry
 
+# import utility module
+assert "utils" in snakemake.input.keys(), "Make sure utils.py module is in your snakemake inputs."
```
brynpickering, line 13: This line isn't needed, since you will get this error no matter what if 'utils' is not in snakemake.input

scripts/biofuels.py
```diff
@@ -1,10 +1,16 @@
...
+import importlib
 from enum import Enum
 from pathlib import Path
 
 import pandas as pd
 import geopandas as gpd
-import pycountry
+
+# import utility module
+assert "utils" in snakemake.input.keys(), "Make sure utils.py module is in your snakemake inputs."
```
brynpickering, line 10: Not needed (see later comment)

brynpickering (changes requested):
Nice, I actually think I like the requirement to point to the utils file in the workflow. It makes it nice and explicit what your script depends upon.

Just some minor comments and then this can go through.

scripts/shapes/nuts.py
timtroendle: Not exactly!

With this line, you receive the following error message:

```bash
Traceback (most recent call last):
  File "/Users/trtim/Develop/euro-calliope/.snakemake/scripts/tmpoqoell39.preprocess_hydro_stations.py", line 13, in <module>
    assert "utils" in snakemake.input.keys(), "Make sure utils.py module is in your snakemake inputs."
AssertionError: Make sure utils.py module is in your snakemake inputs.
```

Without, you receive the following error message:

```bash
Traceback (most recent call last):
  File "/Users/trtim/Develop/euro-calliope/.snakemake/scripts/tmpwmu9mvig.preprocess_hydro_stations.py", line 14, in <module>
    spec = importlib.util.spec_from_file_location("utils", snakemake.input.utils)
AttributeError: 'InputFiles' object has no attribute 'utils'
```

I thought the former is more helpful should anyone run into this situation.

timtroendle:
Yeah, I also like this approach you suggested. For 1 or maybe 2 imports, it works well.

If we ever run into the situation where we need more imports, we may go for another solution. We can separate `scripts` from `library` code in the folder structure and make the library code a versioned Python package. This Python package can then be installed using the conda environment files and imported like any other package.

timtroendle:
Ok, I was too nosy to not follow this idea and so I replace the current approach with the introduction of `eurocalliopelib`: a Python package that provides library code for the workflow. It's easy to test and can be used for stable, reusable functions and routines. We've discussed this last week or so with @suvayu .

I've also added a simple CI test -- see [my fork](https://github.com/timtroendle/euro-calliope/actions) for how this'll look after merging.

For what it currently is (one simple function), it's over-engineered. But I'm sure we'll find more functions and routines to add to `eurocalliopelib`. @brynpickering @suvayu what do you think?

envs/hydro.yaml
```diff
@@ -16,3 +16,6 @@ dependencies:
     - cdsapi=0.5.1
     - pycountry=18.12.8
     - atlite=0.2.1
+    - pip=21.0.1
+    - pip:
+        - eurocalliopelib==0.1.0 @ file:///../../../lib/
```
brynpickering, line 21: Is there some way to be able to flag to a user that the current version of eurocalliopelib doesn't match what is here? I could see the lib being updated and a user not realising that this version needs updating accordingly for it to filter through...

lib/eurocalliopelib/utils.py
```diff
@@ -0,0 +1,18 @@
...
+def eu_country_code_to_iso3(eu_country_code):
+    """Converts EU country code to ISO 3166 alpha 3.
+    The European Union uses its own country codes, which often but not always match ISO 3166.
+    """
+    assert len(eu_country_code) == 2, "EU country codes are of length 2, yours is '{}'.".format(eu_country_code)
+    if eu_country_code.lower() == "el":
+        iso2 = "gr"
+    elif eu_country_code.lower() == "uk":
+        iso2 = "gb"
+    elif eu_country_code.lower() == "bh": # this is a weird country code used in the biofuels dataset
```
brynpickering, line 14: a comment :O

lib/setup.py
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python3
+
+from setuptools import setup, find_packages
+
+setup(
+    name='eurocalliopelib',
+    version='0.1.0',
+    description='Library code of the euro-calliope workflow.',
+    maintainer='Tim Tröndle',
```
brynpickering, line 9: maybe make this generically calliope-project / euro-calliope AUTHORS?

lib/setup.py
```diff
@@ -0,0 +1,22 @@
...
+    maintainer_email='tim.troendle@usys.ethz.ch',
+    packages=find_packages(exclude=['tests*']),
+    include_package_data=True,
+    install_requires=["pycountry"],
```
brynpickering, line 13: specific version?

brynpickering (approved):
I like it, although I'm unfamiliar with github workflows, so will just trust you on that (I guess @suvayu will be able to comment more on that). As you say, a little over-engineered, but I already see use for it for various other helper functions moving forward.

I've made some minor comments, but if they are ultimately ignored I don't mind.

envs/hydro.yaml
timtroendle: Good point. If the versions don't match, the env won't get installed -- that's fine!

However, whenever you make changes to `eurocalliopelib` you'll need to manually recreate the envs; otherwise the changes won't make it through. I will add a comment to `README.md` -- but is that enough?

brynpickering: so you need to manually update the requirements file and that will automatically trigger the env to be recreated (since snakemake sees a change in the dependencies), otherwise it will continue using the old version that is baked into the conda env that already exists... hmmm, there seems to be no way of doing this without the possibility of a user missing a step without realising it...

timtroendle: Oh yes, you are right about the automatic trigger.

It's not ideal and I see us confused in front of our computers on Friday evenings trying to fix this tiny issue on `eurocalliopelib` and the fix simply won't get through...

Still, `eurocalliopelib` is not meant to be touched often and not meant to be touched by "a user" (as in: someone not among us). That should reduce the likelihood of such a situation happening.

In addition to commenting the readme, we could also use a tool (I am looking at you, [versioneer](https://github.com/python-versioneer/python-versioneer)) that handles the version update in all relevant locations with a single command.

scripts/hydro/preprocess_hydro_stations.py
timtroendle: And now I replaced that mechanism with a proper Python package: `eurocalliopelib`.

envs/hydro.yaml
brynpickering: how about eurocalliopelib raises a warning on import when the locally available version in `lib/...` doesn't match the version that has been imported? It would give us the relevant info on a friday evening to quickly debug? For this PR, I say no need to worry, would be worth tracking this issue though.

suvayu:
@timtroendle @brynpickering I had a quick look at the last two commits (c933c2bcf61451e8c98c080ea7668331e23a9de3 & b80b8d33ff695f14904667ea71aa36e6ec0719b4), seems fine. Except for the install issue discussed above.  What you need is an editable install, aka `pip install -e /path` or `conda develop /path` (you need `conda-build` for this).  However I don't know how to do this for an environment.  In case it helps, in a GH action, I achieve this by having `pip install -r requirements-for-everything-else.txt` followed by `pip install -e /path`.  I hope that helps.

brynpickering:
You can make it an editable install in conda too: 

```yaml
- pip:
        - -e /path/to/module
```

I guess the issue of this is the inverse: you have no version control so you can't so easily fix the version for the purposes of reproducibility. You could make it editable and then have the expected version in the config and then have a single assert statement in the main snakemake workflow that verifies that the version on file matches what the workflow expects...?

suvayu:
I'm not clear what's the expected way to run this, from source (git checkout, or unpacked archive)?  Either way, the distributed source tree is clearly versioned with git.  So if someone makes changes, then a check like this redundant, or am I missing something?  Is it the intention to warn a user they have edited the files?  Then the best tool is to use git (`git describe --tags --dirty`).

timtroendle:
Thanks @brynpickering @suvayu . I solved the version / dependency thing the following way now: I install the lib in editable mode but I add its version number to the model build metadata. So whenever you see a model build, you know the library version that has been used to build it. Not sure that's necessary, but let's see.

Let's go with this approach and see how well it works. It's all internal, so there is no risk.

Issue #56, opened 2021-04-08
Add a minimal test of entire workflow and run it continuously


timtroendle:
A continuous integration test would catch errors early. For that, the workflow must be 100% automatic and we should have a configuration that requires minimal downloads and minimal runtime. We can then use a simple GitHub action that runs Snakemake with this configuration ([example](https://github.com/timtroendle/cookiecutter-reproducible-research/blob/main/.github/workflows/reproduction.yaml)).

timtroendle:
Here's a list of things that need to be solved (sorted by severity) so that we have a 100% automatic, low data, low runtime workflow test:

- [ ] Smaller capacity factor data (limited to geographic or temporal scope).
- [ ] Automatic download of EEZ data.
- [ ] Smaller or faster load data download (limited to geographic or temporal scope).
- [ ] Smaller or faster runoff data download (geographic and temporal scope can be limited today, but download can still be slow).

The first point (capacityfactors) is by far the worst. When we solve the issues in the list above, all other steps should be very fast, especially when we limit the geographic scope.

brynpickering:
Agreed that we need this, but limiting scope also creates issues with not picking up some of the more annoying elements of the workflow that we probably want to be testing (e.g. handling Kosovo, filling missing load data, getting ISO vs. EU country codes, assigning plants to basins when they don't fall in one). I can see a different geographic region being needed to catch each of these...

timtroendle:
You are right. Still, any error that is caught by CI is good. So just because the CI test wouldn't catch _all_ errors doesn't mean it's useless.

We can also apply an hierarchical approach: a simplified workflow that is run very often, and a full workflow that is run less often.

timtroendle:
We have a similar issue for the estimation of solar and wind potentials. The discussion over there is relevant here, too. 
https://github.com/calliope-project/solar-and-wind-potentials/issues/11

timtroendle:
Bryn mentioned in the dev call today that data caching is possible on GitHub runners.

PR #60, opened 2021-04-08
Feature minimal workflow

timtroendle:
This is a first step towards continuous integration (#56) and should allow to run integration tests manually. This configuration takes less than 5 minutes to run on my machine, with three concurrent jobs and with automatic downloads available.

I could imagine us to enforce successful runs in each of our future pull requests -- right before merging changes into `develop`. As long as issues in #56 aren't solved, it has to run locally by the author of the pull request.

In the process, I also moved estimation of storage capacities into the preprocessing of the JRC hydro stations database. Apart from problems that arose in its old place, the new place fits logically much better.


timtroendle:
This is currently blocked by #58.

envs/test.yaml
```diff
@@ -3,7 +3,7 @@ channels:
     - conda-forge
     - gurobi
 dependencies:
-    - python=3.8
+    - python=3.7
```
brynpickering, line 6: I would go the other way and update the other dependencies. It seems a more 'robust' test of the workflow, which ultimately works with py3.8 and e.g. pandas>1.0

rules/hydro.smk
```diff
@@ -59,26 +59,32 @@ rule stations_database:
...
     message: "Preprocess hydro stations."
     input:
         script = script_dir + "hydro/preprocess_hydro_stations.py",
         stations = rules.stations_database.output[0],
-        basins = rules.fix_basins.output[0]
-    params: buffer_size = 1 / 60 # move stations up to 1 arcminute < 1 km
+        basins = rules.preprocess_basins.output[0],
+        phs_storage_capacities = config["data-sources"]["national-phs-storage-capacities"]
+    params:
+        buffer_size = 1 / 60, # move stations up to 1 arcminute < 1 km
```
brynpickering, line 85: should be in config...?

scripts/hydro/preprocess_basins.py
```diff
@@ -3,26 +3,33 @@
...
-    """
-    basins = gpd.read_file(path_to_basins)
+def preprocess_basins(path_to_basins, bbox, path_to_output):
+    "Filter and fix basin shapes."
+    basins = gpd.read_file(path_to_basins, bbox=bbox)
     basins.geometry = basins.geometry.map(_buffer_if_necessary)
     basins.to_file(path_to_output, driver=DRIVER)
 
 
 def _buffer_if_necessary(shape):
```
brynpickering, line 13: once incorporated, move to eurocalliopelib and add in the same assertion as I did in `solar_and_wind_potentials`

brynpickering (commented):
Some initial comments before #58 is merged in

envs/test.yaml
timtroendle: Actually, the versions in here should be the recommended versions for the user of `euro-calliope`. In its current state, we'd say `euro-calliope` runs with Python 3.7, calliope 0.6.5, etc etc.

The dependencies in `envs/default.yaml` and the other env files are irrelevant: they are all for _building_ euro-calliope, while only `envs/test.yaml` is for running euro-calliope.

Still, that means we should likely update these dependencies and also communicate them as the recommended versions.

timtroendle: `Problem: package calliope-0.6.5-py_0 requires pandas >=0.25,<0.26`

We could update Calliope, but that is a larger change that I'd do in another PR if at all wanted. Therefore, I leave this env file as is for the moment.

scripts/hydro/preprocess_basins.py
timtroendle: Actually, this function is only used once within this repo _and_ it is difficult to test and therefore I don't see an added value of moving it.

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -32,6 +38,74 @@ def drop_kosovo(stations):
...
+    )
+    geth_national_capacities.index = stations.loc[hphs_mask].index
+    new_storage_capacities = storage_capacity_share * geth_national_capacities
+    stations.loc[hphs_mask, "storage_capacity_MWh"] = new_storage_capacities
+    return stations
+
+
+def read_national_phs_storage_capacities(path_to_data):
+    data = pd.read_csv(path_to_data, index_col=0)
+    data.index = [pycountry.countries.lookup(iso2).alpha_3 for iso2 in data.index]
```
brynpickering, line 101: data.index.map(...) and then ref a function in eurocalliopelib to process iso2->iso3? This process seems to happen a lot in various files, and would be nice to keep in one place in case pycountry does something silly in future that we have to account for.

brynpickering (commented):
This is a bit of a confusing PR since you introduce the minimal workflow alongside restructuring the hydro stuff. I assume the restructuring isn't entirely necessary for the minimal workflow to function?

Anyway, the restructuring makes sense. One thing I wonder about is whether the last hydro step (referencing 'hydro.py') should also be in 'hydro.smk', for consistency?

scripts/hydro/preprocess_basins.py
timtroendle: Yes, but maybe do it then? I am not 100% sure this'll work. I've played around with importing libraries of other workflows vut didn't manage to make it work.

config/minimal.yaml
```diff
@@ -0,0 +1,77 @@
+# A simple config to allow for quick test runs of the workflow.
```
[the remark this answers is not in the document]
timtroendle: True. Plus, it's completely unnecessary. It can inherit from `defaults.yaml`. I'll update this.

scripts/hydro.py
```diff
@@ -1,27 +1,17 @@
...
 import geopandas as gpd
 from shapely.geometry import Point
-import pycountry
 
 WGS_84 = "+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"
 
 
-def main(path_to_plants, path_to_locations, path_to_phs_storage_capacities, scale_phs, path_to_output):
+def main(path_to_plants, path_to_locations, path_to_output):
     locations = gpd.read_file(path_to_locations).to_crs(WGS_84).set_index("id")
```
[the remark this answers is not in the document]
timtroendle: Yeah, this has to do with some `proj4` conda-forge installation problem I was running into. I'll see if it works right now, but I don't think this is a big issue. This constant is correct and it's never supposed to change.

timtroendle:
In fact, the two parts of this PR are related: the minimal workflow would not work without restructuring the hydro module.

timtroendle:
About moving `hydro.py` to the `hydro` subfolder: The logic was that everything related to the `resolution`of euro-calliope  -- to the actual spatial scales -- happens within the top-layer `Snakefile`; everyhing in the subfolder is "just" preprocessing data. But you are right: one could also find good reasons to move it into the subfolder. As long as we don't have strong preferences, shall we keep it as is?

In any case, I renamed it to make its purpose clearer: `hydro.py` -> `hydro_capacities.py`

brynpickering (approved):
Just a changelog entry and we're good to go!

PR #62, opened 2021-04-13
Fix workflow-internal variable name of demand

timtroendle:
The old name was `industrial_demand` which was confusing because it contained both, industrial and total demand.

It took me a while to understand my own code -- so it has to change.


brynpickering (approved):
Makes sense to me!

PR #64, opened 2021-04-14
Add shell tools (curl/unzip) from conda-forge

### LINKED ISSUE #59: Use conda envs for non-Python tools to increase portability
All Python and close-to-Python tools run in a conda environment, which improves portability and reproducibility. Non-Python tools (like `curl` and `zip`) currently do not run within conda environments. As these tools are available on conda-forge, we should replace them with versions that are defined in a conda environment.

This _should_ allow to run the workflow on Windows (update README.md!), but I don't think we should officially support Windows.

timtroendle:
This increases portability of the workflow. Fixes #59.

I've tested this locally. The minimal workflow runs. The workflow uses libs from conda-forge, rather than my local tools.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,12 +6,12 @@
 * **ADD** schema that automatically validates configuration files (#45).
 * **ADD** automatic download of hydro basins data (#34).
 * **ADD** minimal configuration to be able to test the entire workflow more quickly (#60).
+* **ADD** installation of `curl` and `unzip` from conda-forge, to increase portability (#59).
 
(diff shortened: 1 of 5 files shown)


timtroendle:
As far as I can tell, the workflow still uses `cp`, `rm`, `touch`, and `echo`, which are all not installable from conda-forge. It also uses pipes, so this certainly won't work on Windows.

brynpickering:
If it won't work on windows, what form of portability has it added?

timtroendle:
Portability between (versions of) operating systems other than Windows. It ensures that the workflow uses the exact same software on all machines -- in the exact same way as all other conda envs.

brynpickering (approved):

PR #67, opened 2021-04-16
Add schema validation workflow

brynpickering:
I noticed that inconsistencies in the schema itself were being ignored, so this explicitly validates the config files as well as the schema itself (against the JSON draft 7 validator).


brynpickering:
I would like this to be squashed on merge, to hide my trial-and-error approach to getting the paths correct ;)

config/validate-schema.py
```diff
@@ -0,0 +1,16 @@
...
+    with open(configfile, "r") as f:
+        config = yaml.safe_load(f)
+
+    if len(sys.argv) > 2:
+        schemafile = sys.argv[2]
+        with open(schemafile, "r") as f:
+            schema = yaml.safe_load(f)
+        jsonschema.validate(config, schema)
+    else:
+        jsonschema.Draft7Validator.check_schema(config)
```
timtroendle, line 16: Flip the order of the two arguments. The first will consistently be the schema, and this line will make more sense, too:

`jsonschema.Draft7Validator.check_schema(schema)`

.github/workflows/schemavalidation.yaml
```diff
@@ -0,0 +1,24 @@
...
+            - name: Set up Python
+              uses: actions/setup-python@v2
+              with:
+                  python-version: 3.8
+            - name: Install dependencies
+              run: |
+                python -m pip install --upgrade pip
+                pip install jsonschema PyYAML
+            - name: Validate default config
+              run: python ./config/validate-schema.py ./config/default.yaml ./config/schema.yaml
```
timtroendle, line 20: I think the `py` file should live in `./tests/`, because that's what it is.

config/validate-schema.py
```diff
@@ -0,0 +1,16 @@
+import sys
+import jsonschema
+import yaml
+
+if __name__ == "__main__":
+    configfile = sys.argv[1]
```
timtroendle, line 6: A `click` or `argparse` cli would be nicer, but I guess this'll do.

timtroendle (approved):
Great!

Fine with merging this with the two comments tackled.

PR #70, opened 2021-04-19
Feature area weighted time series

timtroendle:
This moves some rather complicated code into `eurocalliopelib`. Apart from the new configuration parameter, this should not change anything to the user.

I implemented it because I need the algorithm in another project. It also includes preparation work to update the published capacity factor time series to a more reasonable format (one netcdf file with dimensions `time`, `x`, `y`).

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -188,8 +188,7 @@ rule capacity_factors_onshore_wind_and_solar:
     params:
         threshold = config["capacity-factors"]["min"],
         year = config["year"],
-        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"],
-        area_weighting_resolution_in_m = config["capacity-factors"]["area-weighting-resolution-in-m"]
+        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"]
     wildcard_constraints:
         technology = "((wind-onshore)|(rooftop-pv)|(open-field-pv)|(rooftop-pv-n)|(rooftop-pv-e-w)|(rooftop-pv-s-flat))"
     output: "build/model/{resolution}/capacityfactors-{technology}.csv"
@@ -208,8 +207,7 @@ rule capacity_factors_offshore:
     params:
         threshold = config["capacity-factors"]["min"],
         year = config["year"],
-        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"],
-        area_weighting_resolution_in_m = config["capacity-factors"]["area-weighting-resolution-in-m"]
+        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"]
     output: "build/model/{resolution}/capacityfactors-wind-offshore.csv"
     conda: "envs/geo.yaml"
     script: "scripts/capacityfactors_offshore.py"
--- a/Snakefile
+++ b/Snakefile
@@ -183,8 +183,8 @@ rule capacity_factors_onshore_wind_and_solar:
     input:
         script = script_dir + "capacityfactors.py",
         locations = rules.units.output[0],
-        ids = ancient("data/automatic/capacityfactors/onshore-locations.tif"),
-        timeseries = ancient("data/automatic/capacityfactors/{technology}-timeseries.nc")
+        timeseries = ancient("data/automatic/capacityfactors/{technology}-timeseries.nc"),
+        coordinates = ancient("data/automatic/capacityfactors/wind-onshore-timeseries.nc")
     params:
         threshold = config["capacity-factors"]["min"],
         year = config["year"],
@@ -203,7 +203,6 @@ rule capacity_factors_offshore:
         script = script_dir + "capacityfactors_offshore.py",
         eez = rules.eez.output[0],
         shared_coast = rules.potentials.output.shared_coast,
-        ids = ancient("data/automatic/capacityfactors/offshore-locations.tif"),
         timeseries = ancient("data/automatic/capacityfactors/wind-offshore-timeseries.nc")
     params:
         threshold = config["capacity-factors"]["min"],
--- a/Snakefile
+++ b/Snakefile
@@ -20,6 +20,7 @@ BIOFUEL_FEEDSTOCKS = [
 
 include: "./rules/shapes.smk"
 include: "./rules/hydro.smk"
+include: "./rules/sync.smk"
 localrules: all, raw_load, model, clean, parameterise_template, potentials_zipped
 localrules: download_capacity_factors_wind_and_solar
 configfile: "config/default.yaml"
--- a/Snakefile
+++ b/Snakefile
@@ -188,7 +188,8 @@ rule capacity_factors_onshore_wind_and_solar:
     params:
         threshold = config["capacity-factors"]["min"],
         year = config["year"],
-        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"]
+        trim_ts = config["capacity-factors"]["trim-ninja-timeseries"],
+        area_weighting_resolution_in_m = config["capacity-factors"]["area-weighting-resolution-in-m"]
     wildcard_constraints:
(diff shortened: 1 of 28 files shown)


Snakefile
```diff
@@ -183,12 +183,13 @@ rule capacity_factors_onshore_wind_and_solar:
     input:
         script = script_dir + "capacityfactors.py",
         locations = rules.units.output[0],
-        ids = ancient("data/automatic/capacityfactors/onshore-locations.tif"),
-        timeseries = ancient("data/automatic/capacityfactors/{technology}-timeseries.nc")
+        timeseries = ancient("data/automatic/capacityfactors/{technology}-timeseries.nc"),
+        coordinates = ancient("data/automatic/capacityfactors/wind-onshore-timeseries.nc")
```
timtroendle, line 187: This input is weird, but currently needed as some published timeseries do not contain coordinates.

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -0,0 +1,221 @@
...
+WGS84 = f"EPSG:{WGS84_EPSG}"
+EPSG3035 = "EPSG:3035"
+
+
+def area_weighted_time_series(shapes, spatiotemporal, resolution=None):
+    """Forms area weighted time series for collections of shapes.
+
+    Inputs:
+        * locations: a GeoDataFrame of shapes, each will receive one time series
+        * spatiotemporal: a DataArray with dimensions "x", "y", "timestep" in CRS WGS84
```
timtroendle, line 27: This is wrong: data can be in any CRS.

lib/setup.py
```diff
@@ -10,7 +10,20 @@
     maintainer_email='tim.troendle@usys.ethz.ch',
     packages=find_packages(exclude=['tests*']),
     include_package_data=True,
-    install_requires=["pycountry>=18.12.8"],
+    install_requires=[
+        "numpy",
```
timtroendle, line 14: I'm undecided as to whether we should define version numbers here. The dependencies of the entire workflow are defined in the envs of the workflow. Maybe that's enough?

We could of course be more restrictive here, but that would glue together the different envs that we have. That's certainly something we do not want.

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -0,0 +1,221 @@
...
+    X, Y = np.meshgrid(X, Y)
+    return interp(X, Y)
+
+
+def isclose(a, b):
+    return math.isclose(a, b, abs_tol=INDEX_EPSILON, rel_tol=0)
+
+
+def infer_resolution(spatiotemporal):
+    y_diffs = spatiotemporal.y.diff("y")
```
timtroendle, line 147: Must be absolute.

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -0,0 +1,221 @@
...
+    x_diffs = spatiotemporal.x.diff("x")
```
timtroendle, line 148: Must be absolute.

brynpickering (commented):
My first impression from this is that it's too complex for what it is. I don't see the point in creating the higher resolution grids when all you're using it for area weighting the coarser input, without any interpolation in the upscaling. Instead I would create a geodataframe of polygons representing the coarse grid (a few lines) and then do a spatial overlay of this grid with the shapes. Then the area of each resulting element will give you the area weighting for the timeseries data. This way you don't have to mess around with this other resolution at all. E.g.:

```python

grid_gdf = gpd.GeoSeries(gpd.points_from_xy(ts.x, ts.y), crs=crs, index=ts.site_id.to_index()).buffer(INPUT_GRID_SIZE_IN_M).envelope  # of course, you could infer GRID_SIZE_IN_M from the input data, as you currently do 
overlaid = gpd.overlay(gpd.GeoDataFrame(geometry=grid_gdf), shape)
overlaid['area_weighting'] = overlaid.area.div(grid_gdf.loc[overlaid.site_id].values)  # this is unnecessary if the gridcells are all the same size, as the overlaid areas can give you the area weighting directly
```
Then, for each shape, 'overlaid' would have a collection of site_id:area_weighting mappings to use.

lib/eurocalliopelib/geo/spatiotemporal.py
timtroendle: Yes, and these tests are already in place.

envs/hydro.yaml
```diff
@@ -18,4 +18,4 @@ dependencies:
     - atlite=0.2.1
     - pip=21.0.1
     - pip:
-        - -e ../../lib
+        - -e ../../lib[geo]
```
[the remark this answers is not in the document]
timtroendle: The dependencies _area_ dealt with within the conda env files. If you do everything correctly, pip won't need to install anything other than `eurocalliopelib` itself. I still think we should make dependencies of `eurocalliopelib` explicit.

The motivation to have the `extra_install` `geo` was not to _install_ dependencies, but to be able to _not install_ (geo) dependencies when they are not needed.

timtroendle:
Good point with using shapes rather than raster data. I thought about that back in 2018, but didn't find a good way. Within this PR, I simply copied the existing approach.

But I'd be happy to move to a polygon-based approach. The example you give will give produce circles, not squares, and therefore the polygons will either overlap or there will be gaps.

Probably we could generate squares with something like the following:
```Python
grid_gdf = gpd.GeoSeries(
     [square(center=point, size=resolution) for point in points]
)
```

I will look into this.

brynpickering:
> The example you give will give produce circles, not squares, and therefore the polygons will either overlap or there will be gaps.

Not so, the 'buffer' produces circles, then 'envelope' creates squares describing the limit of those circles.

timtroendle:
The `overlay`-hint was a good one @brynpickering. I think this is much cleaner now and should be as precise as possible. The only small drawback is that this approach is slower and uses more memory. But it's all reasonable, even with our largest files.

brynpickering (commented):
I suspect the memory footprint is caused by stacking/unstacking of the xarray. Is there any reason to have this newer ("x", "y", "time") structure and not stick with ("site_id", "time")?

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -0,0 +1,163 @@
...
+def weighted_time_series(weights_and_values):
+    ds = (
+        weights_and_values
+        .where(weights_and_values.weight > 0)
+        .dropna(subset=["weight"], dim="xy") # drop all locations with weight == 0
+    )
+    return (ds * ds.weight).value.sum("xy", skipna=False)
+
+
+def infer_resolution(spatiotemporal):
```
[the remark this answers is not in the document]
timtroendle: Hm but why? Do you think this is brittle?

Jan's data, for example, does not have a resolution attribute. That means I'd move this function from here to another repo. As long as we don't have reason to believe this is brittle, this is a rather convenient function I'd say.

timtroendle:
I moved towards the `x/y` structure because this is the most intuitive way I guess. I wanted to do this for a while already. Jan's data has this format, too. Plus, with the support for sparse matrices, this isn't an issue (I am aware I am not using them in here).

The larger memory footprint only partly comes from the unstacking. It's also because a polygon is larger than a raster point I guess. But really, I don't care much as this works without any problem for all of Europe on my local machine -- so no problem whatsoever on the cluster.

lib/eurocalliopelib/geo/spatiotemporal.py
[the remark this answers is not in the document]
brynpickering: Just that it is quite a verbose function for what it does, which is perhaps just due to the number of assertions. It can stay ;)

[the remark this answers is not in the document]
timtroendle: Are you fine with everything else?

brynpickering (approved):
Happy for this to go ahead. The only thing worth tracking in an issue beyond this PR is the `rooftop-pv` dataset needing another dataset to give it coordinates.

PR #75, opened 2021-04-24
Add cluster sync infrastructure

### LINKED ISSUE #74: Add cluster sync infrastructure
A set of rules (send/receive/clean) that can send files to, and receive build files from a cluster.

This should roughly look like [these rules](https://github.com/timtroendle/money-land/blob/master/rules/sync.smk). All constants should be configuration values. Instead of "Euler, call it "cluster" and it should be rather generic then.

Sub workflows can then use the rules with the new module mechanism of Snakemake:

```Python
# Snakefile

module eurocalliope:
    snakefile: "euro-calliope/Snakefile"

use rule send from eurocalliope
use rule receive from eurocalliope
use rule clean from eurocalliope
```

timtroendle:
Fixes #74.

### DIFF
--- a/.syncignore-receive
+++ b/.syncignore-receive
@@ -1,0 +1,2 @@
+# build files to ignore when receiving from cluster
+build/data/*.nc
--- a/.syncignore-send
+++ b/.syncignore-send
@@ -1,0 +1,10 @@
+.snakemake
+.git
+.pytest_cache
+__pycache__
+*.pyc
+*.pyo
+.vscode
+.DS_Store
+build
+data/automatic
--- a/.syncignore-send
+++ b/.syncignore-send
@@ -8,3 +8,4 @@ __pycache__
 .DS_Store
 build
 data/automatic
+notebooks
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -12,6 +12,12 @@ data-sources:
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
     potentials: https://zenodo.org/record/3533038/files/possibility-for-electricity-autarky.zip
 root-directory: . # point to the root directory if working directory is not root directory
+cluster-sync:
+    url: euler.ethz.ch
+    send-ignore: .syncignore-send
+    receive-ignore: .syncignore-receive
+    cluster-base-dir: ~/Develop/euro-calliope/
+    local-results-dir: build/euler
 scaling-factors: # values are tuned for models with a few hours resolution and one year duration
     power: 0.00001 # from MW(h) to 100 GW(h)
     area: 0.0001 # from km2 to 10,000 km2
@@ -44,7 +50,6 @@ parameters:
         potential-year: "2020"
         cost-year: "2030"
     biofuel-efficiency: 0.45
-
 quality-control:
     load:
         acceptable-load-data-gap-hours: 48  # some load data is patchy and we fill it with other years' data if the missing data in a given year is above this many hours.
@@ -57,7 +62,6 @@ quality-control:
             - actual_entsoe_power_statistics
     hydro:
         scale-phs-according-to-geth-et-al: false
-
 sea-connections:
     continental:
     national: # Source: https://www.entsoe.eu/data/map/
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -62,7 +62,7 @@ properties:
                 description: URL of the cluster.
             send-ignore:
                 type: string
-                descirption: Path to local file defining files and folders to ignore when sending to cluster (relative to workdir).
+                description: Path to local file defining files and folders to ignore when sending to cluster (relative to workdir).
             receive-ignore:
                 type: string
                 description: Path to local file defining files and folders to ignore when receiving from cluster (relative to workdir).
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -54,6 +54,24 @@ properties:
     root-directory:
         type: string
         description: Path to the root directory of euro-calliope containing scripts and template folders.
+    cluster-sync:
+        type: object
+        properties:
+            url:
+                type: string
+                description: URL of the cluster.
+            send-ignore:
+                type: string
+                descirption: Path to local file defining files and folders to ignore when sending to cluster.
+            receive-ignore:
+                type: string
+                description: Path to local file defining files and folders to ignore when receiving from cluster.
+            cluster-base-dir:
+                type: string
+                description: Path on cluster into which files are sent from local machine (must exist).
+            local-results-dir:
+                type: string
+                description: Path to local folder into which results from cluster are downloaded.
     quality-control:
         type: object
         properties:
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -62,16 +62,16 @@ properties:
                 description: URL of the cluster.
             send-ignore:
                 type: string
-                descirption: Path to local file defining files and folders to ignore when sending to cluster.
+                descirption: Path to local file defining files and folders to ignore when sending to cluster (relative to workdir).
             receive-ignore:
                 type: string
-                description: Path to local file defining files and folders to ignore when receiving from cluster.
+                description: Path to local file defining files and folders to ignore when receiving from cluster (relative to workdir).
             cluster-base-dir:
                 type: string
                 description: Path on cluster into which files are sent from local machine (must exist).
             local-results-dir:
                 type: string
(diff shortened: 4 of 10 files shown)


config/default.yaml
```diff
@@ -12,6 +12,12 @@ data-sources:
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
     potentials: https://zenodo.org/record/3533038/files/possibility-for-electricity-autarky.zip
 root-directory: . # point to the root directory if working directory is not root directory
+cluster-sync:
```
brynpickering, line 15: Could these be in a separate file? I can imagine wanting to set these values in my development branch of euro-calliope so that I can run tests on the cluster and would prefer to be able to then have these config options in a file that I can easily edit without staging for commit.

.syncignore-send
```diff
@@ -0,0 +1,10 @@
+.snakemake
```
brynpickering, line 1: This looks like a list that's quite bespoke to your device. I wonder if there's a more generic list that can be used (e.g. all files and directories starting with `.`)?

.syncignore-receive
```diff
@@ -0,0 +1,2 @@
+# build files to ignore when receiving from cluster
+build/data/*.nc
```
brynpickering, line 2: any reason why?

config/default.yaml
```diff
@@ -12,6 +12,12 @@ data-sources:
...
+    url: euler.ethz.ch
+    send-ignore: .syncignore-send
+    receive-ignore: .syncignore-receive
+    cluster-base-dir: ~/Develop/euro-calliope/
+    local-results-dir: build/euler
```
brynpickering, line 20: is this relative to your local snakemake root directory?

rules/sync.smk
```diff
@@ -0,0 +1,41 @@
...
+rule send:
+    message: "Send changes to cluster"
+    params:
+        send_ignore = config["cluster-sync"]["send-ignore"],
+        url = config["cluster-sync"]["url"],
+        cluster_base_dir = config["cluster-sync"]["cluster-base-dir"]
+    conda: "../envs/shell.yaml"
+    shell:
+        """
+        rsync -avzh --progress --delete -r . --exclude-from={params.send_ignore} \
```
brynpickering, line 13: could the 'delete' argument be dangerous? E.g. if you have notebooks on the cluster that you're using to do tests, would these not be deleted?

brynpickering (commented):
Good idea to generalise this, just a few comments to address.

config/default.yaml
timtroendle: Hmm, there is nothing that holds you back doing so. You can either use another config file that overrides these defaults, or you can override them on the command line. I don't really see how a separate (checked-in) file would improve. Let me know if I'm missing something here.

.syncignore-send
timtroendle: I see your point. I cannot come up with a general rule, however. Besides, there are only two entries here that are specific (`vscode` is VS Code-specific, `.DS_Store` is macOS-specific), all others are generic.

This file simply mimics `.gitignore` which also does not rely on general rules.

.syncignore-receive
timtroendle: They are large and usually not needed (at least for me). I agree this is rather arbitrary. I decided to keep it as an example of how this works, but I'm fine removing it as well.

config/default.yaml
timtroendle: Yes.

rules/sync.smk
timtroendle: Yes, that's what `--delete` does. To me, that's totally fine. That's the whole point of all this: you don't need to work on the cluster at all (other than executing Snakemake).

This could be a configuration option but I don't see myself using it in the future. Would you prefer this being a config option?

.syncignore-send
brynpickering: ok, sure

config/default.yaml
timtroendle: This is consistent with `send-ignore` and `receive-ignore`. Do you think this needs more explanation?

config/default.yaml
brynpickering: I suppose it's just the fact that all other defaults are valid fall-backs that will work for anyone. These are examples of the use of `cluster-sync`, but nobody will have the same values for their own implementation (certainly for 'cluster-base-dir' and often for 'url'), so it feels odd to have them in 'defaults'. I will set them to something different straight away and possibly have that in a separate config file, but I then need to remember to always load that in when I run snakemake since I can't add it to the `Snakefile` without having to revert those changes on each commit.

rules/sync.smk
brynpickering: during testing, if I rely on the larger input datasets, I prefer to sync minor updates to the cluster and run snakemake with those in place. Looking at the syncignore list though, one of my concerns that it would override a partial 'build' directory is unfounded, since that is ignored. However, I use notebooks extensively to test larger datasets and build results - often this is only possible on the cluster, where those datasets reside. I'm happy to move my notebooks out of the workflow directory if it is considered too bad practice ;)

config/default.yaml
timtroendle: I understand and agree. What do you suggest instead?

One think you _could_ do is add this to each of your Snakemake calls (I do these things in a fish function):

`snakemake --configfile bryns-cluster-settings.yaml`

E.g. in fish:

```
function smake --description "shortcut for snakemake"
    command snakemake --use-conda --cores 1 --configfile bryns-cluster-settings.yaml $argv
end
```

config/default.yaml
brynpickering: Can you add that piece of information into the schema descriptions? When we get those parsed in some publicly accessible way, that would be sufficient explanation.

rules/sync.smk
timtroendle: I would not say notebooks in the workflow directory are bad practise. I use them myself every now and then. I, however, would always run them locally, not on the cluster.

I personally would never remove the `--delete` option. The reason is: I want to be 100% certain the Euler repo mirrors my local repo. When you delete files locally and those don't get deleted on Euler, repos get out of sync. There are corner cases where that can cause issues and that's why I introduced the `--delete` flag.

We can always make the `--delete` flag optional if we see need for that. I do not at this point, but if you want, I'll add it.

brynpickering: I never run notebooks locally, only on the cluster :P Should we just add notebooks to the ignore list? They are either run locally or on the cluster, either way it doesn't matter to either of us if they are synced with the other location. then `--delete` can stay. I agree that not having it can cause issues.

timtroendle: Good idea!

config/default.yaml
timtroendle: Another -- very simple -- solution: Make ´~/Develop´ a symbol link to the directory in which you keep your projects on Euler (I do the same). I am also happy to agree on another name for that symbolic link if you absolutely prefer another.

That does not solve the solution for users not using Euler. These users can use the approach I described above, and they anyway have to do other changes (they need a cluster profile).

brynpickering: hmm, could you have a 'sync.yaml' config file in the `confg/euler` sub-directory, and then easily point to that? I just like the idea of keeping the euler-specific stuff in one place. If you think this hinders maintainability, then I'm OK with ignoring this discussion chain entirely.

timtroendle: > hmm, could you have a 'sync.yaml' config file in the confg/euler sub-directory, and then easily point to that?

I see your point here. Still, it wouldn't change the fact that this config is checked-in and hardwired.

I guess the cleanest solution would be to remove the default values and require the user to provide the cluster config; through a config file or through the command line (possibly only in case the user tries to sync to a cluster). That would make the entire mechanism more complicated and would require more work on our side though. Right now, that doesn't seem time worth spent to me -- all (three?) users run on Euler anyways.

brynpickering: OK, let's stick to it being in the defaults.

config/schema.yaml
```diff
@@ -62,7 +62,7 @@ properties:
                 description: URL of the cluster.
             send-ignore:
                 type: string
-                descirption: Path to local file defining files and folders to ignore when sending to cluster (relative to workdir).
+                description: Path to local file defining files and folders to ignore when sending to cluster (relative to workdir).
```
timtroendle, line 65: Our schema validation unfortunately didn't cover this nasty typo, @brynpickering . Also Snakemake wasn't complaining.

Issue #66, opened 2021-04-16
More flexible and powerful gap-filling of national electricity load data


timtroendle:
Electricity load time series are based on national electricity load time series from Entso-e. The source data has many gaps: sometime single missing values, sometimes several hours or days are missing, sometimes entire years are missing for certain countries.

As the entso-e dataset contains several data sources and several years, we can fill gaps by taking data from other data sources or other years. This approach can lead to unwanted consequences and should only happen if explicitly demanded by the user. Per default, we interpolate only up to a few hours. If gaps are longer than that, euro-calliope will fail but let the user know that it is possible to use other data sources and other years.

timtroendle:
The inclusion of other data sets is discussed in #41 and implemented in #42.

timtroendle:
The use of other data from other years in implemented in #3.

brynpickering:
This should also include a limit on how many hours data can be interpolated over before it just gets out of hand. This is not implemented in any way right now, but mentioned in a later comment by @timtroendle in #3.

timtroendle:
@brynpickering: `develop` does not run anymore with the default configuration. That's because of missing data for Albania. It fails with:

`AssertionError: Not enough load data for country ALB`

I understand this means there is not enough data, neither in `actual_entsoe_transparency` nor in `actual_entsoe_power_statistics` (both are in the default configuration). `euro-calliope 1.0` used to take 2017 data for Albania. I guess I still don't fully understand the gap-filling approach. Why is it not using 2017 data?

timtroendle:
I moved this ^ to its own issue (#71).

PR #89, opened 2021-04-26
Hotfix national load

### LINKED ISSUE #71: Gap filling of load time series does not work for the default configuration
In the current default configuration, load time series from 2016 is used for all countries in Europe. We have load data of Albania (ALB) only starting from 2017 and thus, `euro-calliope 1.0` used to use 2017 data for Albania.

The current gap filling method should do the same (as I understand it), but it does not do it. Therefore, `develop` is currently broken with the following error message: 

``AssertionError: Not enough load data for country ALB``

brynpickering:
This should do the trick to fix #71. It is not a change in the data processing, only in the assertion that the data processing has done what was expected.

This will perform slightly differently to euro-calliope v1.0 as it does not attempt to fill 29th of Feb with data from another year for Albania (it will get it by interpolation), whereas I think the old one filled that day with the values from the day before/after. This 29th of Feb issue will be dealt with in #66 more generally.

### DIFF
--- a/scripts/national_load.py
+++ b/scripts/national_load.py
@@ -58,7 +58,7 @@ def _get_index_of_missing_data(series, model_year):
     return series[(series < 0) | series.isna()][str(model_year)].index
 
 
-def _handle_feb_29th(this_year, next_avail_year, missing_timesteps):
+def _ignore_feb_29th(this_year, next_avail_year, missing_timesteps):
     """ Ignore February 29th if next available year doesn't have that data available """
     if (pd.Period(freq='Y', year=this_year).is_leap_year and
             not pd.Period(freq='Y', year=next_avail_year).is_leap_year and
@@ -78,7 +78,7 @@ def fill_missing_data_in_country(country_series, model_year, acceptable_gap_hour
     # We're OK with some gap, as defined by 'acceptable_gap_hours'
     years_with_data = country_series.dropna().index.year.drop(model_year, errors="ignore").unique()
 
-    def _get_year_order(years):
+    def _give_older_years_lower_priority(years):
         year_order = []
         for year in years:
             if year - model_year < 0:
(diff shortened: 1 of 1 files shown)


scripts/national_load.py
```diff
@@ -58,7 +58,7 @@ def _get_index_of_missing_data(series, model_year):
     return series[(series < 0) | series.isna()][str(model_year)].index
 
 
-def _handle_feb_29th(this_year, next_avail_year, missing_timesteps):
+def _ignore_feb_29th(this_year, next_avail_year, missing_timesteps):
     """ Ignore February 29th if next available year doesn't have that data available """
     if (pd.Period(freq='Y', year=this_year).is_leap_year and
```
timtroendle, line 63: You could also use `calendar` here:

`calendar.isleap(year)`

scripts/national_load.py
```diff
@@ -78,31 +78,30 @@ def fill_missing_data_in_country(country_series, model_year, acceptable_gap_hour
...
                 year_order.append(abs(year - model_year) + 0.1)
             else:
                 year_order.append(year - model_year)
         return year_order
-    preferred_order_of_years_with_data = list(years_with_data.sort_values(key=_get_year_order, ascending=False))
+    preferred_order_of_years_with_data = list(years_with_data.sort_values(key=_give_older_years_lower_priority, ascending=False))
 
-    while country_series.loc[all_missing_timesteps].isnull().sum() > acceptable_gap_hours:
+    _missing_timesteps = all_missing_timesteps.copy()
+    while len(_missing_timesteps) > acceptable_gap_hours:
```
timtroendle, line 92: Rather combine the two conditions:

```Python
while (len(_missing_timesteps) > acceptable_gap_hours) & (preferred_order_of_years_with_data != []):
    do something
```    

This should also work but isn't more readable I think:

```Python
while (len(_missing_timesteps) > acceptable_gap_hours) & preferred_order_of_years_with_data:
    do something
```

timtroendle (approved):
I actually have no clue why this works. But if you say it does, I am happy with it. :)

scripts/national_load.py
brynpickering: neat

PR #90, opened 2021-04-26
Make schema checker strict to typos

brynpickering:
The JSON schema checker ignores typos, which seems to me to a big oversight (one which caught us out in #75). This change is a bit of a hack as I have not found an approach to make the metaschema strict with an explicit argument. Nonetheless, it's quite a transparent hack: it just ensures no additional 'properties' are allowed in the schema other than those defined in the schema draft. On testing it locally, it seems to be effective.


timtroendle (approved):
Nice!

PR #95, opened 2021-04-28
Move unit filter to lib

timtroendle:


brynpickering (approved):
Is `filters` following some snakemake syntax? It doesn't seem to actually make sense in terms of what it does. Otherwise, all looks good!

timtroendle:
It's `jinja2` syntax.

PR #96, opened 2021-04-29
Update rules to make downloads consistent

timtroendle:
* clearly mark rules that download data
* only rules that download data fiddle with data/automatic folder


timtroendle:
I deleted the automatic downloads and tested this with the minimal workflow.

brynpickering (approved):
Makes sense. I guess the one glaring thing is that (as with all other config options), these files won't be changed if you change the data source in config...

PR #99, opened 2021-05-04
Update EEZ data source to enable direct download in the workflow

brynpickering:
Would aid #56 and simplify everything by removing the remaining manual download requirement for the workflow. It has the option to be enhanced by specifying only the countries of interest in the API request (`https://geo.vliz.be/geoserver/MarineRegions/wfs?service=WFS&version=1.0.0&request=DescribeFeatureType&typeNames=MarineRegions:eez&cql_filter=territory1=...&outputFormat=SHAPE-ZIP`) but perhaps it is unnecessary as the filesize is relatively small.


rules/shapes.smk
```diff
@@ -107,17 +107,28 @@ rule units_without_shape:
...
-        countries=",".join(["'{}'".format(country) for country in config["scope"]["countries"]])
+        countries=",".join(["'{}'".format(country) for country in config["scope"]["countries"]]),
+        temp_dir = "build/data/eez"
     conda: "../envs/geo.yaml"
     shell:
         """
-        fio cat --bbox {params.bounds} {input}\
-        | fio filter "f.properties.Territory1 in [{params.countries}]"\
+        unzip -o {input} -d {params.temp_dir}
+        fio cat --bbox {params.bounds} {params.temp_dir}/eez.shp\
```
timtroendle, line 130: Fiona can (in principle) read zip files directly, without unpacking. Not sure the version we are using supports that, but would be good to give it a try. This rule would be much cleaner.

> Zip and Tar archives can be treated as virtual filesystems and Collections can be made from paths and layers within them. In other words, Fiona lets you read and write zipped Shapefiles.

timtroendle: https://fiona.readthedocs.io/en/latest/README.html#collections-from-archives-and-virtual-file-systems

timtroendle: I like that feature quite a lot. I dislike the fact that a "shapefile" is actually a collection of files.

rules/shapes.smk
```diff
@@ -107,17 +107,28 @@ rule units_without_shape:
...
+        | fio filter "f.properties.territory1 in [{params.countries}]"\
```
timtroendle, line 131: This seems that the dataset is not the same. Probably, it's a newer version? What can we say about that? The old version was v10.

timtroendle (approved):
That's great. How come this is possible suddenly? Was this always possible?

My main concern is the dataset seems to be not exactly the same. What do we know about the differences? That info should be in the changelog.

rules/shapes.smk
brynpickering: In fact I looked into this for this very reason: v10 is no longer available at the source (it is v11 only). As with protected areas, it seems they scrub older versions when they add a new one. Not sure there's any way to pin this except to do the processing ourselves and then upload it to zenodo...

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
 * **ADD** sync infrastructure to easily send and receive files to and from a cluster (#74).
 * **ADD** optional email notifications whenever builds fail or succeed (#92).
 
-* **UPDATE** ENTSOE national electricity load data gap filling priority order included in config (#42)
+* **UPDATE** EEZ is now be downloaded within the workflow, rather than depending on a manual download.
```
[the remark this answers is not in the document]
brynpickering: ha, well since the data is supposed to be the same and we're instead changing the sources, I thought of it as an update :P I've no strong opinion on it, but agree it should be consistent.

rules/shapes.smk
timtroendle: Fortunately, that's not correct: v10 is still available (under "Older version").

Still, my main point remains: what do we know about this other version? Is it v11, meaning that it's simply a newer (better!) version?

CHANGELOG.md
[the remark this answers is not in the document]
timtroendle: As long as we are only adding the infastructure to download automatically, it sounds like an ADD to me. If we are changing the dataset, it sounds more like an UPDATE to me.

rules/shapes.smk
brynpickering: it is v11, and the differences are very minor:

<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>v10_area</th>
      <th>v11_area</th>
      <th>diff</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>Albania</th>
      <td>12150</td>
      <td>12165</td>
      <td>1.001235</td>
    </tr>
    <tr>
      <th>Belgium</th>
      <td>3482</td>
      <td>3495</td>
      <td>1.003733</td>
    </tr>
    <tr>
      <th>Bosnia and Herzegovina</th>
      <td>13</td>
      <td>13</td>
      <td>1.000000</td>
    </tr>
    <tr>
      <th>Bulgaria</th>
      <td>34685</td>
      <td>34745</td>
      <td>1.001730</td>
    </tr>
    <tr>
      <th>Croatia</th>
      <td>55399</td>
      <td>55502</td>
      <td>1.001859</td>
    </tr>
    <tr>
      <th>Cyprus</th>
      <td>98468</td>
      <td>98450</td>
      <td>0.999817</td>
    </tr>
    <tr>
      <th>Denmark</th>
      <td>104521</td>
      <td>105021</td>
      <td>1.004784</td>
    </tr>
    <tr>
      <th>Estonia</th>
      <td>36257</td>
      <td>36451</td>
      <td>1.005351</td>
    </tr>
    <tr>
      <th>Finland</th>
      <td>81072</td>
      <td>81553</td>
      <td>1.005933</td>
    </tr>
    <tr>
      <th>France</th>
      <td>347315</td>
      <td>348166</td>
      <td>1.002450</td>
    </tr>
    <tr>
      <th>Germany</th>
      <td>56512</td>
      <td>56763</td>
      <td>1.004442</td>
    </tr>
    <tr>
      <th>Greece</th>
      <td>482784</td>
      <td>482910</td>
      <td>1.000261</td>
    </tr>
    <tr>
      <th>Ireland</th>
      <td>425346</td>
      <td>427039</td>
      <td>1.003980</td>
    </tr>
    <tr>
      <th>Italy</th>
      <td>536134</td>
      <td>536654</td>
      <td>1.000970</td>
    </tr>
    <tr>
      <th>Latvia</th>
      <td>28212</td>
      <td>28353</td>
      <td>1.004998</td>
    </tr>
    <tr>
      <th>Lithuania</th>
      <td>6800</td>
      <td>6832</td>
      <td>1.004706</td>
    </tr>
    <tr>
      <th>Montenegro</th>
      <td>6374</td>
      <td>6383</td>
      <td>1.001412</td>
    </tr>
    <tr>
      <th>Netherlands</th>
      <td>64058</td>
      <td>64328</td>
      <td>1.004215</td>
    </tr>
    <tr>
      <th>Norway</th>
      <td>926772</td>
      <td>933167</td>
      <td>1.006900</td>
    </tr>
    <tr>
      <th>Poland</th>
      <td>29847</td>
      <td>29982</td>
      <td>1.004523</td>
    </tr>
    <tr>
      <th>Portugal</th>
      <td>315287</td>
      <td>315501</td>
      <td>1.000679</td>
    </tr>
    <tr>
      <th>Romania</th>
      <td>29545</td>
      <td>29606</td>
      <td>1.002065</td>
    </tr>
    <tr>
      <th>Slovenia</th>
      <td>214</td>
      <td>214</td>
      <td>1.000000</td>
    </tr>
    <tr>
      <th>Spain</th>
      <td>560957</td>
      <td>561763</td>
      <td>1.001437</td>
    </tr>
    <tr>
      <th>Sweden</th>
      <td>154510</td>
      <td>155347</td>
      <td>1.005417</td>
    </tr>
    <tr>
      <th>United Kingdom</th>
      <td>737092</td>
      <td>739320</td>
      <td>1.003023</td>
    </tr>
  </tbody>
</table>

timtroendle: Perfect, can you mention this in the changelog? (Update from v10 to v11)

rules/shapes.smk
brynpickering: I take that back, on updating the syntax, fiona CLI can handle zips. I would still look to do all this in future with the `renewablepotentialslib` shape utils.

brynpickering:
So this cannot be merged until `solar-and-wind-potentials` is also updated (see https://github.com/calliope-project/solar-and-wind-potentials/issues/22), since there is now a mismatch in IDs.

timtroendle:
There two alternatives, so that we don't have to delay that. One would be to copy the code into this repository. Not perfect of course, but pragmatic. [The code to determine shared coasts is neither long nor complicated](https://github.com/calliope-project/solar-and-wind-potentials/blob/main/scripts/shared_coast.py). We could use that chance to rethink the algorithm. The old takes a long time to compute, and maybe we can improve that here.

The other would be to provide a map between the two ids here.

brynpickering:
If updating this shared coast code for speed, I would do something like the following:

```python
intersections = {}
areas = {}
# This can be multithreaded
for geom_idx, geom_data in units.iterrows():
    for eez_idx, eez_data in eez_new.iterrows():
        # Speed up the process by only considering EEZs where the country is involved in some way
        if geom_data.country_code.isin(eez_data[["iso_ter1", "iso_ter2", "iso_ter3"]]):
            # The EEZ buffer ensures that intersections are definitely covered, even when the two shapefiles don't perfectly align.
            # The exact buffer could/should be configurable
            if geom_data.geometry.intersects(eez_data.geometry.buffer(0.1)):
                # Intersection length and buffered area are both possible valid measures, I'd only pick one in the end
                # The buffer here is more restricted.
                intersections[(geom_data.id, eez_data.mrgid)] = geom_data.geometry.intersection(eez_data.geometry.buffer(0.01)).length
                area[(geom_data.id, eez_data.mrgid)] = geom_data.geometry.intersection(eez_data.geometry.buffer(0.01)).area
intersections_series = pd.Series(intersections)
intersections_normalised = intersections_series.div(intersections_series.sum(level=1))
```

timtroendle:
@brynpickering, I'd like to merge this soon instead of waiting for custom shapes. What do you think of fixing the ID in `solar-and-wind-potentials` and release that data on Zenodo? It's basically a tiny fix in https://zenodo.org/record/3533038.

brynpickering:
@timtroendle see https://github.com/calliope-project/solar-and-wind-potentials/pull/23

brynpickering:
@timtroendle would you be able to update zenodo so that this PR can be updated to point to the correct EEZ IDs?

brynpickering:
@timtroendle I've tested this with the minimal workflow locally without issue. This should be good to go.

timtroendle:
@brynpickering, did you run the entire minimal workflow including tests? I let them run just now to analyse the difference this makes to the capacity factors and the tests fail. In fact, all offshore wind capacity factors are 0. Is that not the case for you?

timtroendle:
Forget what I said, I forgot to force a re-download of the data.

timtroendle:
Great work removing the final manual download!

PR #111, opened 2021-05-05
Update config schema to cover all config items.

FebinKa:
Transfer descriptors from default.yaml to schema.yaml inside config/.

Co-authored-by: FebinKa <FebinKa@users.noreply.github.com>


config/default.yaml
```diff
@@ -34,25 +34,25 @@ capacity-factors:
 year: 2016
 crs: "EPSG:4326"
 parameters:
-    maximum-installable-power-density: # this is not the yield, but the density of installed power
-        pv-on-tilted-roofs: 160 # (MW/km^2) from [@Gagnon:2016][@Klauser:2016], i.e. 16% efficiency
-        pv-on-flat-areas: 80 # (MW/km^2) from [@Gagnon:2016][@Klauser:2016][@Wirth:2017]
-        onshore-wind: 8 # (MW/km^2) from [@EuropeanEnvironmentAgency:2009]
-        offshore-wind: 15 # (MW/km^2) from [@EuropeanEnvironmentAgency:2009]
+    maximum-installable-power-density:
```
brynpickering, line 37: I'm not happy about this (test)

config/schema.yaml
```diff
@@ -104,11 +246,84 @@ properties:
...
+        type: string
+        pattern: ^EPSG:\d+$
+        description: Coordinate reference system to which all geographic data will be conformed
+    sea-connections:
+        type: object
+        propertyNames:
+            enum: [continental, national, regional]
+        additionalProperties: false
+        patternProperties:
+            ^.*$:
```
suvayu, line 305: I'm not so sure about these.  It allows for an empty string, so don't really know what it is validating.  Also from what I can tell by reading the [spec](https://json-schema.org/draft/2020-12/json-schema-core.html#rfc.section.10.3.2), the values of the properties should match this pattern, but as far as I get, this is the structure that is being described:

    sea-connections:
        continental:
          - [src1, dst1]
          - [src2, dst2]
         national:
          ...

The values are arrays, so not sure how `patternProperties` does anything.

config/schema.yaml
```diff
@@ -104,11 +246,84 @@ properties:
...
+                    items:
+                        type: string
+                        description: Region code
+    shapes:
+        type: object
+        propertyNames:
+            enum: [continental, national, regional]
+        additionalProperties: false
+        patternProperties:
+            ^.*$:
```
suvayu, line 323: This makes more sense, but still it matches an empty string

config/schema.yaml
```diff
@@ -104,11 +246,84 @@ properties:
...
+                type: object
+                propertyNames: {"$ref": "#/properties/scope/properties/countries/items"}
+                patternProperties:
+                    ^.*$:
```
suvayu, line 327: Same as above, matches an empty string

suvayu (commented):
Disclaimer: I'm not very well versed in this kind of meta specs :-p

Looks fine, except for the`patternProperties` comments (which could also be misunderstanding on my part :-p )

config/schema.yaml
brynpickering: This is taken from the top answer [here](https://stackoverflow.com/questions/52897781/unclear-about-the-meaning-of-propertynames). `propertyNames` + `patternProperties` allows one to force the schema for 'continental', 'national', 'regional' to be the same, without needing to duplicate the schema for each property. I'm not sure this is the *best* way to do it, but it works ;)

config/schema.yaml
brynpickering: It doesn't really matter what it matches, since 'propertyNames' then sets an allowed list of items that the name can be, I should add 'additionalProperties: false' here, though.

config/schema.yaml
brynpickering: the property pattern could also be `""` as given in the SO answer, but I prefer something a bit more verbose. Happy to change it to `""` or `"^[a-zA-Z]+.*$"`, do you have a preference?

suvayu: Okay, I understand it now.  The language in the spec for `patternProperties` is a bit weird.  Anyway, for the pattern I would prefer something that's explicit and easier to read.  Since the meaning is "all properties", as the allowed properties are determined by `propertyName` and `additionalProperties: false`, I would say `""` is clearer.

suvayu (approved):

config/schema.yaml
brynpickering: > This is taken from the top answer [here](https://stackoverflow.com/questions/52897781/unclear-about-the-meaning-of-propertynames).

This was just wrong (maybe matched a previous schema draft?). Anyway, since the vlaidation failed with `""` I went instead with `"^[a-zA-Z]+.*$"`.

suvayu:
On Wed, 5 May, 2021, 22:13 Bryn Pickering, ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In config/schema.yaml
> <https://github.com/calliope-project/euro-calliope/pull/111#discussion_r626866391>
> :
>
> > +                type: boolean
> +                description: If true, trims renewables.ninja timeseries to the year in question
> +    year:
> +        type: integer
> +        description: Model year, which will be used to subset data during processing
> +    crs:
> +        type: string
> +        pattern: ^EPSG:\d+$
> +        description: Coordinate reference system to which all geographic data will be conformed
> +    sea-connections:
> +        type: object
> +        propertyNames:
> +            enum: [continental, national, regional]
> +        additionalProperties: false
> +        patternProperties:
> +            ^.*$:
>
> This is taken from the top answer here
> <https://stackoverflow.com/questions/52897781/unclear-about-the-meaning-of-propertynames>
> .
>
> This was just wrong (maybe matched a previous schema draft?). Anyway,
> since the vlaidation failed with "" I went instead with "^[a-zA-Z]+.*$".
>

Strange, I tried it locally with fastjsonschema before suggesting it. Maybe
I forgot to save the file before running it O_o

brynpickering:
hmm. Maybe you can try again at some point. there should be no reason for it to work with fastjsonschema and not jsonschema....

suvayu:
With the following:

```
diff --git a/config/schema.yaml b/config/schema.yaml
index 469aaf6..42f4b8f 100644
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -301,7 +301,7 @@ properties:
         propertyNames: {"$ref": "#/properties/shapes/propertyNames"}
         additionalProperties: false
         patternProperties:
-            ^.*$:
+            "":
                 type: array
                 description: Sea connections between regions at each resolution, based on e.g. https://www.entsoe.eu/data/map/
                 items:
@@ -319,11 +319,11 @@ properties:
             enum: [continental, national, regional]
         additionalProperties: false
         patternProperties:
-            ^.*$:
+            "":
                 type: object
                 propertyNames: {"$ref": "#/properties/scope/properties/countries/items"}
                 additionalProperties: false
                 patternProperties:
-                    ^.*$:
+                    "":
                         type: string
```
I get this:
```
In [1]: import yaml
In [2]: import fastjsonschema
In [3]: definition = yaml.safe_load(Path("schema.yaml").read_text())
In [4]: data = yaml.safe_load(Path("default.yaml").read_text())
In [5]: data == fastjsonschema.validate(definition, data)
Out[5]: True
```
:shrug:

brynpickering:
weird. Can fastjson schema check the schema itself against the draft? if so, we could just move to using it instead of jsonschema.

suvayu:
I'm not quite sure.  There's no dedicated method.  So I deliberately fed it a bad definition (the last `patternProperties` was misspelled), that led to the following:
```
In [14]: fastjsonschema.validate(baddefinition, data)
---------------------------------------------------------------------------
JsonSchemaValueException                  Traceback (most recent call last)
<ipython-input-14-0e093e1ca376> in <module>
----> 1 fastjsonschema.validate(baddefinition, data)

~/.virtualenvs/py39/lib/python3.9/site-packages/fastjsonschema/__init__.py in validate(definition, data, handlers, formats, use_default)
    111     Preferred is to use :any:`compile` function.
    112     """
--> 113     return compile(definition, handlers, formats, use_default)(data)
    114 
    115 

<string> in validate(data)

JsonSchemaValueException: data.shapes.continental must not contain {'Belgium', 'Estonia', 'Austria', 'Bosnia and Herzegovina', 'Hungary', 'Italy', 'Slovenia', 'Lithuania', 'Croatia', 'Romania', 'Slovakia', 'Switzerland', 'Serbia', 'France', 'Greece', 'United Kingdom', 'Macedonia, Republic of', 'Germany', 'Poland', 'Cyprus', 'Bulgaria', 'Denmark', 'Finland', 'Albania', 'Czech Republic', 'Netherlands', 'Sweden', 'Ireland', 'Spain', 'Portugal', 'Montenegro', 'Norway', 'Luxembourg', 'Latvia'} properties
```
If I try to compile with the bad json, I get back a compiled function without errors, but validation fails
```
In [7]: bad_compiled = fastjsonschema.compile(baddefinition)
In [13]: bad_compiled(data)
---------------------------------------------------------------------------
JsonSchemaValueException                  Traceback (most recent call last)
<ipython-input-13-45fedb56266d> in <module>
----> 1 bad_compiled(data)

<string> in validate(data)

JsonSchemaValueException: ...
```
Maybe that key can be misspelled as it is optional.  I tried misspelling one of the types: `object` -> `ob-ject`, that led to a `JsonSchemaDefinitionException` during the compilation phase.  So it seems it does verify, but I'm not quite clear how much flexibility JSON schema itself allows.  I have the feeling `jsonschema` would also let the first one pass (misspelled optional key).

brynpickering:
So we can cover both misspelled keys and values with jsonschema, so I think we'll stick with that. But, since jsonschema is being weird with  `""`, I'll stick with `"^[a-zA-Z]+.*$"`.

Issue #104, opened 2021-05-05
Split out datasets included in the project


suvayu:
Including datasets in the main repository means it is there for perpetuity making it progressively more difficult to work with the repository or distribute it in the future.  So ideally we want to split it out and pull it back in as a work flow step.  Possible implementations:
1. host externally at a data hosting service like zenodo
    - *Pro:* standard tools to download the data and include in any work flow
    - *Con:* signals that we are responsible for maintaining the dataset, and a rather permanent, which is counter to the goal "Euro Calliope should not include any raw data"
2. collect them in a separate "data repository" in GH under the Calliope organisation, add it as a submodule, and in the setup instructions include the step `git clone --recurse-submodules`
    - *Pro:* relatively simple and easy to manage versions
    - *Con:* working with git-submodules can be a little difficult for beginners, and probably not as flexible to be included any generic "add my dataset" rule to future work flows
3. same as (2) but do not use git-submodules, write a general rule to clone a dataset into the data directory
    - *Pro:* simple, easy to extend for other data sources
    - *Con:* managing versions is potentially manual (and tedious) process.

sjpfenninger:
I'd be in favor of (3) for reasons of simplicity

suvayu:
same here; simplicity and more extensibility in the future (including the option to remove it)

timtroendle:
(3) seems to be the favourite option among people in the sprint.

PR #112, opened 2021-05-05
split data folder into separate repo

silas-schweizer:
Simple split, doesn't support multiple repos and subfolder selection


silas-schweizer:
@brynpickering and @sjpfenninger  can you please review this?

suvayu:
This fixes #104. Can someone link the issues? I do not have permission.

sjpfenninger:
@suvayu @silas-schweizer  Did you test this locally?

FLomb (changes requested):
Reviewed the pull request with @sjpfenninger. We fixed some minor issues with the rule definition (see commits), but still it looks like the rule isn't applied. We weren't able to test the rule, it doesn't seem to be recognised by snakemake

suvayu:
> @suvayu @silas-schweizer Did you test this locally?

No, clearly this wasn't trivial. I'm not sure I understand the dependency resolution used by snakemake. I'll scour the docs.

Snakefile
```diff
@@ -353,3 +353,12 @@ rule test:
     output: "build/logs/{resolution}/test-report.html"
     conda: "./envs/test.yaml"
     script: "./tests/test_runner.py"
+
+rule import_data_submodule:
```
brynpickering, line 357: Since you can curl any file you like from github, you could turn this into a wildcarded rule:

```suggestion
rule import_data_submodule:
    params: url = lambda wildcards: config["data-sources"]["data-repository"].format(dataset=wildcards.dataset)
    output: "data/{dataset}"
    shell: "curl -sLo {output} {params.url}"
```
Here, the config url would be `https://raw.githubusercontent.com/calliope-project/euro-calliope-datasets/main/{dataset}`.

Then if snakemake can't find the data (i.e. a user hasn't chosen to put in their own version of it), it will extract it from the repository on the fly, when looking for this dataset (e.g., [here](https://github.com/calliope-project/euro-calliope/blob/4de2d8af279990251ccfa7e7b05b18ce8c0c63cd/rules/hydro.smk#L115))

brynpickering (commented):
Possible alternative given, which downloads individual files on the fly (also reducing the number of files that need to be downloaded, if a user chooses to only run part of the workflow).

This might need updating so it doesn't clash with the search for data in `data/automatic`.

Snakefile
brynpickering: Committed this alternative as it seems to work well

config/default.yaml
```diff
@@ -11,6 +11,7 @@ data-sources:
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
     nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip
     potentials: https://zenodo.org/record/3533038/files/possibility-for-electricity-autarky.zip
+    data-repository: https://raw.githubusercontent.com/calliope-project/euro-calliope-datasets/main/{dataset}
```
sjpfenninger, line 14: Seems like an elegant solution that also allows versioning, i.e. the v.1.1 version of the euro-calliope repo can set data-repository to a v1.1 tagged version of the data

sjpfenninger (commented):
Ideally, the data repository would be a modified copy of the main repository with everything but `data` removed, so that the history of edits to `data` is preserved.

suvayu:
> Ideally, the data repository would be a modified copy of the main repository with everything but `data` removed, so that the history of edits to `data` is preserved.

IIUC, the datasets are final results from papers, or post-processed data releases, so what do you mean by history here?  I would think it's just addition of whole datasets.  I'm not sure if there's any useful information there.  If retaining the history is important, I can of course do some surgery and retain only the history of the `data` folder and push that to the data repository.

suvayu:
There's only one notebook in the repo (in `notebooks/`), maybe you mean the scripts in `scripts/`?  Also, keeping preprocessing steps in a notebook isn't a good idea.  If a notebook is necessary because of accompanying visualisations, I would either split that out, or use something like [`jupytext`](https://jupytext.readthedocs.io/en/latest/install.html) to convert to a script and keep them in sync.  But that can be done after the repo split.

brynpickering:
> There's only one notebook in the repo (in `notebooks/`), maybe you mean the scripts in `scripts/`? Also, keeping preprocessing steps in a notebook isn't a good idea. If a notebook is necessary because of accompanying visualisations, I would either split that out, or use something like [`jupytext`](https://jupytext.readthedocs.io/en/latest/install.html) to convert to a script and keep them in sync. But that can be done after the repo split.

No, it's the notebook (I hadn't notceid that there was only one). @timtroendle how do you want to proceed on data pre-processing scripts that are outside the scope of the workflow?

timtroendle:
I agree with @brynpickering that 
* the notebook should move as well,
* data should be deleted here.

In addition, data and notebook folders can be deleted (when you do, please also update the documented repo structure -- the documentation exists twice now: once in README.md and once in the new documentation folder).

When it comes to the question of notebook format or not, I have no strong feelings. Just keep in mind that this script must be execute only once per update of the data. Since the raw data is always in a slightly different format, I found a notebook easier to handle that than a script. But really, it does not matter much in my opinion.

brynpickering:
@suvayu @silas-schweizer I think this now only needs a changelog update and the notebook to be added to the `euro-calliope-datasets` repository....

suvayu:
@calliope-project/euro-calliope-devs can you have look at https://github.com/calliope-project/euro-calliope-datasets, and check if it meets expectations?

suvayu:
The commit I wanted to push is [attached](https://github.com/calliope-project/euro-calliope/files/6478475/0001-data-sources-update-data-repository-url.txt).  To apply: `git am path/to/patch`.

sjpfenninger:
Looks good to me. From a documentation and understandability perspective, would it make sense to have each dataset be in a top-level dir (either with a README inside, or along with a central list of all datasets = top-level dirs). That would mean having files with different spatial resolution inside the top-level dir, rather than top-level dirs for each spatial resolution with multiple datasets inside.

timtroendle:
https://github.com/calliope-project/euro-calliope-datasets looks good to me. Nice conservation of the Git history there! 👍

Some datasets are not needed anymore as they are automatically downloaded from Zenodo now. Those are the two datasets "farmland.csv" and "land-cover.csv" for all three resolutions. That means all three top-level folders `continental`, `national`, `regional` can go entirely. This resolves your comment, @sjpfenninger.

brynpickering:
@timtroendle Is there any reason this can't be merged in?

timtroendle:
Hm, I am not a reviewer here. But none of the three reviewers has accepted the change. 

I have on comment, though: As I mentioned above, some datasets aren't needed anymore and I suggest to remove them before merging this.

brynpickering (approved):

README.md
```diff
@@ -106,17 +106,19 @@ You can easily change the units and scale all values using the `scaling-factor`
 
 * `build/model`: Contains the entire model after the build step, including Calliope definition files and data (does not exist initially).
 * `config`: Files with configuration parameters which influence the model build process.
-* `data`: Small input data used within the model build process.
-* `data/automatic`: Contains data automatically downloaded within the model build process (does not exist initially).
+* `data`: Contains data automatically downloaded within the model build process (does not exist initially).
```
timtroendle, line 109: This change is incorrect. `data/automatic` still exists and will be used for all data automatically downloaded from other sources than euro-calliope-datasets.

I don't think it makes sense to load euro-calliope-datasets into `data` and load all other datasets into `data/automatic`. We should either load everything into `data` or everything into `data/automatic` because to the user, it really makes no difference where the data are downloaded.

I prefer the latter. One may use manually downloaded data (I do that often) and one would want to differentiate it from automatically downloaded data. Therefore, I'd go for downloading everything into `data/automatic`.

timtroendle (changes requested):
The use of data folders is inconsistent now and I think we should fix it before merging this change.

Snakefile
```diff
@@ -363,3 +363,12 @@ rule test:
     output: "build/logs/{resolution}/test-report.html"
     conda: "./envs/test.yaml"
     script: "./tests/test_runner.py"
+
+
+rule import_data_submodule:
```
timtroendle, line 368: Can we rename this rule to `download_xxx`, e.g. `download_eurocalliope_dataset`?

It's really useful to be able to tell by the name that this is a rule downloading data.

README.md
brynpickering: Ideally it would be differentiated from the rest of the data in `data/automatic` so that we can have a wildcard constraint that ignores this rule unless downloading from the eurocalliope-dataset repo (as is currently the case). It could go into something like a `data/eurocalliope-dataset` directory?

brynpickering: This would also enable a user to delete all the data from the dataset in one go (e.g. if they update the GH repo pointed to in the config) without having to work out which subdirs in `automatic` or elsewhere are coming from there.

brynpickering (changes requested):
Need to sort out mistake with EEZ filepath change and update curl to return error on 404, not to save a file with the error in it.

brynpickering (approved):

brynpickering:
Minimal workflow completes successfully with latest commit.

timtroendle (approved):

timtroendle:
Thanks @silas-schweizer @suvayu @sjpfenninger @FLomb @brynpickering for your work on this!

PR #113, opened 2021-05-05
Move across annual energy balance from 2.0 repository

### LINKED ISSUE #100: Get annual energy balances from Eurostat
Energy consumption by sector (industry subsectors, commercial, household, agriculture, and others) as well as by fuel type (solid and liquid fossils, gas, biofuels, renewables, waste, district heat, direct electricity, etc.) are available on Eurostat (nrg_bal_c), as well as the allocation of that consumption to end uses for households (space heat, water heat, cooking - nrg_d_hhq). This is the starting point for calculating annual energy demands for each country for the years 2000-2018.

On top of this, to include Switzerland in everything, specific [Swiss datasets](https://www.bfe.admin.ch/bfe/de/home/versorgung/statistik-und-geodaten/energiestatistiken.html) need to be processed alongside the Eurostat datasets.

Together, these are currently processed in our sector-coupled sibling [here](https://github.com/calliope-project/euro-calliope-2.0/blob/main/src/construct/annual_energy_balance.py), including grouping carriers to a subset of interest and combining Eurostat and Swiss energy balances and household end uses.

sjpfenninger:
Fixes #100

### DIFF
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -9,6 +9,45 @@ properties:
     data-sources:
         type: object
         properties:
+            eurostat:
+                type: object
+                properties:
+                    base-url:
+                        type: string
+                        description: Base UR for Eurostat data
+                    energy-balance:
+                        type: string
+                        description: Eurostat table name for energy balances
+                    hh-end-use:
+                        type: string
+                        description: Eurostat table name for end uses
+                    freight:
+                        type: string
+                        description: Eurostat table name for freight
+                    employees:
+                        type: string
+                        description: Eurostat table name for employees
+                    gva:
+                        type: string
+                        description: Eurostat table name for gva
+                    dwellings:
+                        type: string
+                        description: Eurostat table name for dwellings
+            swiss-stat:
+                type: object
+                properties:
+                    end-use:
+                        type: string
+                        description: Swiss-stat table name for end uses
+                    energy-balance:
+                        type: string
+                        description: Swiss-stat table name for energy balances
+                    industry-energy-balance:
+                        type: string
+                        description: Swiss-stat table name for industrial energy balances
+                    gva:
+                        type: string
+                        description: Swiss-stat table name for gva
             biofuel-potentials:
                 type: string
                 pattern: .*{feedstock}.*\.csv$
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -9,30 +9,9 @@ properties:
     data-sources:
         type: object
         properties:
-            eurostat:
-                type: object
-                properties:
-                    base-url:
-                        type: string
-                        description: Base UR for Eurostat data
-                    energy-balance:
-                        type: string
-                        description: Eurostat table name for energy balances
-                    hh-end-use:
-                        type: string
-                        description: Eurostat table name for end uses
-                    freight:
-                        type: string
-                        description: Eurostat table name for freight
-                    employees:
-                        type: string
-                        description: Eurostat table name for employees
-                    gva:
-                        type: string
-                        description: Eurostat table name for gva
-                    dwellings:
-                        type: string
-                        description: Eurostat table name for dwellings
+            eurostat-base-url:
+                type: string
+                description: Base URL for Eurostat data
             swiss-stat:
                 type: object
                 properties:
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -11,22 +11,28 @@ properties:
         properties:
             eurostat-base-url:
                 type: string
+                pattern: ^(https?|http?):\/\/.+{dataset}.tsv.gz$
                 description: Base URL for Eurostat data
             swiss-stat:
+                description: Set of data sources for Swiss-specific data 
                 type: object
                 properties:
                     end-use:
                         type: string
-                        description: Swiss-stat table name for end uses
+                        pattern: ^(https?|http?):\/\/.+
+                        description: Source of Swiss annual end use energy consumption
                     energy-balance:
                         type: string
-                        description: Swiss-stat table name for energy balances
+                        pattern: ^(https?|http?):\/\/.+
+                        description: Source of Swiss annual energy balances, broadly equivalent to Eurostat energy balances
                     industry-energy-balance:
                         type: string
-                        description: Swiss-stat table name for industrial energy balances
+                        pattern: ^(https?|http?):\/\/.+
+                        description: Source of Swiss annual industry sub-sector energy balances
                     gva:
                         type: string
-                        description: Swiss-stat table name for gva
+                        pattern: ^(https?|http?):\/\/.+
+                        description: Source of Swiss Gross Added Value per canton (NUTS3 administrative level)
             biofuel-potentials:
                 type: string
                 pattern: .*{feedstock}.*\.csv$
--- a/lib/eurocalliopelib/utils.py
+++ b/lib/eurocalliopelib/utils.py
@@ -1,18 +1,55 @@
 """Utility functions."""
 import pycountry
+import pandas as pd
 
 
 def eu_country_code_to_iso3(eu_country_code):
     """Converts EU country code to ISO 3166 alpha 3.
     The European Union uses its own country codes, which often but not always match ISO 3166.
     """
-    assert len(eu_country_code) == 2, "EU country codes are of length 2, yours is '{}'.".format(eu_country_code)
+    assert (
+        len(eu_country_code) == 2
+    ), "EU country codes are of length 2, yours is '{}'.".format(eu_country_code)
     if eu_country_code.lower() == "el":
         iso2 = "gr"
     elif eu_country_code.lower() == "uk":
         iso2 = "gb"
-    elif eu_country_code.lower() == "bh": # this is a weird country code used in the biofuels dataset
+    elif (
+        eu_country_code.lower() == "bh"
+    ):  # this is a weird country code used in the biofuels dataset
         iso2 = "ba"
     else:
         iso2 = eu_country_code
     return pycountry.countries.lookup(iso2).alpha_3
+
+
+def get_alpha2(country, eurostat=True):
+    """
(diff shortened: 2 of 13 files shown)


sjpfenninger:
Needs contributor info added in when merging

sjpfenninger:
@brynpickering Also this is missing the two files that need to be moved across to the `data` directory -- pending that being possibly removed from the repository

lib/eurocalliopelib/utils.py
```diff
@@ -1,18 +1,55 @@
...
+    elif (
+        eu_country_code.lower() == "bh"
+    ):  # this is a weird country code used in the biofuels dataset
         iso2 = "ba"
     else:
         iso2 = eu_country_code
     return pycountry.countries.lookup(iso2).alpha_3
+
+
+def get_alpha2(country, eurostat=True):
```
brynpickering, line 26: This should possibly have the same structure as `eu_country_code_to_iso3`, or merge them with the option to spit out ISO2 or ISO3, or make it a class which subclasses pycountry?

config/schema.yaml
```diff
@@ -9,6 +9,45 @@ properties:
     data-sources:
         type: object
         properties:
+            eurostat:
+                type: object
+                properties:
+                    base-url:
+                        type: string
+                        description: Base UR for Eurostat data
```
FebinKa, line 17: *URL

rules/data-processing.smk
```diff
@@ -0,0 +1,59 @@
+"""Rules to process input data."""
+
+configfile: "./config/default.yaml"
+localrules: eurostat_data_tsv, ch_data_xlsx
+root_dir = config["root-directory"] + "/" if config["root-directory"] not in ["", "."] else ""
+script_dir = f"{root_dir}scripts/"
+
+
+rule eurostat_data_tsv:
```
brynpickering, line 9: Agreed with your PR comment that this would be better generalised. A possible solution would be: 
```suggestion
rule eurostat_data_tsv:
message: "Get {wildcard.dataset} from Eurostat"
params: 
    base_url = config["data-sources"]["eurostat"]["base-url"],
    file_url = lambda wildcards: config["data-sources"]["eurostat"]["{wildcards.dataset}"]
output: protected("data/automatic/{dataset}.tsv.gz")
shell: "curl -sLo {output} {params.base_url}{params.file_url}.tsv.gz"
```

Then when using this data as input you need to refer to the filename rather than the rule (`rules.eurostat_data_tsv.output.energy_balance` -> `"data/automatic/energy-balance.tsv.gz"`)

rules/data-processing.smk
```diff
@@ -0,0 +1,59 @@
...
+        curl -sLo {output.energy_balance} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][energy-balance]}.tsv.gz'
+        curl -sLo {output.hh_end_use} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][hh-end-use]}.tsv.gz'
+        curl -sLo {output.freight} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][freight]}.tsv.gz'
+        curl -sLo {output.employees} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][employees]}.tsv.gz'
+        curl -sLo {output.gva} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][gva]}.tsv.gz'
+        curl -sLo {output.dwellings} '{config[data-sources][eurostat][base-url]}{config[data-sources][eurostat][dwellings]}.tsv.gz'
+        """
+
+
+rule ch_data_xlsx:
```
brynpickering, line 29: See suggestion above, could be reproduced here, just with `protected("data/automatic/ch_{dataset}.tsv.gz")`

rules/data-processing.smk
brynpickering: To differentiate from the ch one, output should probably be `protected("data/automatic/eurostat_{dataset}.tsv.gz")`

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+    statistics.
+    """
+    # Names for each consumption category/sub-category and carriers have been prepared by hand
+    cat_names = pd.read_csv(path_to_cat_names, header=0, index_col=0)
+    carrier_names = pd.read_csv(path_to_carrier_names, header=0, index_col=0)
+    country_codes = [utils.get_alpha2(i, eurostat=True) for i in countries]
+
+    df = pd.read_csv(path_to_input, delimiter="\t", index_col=0)
+    df.index = df.index.str.split(",", expand=True).rename(
+        ["cat_code", "carrier_code", "unit", "country"]
```
brynpickering, line 34: maybe this list could be in config?

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+                ch_transport_energy_use,
+            ]
+        ]
+    )
+
+    return ch_energy_use_tdf
+
+
+def get_ch_energy_balance_sheet(path_to_excel, sheet, skipfooter, cat_code):
+    ch_energy_carriers = {
```
brynpickering, line 92: Maybe have these at the top of the script as a constant?

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+        .assign(cat_code=cat_code)
+        .set_index("cat_code", append=True)
+        .stack()
+    )
+
+    return _df
+
+
+def get_ch_waste_consumption(path_to_excel):
+    """
```
FebinKa, line 139: Why does only this function have a description? Would probably be helpful to have one for the others as well

scripts/annual_energy_balance.py
brynpickering: In fact, maybe this could be in a new lib util file related to opening files?

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+        .dropna(subset=["TOTAL"])
+    )
+    df.columns = (
+        df.columns.str.extract("(\d+)", expand=False)
+        .fillna(0)
+        .astype(int)
+        .map(ch_subsectors)
+    )
+    # combine any data that now has the same cat_code or carrier_code by using groupby
+    df = df.groupby(axis=1, level=0).sum().groupby(level=[0, 1]).sum()
```
brynpickering, line 263: use explicit level names rather than numbers? Would make the script more explicit.

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+    }
+
+    ch_carriers = {
+        "ELEKTRIZITÄT": "E7000",  # 'electricity',
+        "HEIZÖL EXTRA-LEICHT": "O4000XBIO",  # 'oil',
+        "ERDGAS": "G3000",  # 'gas',
+        "KOHLE": "C0000X0350-0370",  # 'solid_fuel',
+        "INDUSTRIEABFÄLLE": "W6100_6220",  # 'waste',
+        "HEIZÖL MITTEL UND SCHWER": "O4000XBIO",  # 'oil',
+        #    'FERNWÄRME KUMULIERT': 'heat',  # total
```
brynpickering, line 237: Maybe the source of 'heat' could be configurable - choosing between "FERNWÄRME KUMULIERT", "FERNWÄRME BEZUG", "FERNWÄRME ABGABE"

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+    carrier_name_func = (
+        lambda x: _df.columns.to_frame()
+        .iloc[:, x]
+        .str.translate(remove_digits)
+        .map(carriers)
+    )
+    _df.columns = carrier_name_func(0).fillna(carrier_name_func(1)).values
+
+    _df = (
+        _df.groupby(axis=1, level=0)
```
brynpickering, line 200: explicit level name

FebinKa (commented):
Only minor comments, took me a while to get oriented but as far as I understood the code, it seems to make sense.

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+    _df = pd.read_excel(
+        path_to_excel,
+        skiprows=6,
+        skipfooter=12,
+        index_col=0,
+        sheet_name="T17e",
+        header=[0, 1, 2, 3, 4],
+    ).xs("TJ", level=-1, axis=1)
+    # Footnote labels lead to some strings randomly ending in numbers; we remove them here
+    remove_digits = str.maketrans("", "", digits)
```
brynpickering, line 189: could be a util function, then we can test it!

scripts/annual_energy_balance.py
```diff
@@ -0,0 +1,277 @@
...
+        .str.translate(remove_digits)
+        .map(ch_energy_carriers)
+        .rename("carrier_code")
+    )
+    _df.index.rename("year", inplace=True)
+
+    _df = (
+        _df.apply(utils.to_numeric)
+        .astype(float)
+        .assign(cat_code=cat_code)
```
brynpickering, line 130: I frequently added multiindex levels using this 'assign->set_index' functionality. This could be moved to a lib util function.

brynpickering (changes requested):
The first 2.0 update - woohoo!

A bunch of changes to make to try and simplify things (my initial implementation was perhaps too verbose ;) )

Functions added to eurocalliopelib should be tested, and the final `annual-energy-balance.csv` file should be tested, to ensure it has at least the expected countries and multi-index levels.

Final point: this feature should be added to the `feature-sector-coupling`, so the PR needs updating (or re-submitting?) to point to a merge on that branch.

scripts/annual_energy_balance.py
sjpfenninger: @brynpickering could you supply minimal docstrings for all these functions?

brynpickering (approved):

Issue #106, opened 2021-05-05
Add sphinx documentation


sjpfenninger:
Add minimal sphinx documentation:

* Set up Sphinx with markdown: https://www.sphinx-doc.org/en/master/usage/markdown.html
* Start documentation based on the README
* Set it up to auto-build on readthedocs

Issue #103, opened 2021-05-05
More documentation


sjpfenninger:
Initial tasks:

* [x] Create minimal sphinx documentation #106
* [x] Add background on the purpose #107 (also see: Clarify purpose and naming #105)
* [x] Add overview of repository layout #108
* [x] Add developer docs #109

Later possible extensions:

* Overview of models that the workflow can build (including links to pre-builts)
* Various auto-built documentation on the components, data sources, models, ...

sjpfenninger:
We can use sphinx with markdown: https://www.sphinx-doc.org/en/master/usage/markdown.html

sjpfenninger:
This is a meta-issue that tracks all the initial setup for the docs, so should remain open?

timtroendle:
The four issues mentioned above have been covered and in addition several more (#159, #156, #68, #165).

PR #114, opened 2021-05-05
Add basic docs

FebinKa:
Adding README.md as documentation using Sphinx to address issue #106


docs/source/conf.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = "euro-calliope-workflow"
+copyright = "2021, calliope-project authors"
+author = "calliope-project authors"
+
+# The full version, including alpha/beta/rc tags
+release = "1.1"
```
sjpfenninger, line 26: Should the version be pulled from somewhere else, e.g. at the top level of the repo?

sjpfenninger (changes requested):
Looks good. I have only one comment regarding setting the version number, and a question:

Should the dependencies be in the environment.yml file (I added them in there for now) or a separate docs file? If we later build DAG graphics for the docs I guess we need all requirements in one place to build the docs e.g. on RTD.

I also made two minor changes:

* increased header level in markdown file as prep for splitting across top-level headings
* changed to RTD sphinx theme (cleaner than alabaster I think but happy to revert if you disagree)

sjpfenninger (approved):

timtroendle:
@FebinKa @sjpfenninger For my understanding: you copied the content from README.md to the new doc. As with any copy, it's only a matter of time until the two will get out of sync. How do you think we should handle that?

FebinKa:
It's not exactly the same as @sjpfenninger changed the header levels and the file was formatted to include the sources before the license. The goal would be to address #103 through the Sphinx documentation and using the copy allows to paste the new inputs at the appropriate places. However, it would also be possible to [directly link](https://stackoverflow.com/a/17217041/15452089) to the readme instead by creating `readme_link.rst` inside `docs/source` which contains ` include:: ../../README.md` and calling that instead of `overview` in `index.md`.

timtroendle:
Thanks for the clarification. I am really just worried that we are now maintaining two things, and we should have a strategy moving forward how to handle that.

brynpickering:
@FebinKa I like the suggestion to point to the README directly to fill in the 'overview' information, instead of the copy-paste job we did. Perhaps you could implement that? In fact, the index could just show that, then we don't have to worry about having moved some but not all things from the readme to the index. The sub-pages can be reserved for more details on the workflow.

PR #117, opened 2021-05-05
Add pull request template

brynpickering:
These were two checklist items I thought necessary to have in every PR, to ensure that they're not overlooked. We could also have the PR fail without them using: https://github.com/marketplace/task-list-completed

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,8 +5,4 @@ Fixes # .
 - [ ] CHANGELOG updated
 - [ ] Minimal workflow tests pass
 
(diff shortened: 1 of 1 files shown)


timtroendle (approved):
Looks good. What do you do if your change does not require a CHANGELOG update and/or a test run? (like this PR for example)

brynpickering:
My preference would be to always have the checklist items there, and they are just checked by the PR creator/reviewer if they are seen to be redundant. This makes more sense than deleting them, in case it turns out that the creator accidentally deleted something that was necessary.

PR #118, opened 2021-05-06
Getting README from root directory

FebinKa:
Re: the discussion in #114, the README is retrieved from the root directory instead of copying it into the project folder.


sjpfenninger (approved):
Seems like the right solution to me!

brynpickering:
I've updated this to have the README be the index page. I guess in the long run we wouldn't have much overlap between the README and the docs, but this acts as a placeholder.

Issue #17, opened 2020-04-14
Add annual transport demand (passenger, freight, shipping, aviation)


brynpickering:
As with #101, transport demand is needed to research sectoral integration. The plan would be to model passenger demand assuming electric vehicles with time-dependent plug-in schedules (#81) and to model everything else using annual demand data provided in [Eurostat energy balances](https://ec.europa.eu/eurostat/web/energy/data/energy-balances), processed in #100.

brynpickering:
This is implemented in our sector-coupled sibling based on a mix of eurostat and JRC IDEES data [here](https://github.com/calliope-project/euro-calliope-2.0/blob/main/src/construct/annual_transport_demand.py)

timtroendle:
Blocked by #100 and #101.

timtroendle:
Definition of done: tidy dataframe (csv) or netcdf with indices country, year, subsector (commercial, non-commercial), mode.

The quantity of the variable of that dataset is fuel consumption for all modes other than road transport for which it is distance travelled.

timtroendle:
Closing this as this was (partly) implemented on the now-abandoned feature branch.

PR #119, opened 2021-05-10
Import annual transport demand, as in 2.0

FLomb:
Addresses issue #17


lib/eurocalliopelib/utils.py
```diff
@@ -1,18 +1,69 @@
...
+def pj_to_twh(array):
+    """Convert PJ to TWh"""
+    return array / 3.6
+
+def tj_to_twh(array):
+    """Convert TJ to TWh"""
+    return pj_to_twh(array) * 1e-3
+
+def read_tdf(filename):
+    """ TBA """
```
brynpickering, line 66:
```suggestion
    """ 
    Read a tidy dataframe from CSV. 
    This assumes that all except the final column is an index level,  so returns a MultiIndexed Pandas Series. 
    """
```

rules/data-processing.smk
```diff
@@ -0,0 +1,81 @@
...
+        cat_names = "data/energy_balance_category_names.csv",
+        carrier_names = "data/energy_balance_carrier_names.csv"
+    output: "build/annual_energy_balances.csv"
+    params:
+        countries = config["scope"]["countries"]
+    conda: "../envs/default.yaml"
+    shadow: "minimal"
+    script: "../scripts/annual_energy_balance.py"
+
+rule annual_transport_demand:
```
brynpickering, line 61: It would actually be good to split this into multiple rules which output information on road, rail, air, and shipping separately. Then it would be possible to avoid generating one or more of them as necessary.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
...
+        road_distance_df, road_vehicles_df, total_road_distance
+    )
+    # Some cleanup that's specific to road data
+    road_efficiency = (
+        # 25th percentile of 2015 efficiency in TWh/mio km
+        (1 / road_efficiency.xs(2015, level='year'))
+        .unstack(['vehicle_type', 'vehicle_subtype'])
+        .quantile(EFFICIENCY_QUANTILE)
+        .unstack(0)
+        .groupby({
```
brynpickering, line 70: Move this dictionary to be another constant (like CARRIERS) at the top of the doc.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
...
+    road_distance_df = utils.read_tdf(jrc_road_distance_path)
+    road_vehicles_df = utils.read_tdf(jrc_road_vehicles_path)
+    rail_energy_df = utils.read_tdf(jrc_rail_energy_path)
+    rail_distance_df = utils.read_tdf(jrc_rail_distance_path)
+
+    # Add transport energy demand from agriculture and 'not elsewhere specified' (military)
+    # assumption: agriculture oil use goes to 'road' transport demand;
+    # fisheries' oil use goes to marine fuel demand;
+    # 'not elsewhere specified' oil use goes predominantly to 'road' transport, except kerosene which goes to aviation
+    other_transport_aviation = (  # all kerosene from the military destined for aviation
```
brynpickering, line 36: Processing "other" energy demands should probably happen in a different rule and file, then this script can have it as an input.

brynpickering: Or, at least have it contained in its own function in this script.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
+import pandas as pd
+
+from eurocalliopelib import utils
+
+EFFICIENCY_QUANTILE = 0.25
```
brynpickering, line 5: Move to config so that it can be user-defined. This option basically sets the future road vehicle efficiency based on the distribution of efficiencies across countries in Europe according to JRC IDEES.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
...
+        .xs('electricity')
+    )
+    # High speed and metro are all electrified, so bau_electricity == eurocalliope_electricity
+    assert abs(
+        (rail_energy - rail_bau_consumption.xs('electricity', level='carrier'))
+        [['High speed passenger trains', 'Metro and tram, urban light rail']]
+    ).max() < 1e-10
+
+    # Air and shipping energy
+    air_energy = (
```
brynpickering, line 114: Separate into a different function.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
...
+        .unstack('carrier_code')
+        ['O4000XBIO']
+        .sum(level=['country', 'year'])
+        .add(other_transport_aviation, fill_value=0)
+        .apply(utils.tj_to_twh)
+        .to_frame('twh')
+        .rename_axis(columns='unit')
+        .stack()
+    )
+    marine_energy = (
```
brynpickering, line 126: As above, move to own function.

scripts/annual_transport_demand.py
```diff
@@ -0,0 +1,273 @@
...
+        }).mean()
+        .assign(unit='twh_per_mio_km').set_index('unit', append=True)
+        .stack()
+    )
+    road_electricity_bau = (
+        road_bau_consumption
+        .sum(level=['carrier', 'vehicle_type', 'country_code', 'year', 'unit'])
+        .xs('electricity')
+    )
+    total_rail_distance, rail_efficiency, rail_bau_consumption = get_all_distance_efficiency(
```
brynpickering, line 86: Move to its own rail-specific rule and file

brynpickering (commented):
Broadly looks fine, but perhaps we can improve the annual transport demand file to improve understanding of what's going on. This would involve smaller functions, clearer naming, and moving some of the functionality into an entirely separate file associated with a different rule (e.g. rail demand).

scripts/annual_transport_demand.py
FLomb: Now that every mode has its own script maybe having the 'other' energy demand, for each mode, in the respective script makese more sense?

FLomb:
Updated the PR in such a way to have separate snakemake rules and scripts for each transport mode, for users to be free to decided which of those to integrate in their workflow and which not. 

Tested locally and it works. Some notes, though:
- for road and rail rules I get this message (despite not seeming to affect the output): 
```
/home/fl/GitHub-repos/euro-calliope/.snakemake/conda/fc00feb3/lib/python3.8/site-packages/pandas/core/indexes/multi.py:3587: RuntimeWarning: The values in the array are unorderable. Pass `sort=False` to suppress this warning.
  uniq_tuples = lib.fast_unique_multiple([self._values, other._values], sort=sort)
```
- there are 2 functions which are re-used identically by both road and rail scripts. I tried to put those just in road and let rail import them from the latter, but that doesn't work. Any suggestions to avoid such a repetition?
- I noticed that "Motor coaches, buses and trolley buses" are the only vehicle type which, in the output .csv, is shown as a string between quotation marks, unlike all others which are displayed without quotation marks. Not sure if this has any relevance, nor how to address it (I guess it's a bug of the JRC IDEES pre-processing stage?).

brynpickering (commented):
Good to see a split, but we could do with not having the same code duplicated across multiple files, for functions that get used by more than one rule. Two options would be:

1. Keep all the functionality in one script and have a rule parameter which gives the transport mode (e.g. "road"). Then, in the script `__name__ == "__main__"` part, run the function that is relevant to the given parameter (so you would then run  `get_road_transport_demand` because `snakemake.params.mode == "road"`).
2. Move overlapping functions to the eurocalliopelib.

I'm in favour of (1) as these functions are of limited use outside processing road transport stuff. However, (2) is more in line with what we currently do; having these snakemake-induced script switches isn't something we do and I don't know if maybe it's a bad idea in some way...

brynpickering:
I have a few local changes to make to update this, but the PR is blocked by #133 as testing the PR requires the processed jrc-idees data to be rolled into the workflow.

brynpickering (approved):
I'm no longer an unbiased reviewer, so will leave it to @timtroendle to make the final decision.

timtroendle:
Closing this as it became stale. Also, it appears to me that some additions from the PR are already on `feature-sector-coupling`.

PR #120, opened 2021-05-12
Add infrastructure to test scripts

timtroendle:
This allows to run tests not only of the library but also scripts by running `py.test` in the root directory. See #91 for a use case for this.

The downside of this approach is that py.test can not handle conda environments. All tests will run in the same environment.


Snakefile
```diff
@@ -348,10 +350,9 @@ rule docs:
 rule test:
     message: "Run tests"
     input:
-        test_dir + "test_runner.py",
-        test_dir + "test_model.py",
-        test_dir + "test_capacityfactors.py",
         "build/logs/{resolution}/model.done",
+        test_dir = worfklow_test_dir,
+        tests = lambda wildcards: Path(worfklow_test_dir).glob("**/*.py"),
```
brynpickering, line 355:
```suggestion
        tests = lambda wildcards: Path(worfklow_test_dir).glob("**/test_*.py"),
```

docs/source/overview.md
```diff
@@ -124,3 +124,14 @@ As a developer, you may want to run the entire workflow often to spot errors ear
     snakemake --use-conda --configfile="config/minimal.yaml"
 
 Make sure to run this in a clean working directory. Do not use the working directory in which you are using your normal configuration.
+
```
brynpickering, line 127: This file is going in #118, in favour of directly printing the README as the docs overview. Either put this in the README or as a new file concerning testing in the docs.

.github/workflows/pythonpackage.yaml
```diff
@@ -16,8 +16,8 @@ jobs:
...
                 auto-update-conda: true
                 python-version: 3.8
-                activate-environment: test-eurocalliopelib
-                environment-file: lib/test-requirements.yaml
+                activate-environment: test-eurocalliope
+                environment-file: test-requirements.yaml
             - name: Test with pytest
               run: |
-                pytest ./lib
+                pytest
```
brynpickering, line 23: So this will run all tests that it can (in lib and workflow), right? In most cases this will be quite limited with the workflow, as it will ignore all tests associated with workflow outputs, but it will still pick up the function unit tests?

Snakefile
brynpickering: or maybe not, if don't actually want a list of the test paths, but rather just to have snakemake track all file timestamps.

brynpickering (commented):
It certainly increases the coupling of the eurocalliopelib with the euro-calliope workflow, which is probably fine.

I would call the workflow test directory `scripts` rather than `workflow`, to match the respective folder names.

.github/workflows/pythonpackage.yaml
timtroendle: Mhm so the idea was that there are three types of test:

* tests/scripts -> unit tests of functions in script
* tests/lib -> unit tests of functions in the library
* tests/workflow -> tests run within the workflow that rely on workflow outputs

The former two are performed by simply running `py.test`. The latter is run whenever the entire workflow is run.

Does that make sense? Which names make most sense for these folders?

docs/source/overview.md
timtroendle: I added it to README, too.

Snakefile
timtroendle: Which _should_ be the same, really. Still, I think that suggestion is making the purpose more clear.

.github/workflows/pythonpackage.yaml
brynpickering: I see, that is clear. I missed the fact that there is a `tests/scripts` and `tests/workflow`. Maybe it could be `tests/build` instead of `workflow`? And then this directory of tests could be actively omitted in the CI tests?

timtroendle: Sounds good. The directory is already omitted in the CI tests (and when you run `py.test` yourself): see the base `pytest.ini`

Snakefile
```diff
@@ -35,6 +36,7 @@ __version__ = open(f"{root_dir}VERSION").readlines()[0].strip()
 script_dir = f"{root_dir}scripts/"
 template_dir = f"{root_dir}templates/"
 test_dir = f"{root_dir}tests/"
+worfklow_test_dir = f"{test_dir}workflow"
```
suvayu, line 39: typo, but maybe doesn't matter, as the typo is repeated later

pytest.ini
```diff
@@ -1,5 +1,3 @@
 [pytest]
-filterwarnings=
-    ignore::calliope.exceptions.ModelWarning:calliope
-    ignore:.*inspect.getargspec\(\) is deprecated.*:DeprecationWarning:pyomo
-norecursedirs = .git lib .snakemake
+norecursedirs = .git .snakemake tests/workflow
```
brynpickering, line 2: Add comment here to mention why `tests/workflow` is in this list.

suvayu (approved):
Discussed with @brynpickering , since the two key points are covered:
- including `pytest-pythonpath`
- adding `scripts` to `python_paths` in `pytest.ini` (along with dummy example)

I don't have anything to add.

timtroendle:
I named tests of workflow results `tests/model` instead of `tests/build`. The reason is that folders called `build` are likely to be ignored by default by applications like Git (on my machine it was) and we should avoid running into such a situation. `tests/model` also nicely describes what these tests do.

brynpickering:
I think it is still most descriptive to call it build, since one could imagine testing intermediate build outputs which are not the model itself. As to the gitignore issue, this is easily solved by updating the gitignore to a. not have reference two "build" twice and b. set the one ignore to `/build` (see [this](https://stackoverflow.com/questions/44155648/ignore-a-particular-folder-only-in-gitignore-but-allow-other-folders-with-same) nice summary).

timtroendle:
Of course we can change the `.gitignore` in this repo, but that's not what I was talking about. `build` is often ignored in default settings and thus may be set globally on a machine (on mine, it is!). [Github's seemingly official .gitignore for Python](https://github.com/github/gitignore/blob/master/Python.gitignore) ignores `build`, as an example.

I agree that `build` is more accurate than `model`, but the risk of this causing unseen issues does not outweigh this advantage in my opinion.

brynpickering:
OK, fair enough. Then I would consider renaming the `build` directory in the repository. This issue doesn't block this PR, but is worth dealing with in the near-term.

timtroendle:
@brynpickering, any more comments here? Another wise I will merge.

brynpickering:
merge away!

timtroendle:
Thanks. Oh and I meant otherwise not another wise. 🤦‍♂️😄

PR #125, opened 2021-05-14
Fix descriptions of configuration values

### LINKED ISSUE #124: Some configuration value descriptions are broken
In #111, all configuration values were added to the schema. In the same time, comments from the config file were removed and added as descriptions to the schema. For some comments, this makes sense, for others it does not. The latter ones must be fixed. Some of the comments should be reintroduced to the config in my opinion.

The reason why some descriptions are broken now is that some comments were not describing the configuration parameter, but the configuration parameter _value_.  For example, the comment `# from MW(h) to 100 GW(h)` told the reader that the configuration parameter value of `0.00001` was scaling from MW to 100 GW. When the value is changed, this comment is of course not valid anymore. Therefore, this description must be removed from the schema. In addition, I suggest to reintroduce the comment to the config.

timtroendle:
Fixes #124.

## Checklist

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -1,3 +1,4 @@
+# For descriptions of the configuration values, see `./schema.yaml`.
 data-sources:
     biofuel-potentials: data/biofuels/potentials/{feedstock}.csv
     biofuel-costs: data/biofuels/costs/{feedstock}.csv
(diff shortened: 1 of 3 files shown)


brynpickering (approved):
👍 Could you also add a description to 'shapes' in schema?

timtroendle:
Ok, done. I kept the original `shapes` comment within the default configuration. It's too important to be overlooked.

Issue #102, opened 2021-05-04
Add annual industry demand to account for new processes and avoid reliance on solid fossils


brynpickering:
looking beyond current industry demand and considering the possibility of a decarbonised industry in future leads to the need to disaggregate industry subsection processes and then re-build them based on future processes.

This means working out what current energy demand is per industry sub-sector, allocating it to processes, and then recalculating demand based on updates to those processes (electrification, hydrogen as a reduction agent in steel-making, methanol/naphtha as input streams to plastics production). This has been implemented in our sector-coupled sibling [here](https://github.com/calliope-project/euro-calliope-2.0/blob/main/src/construct/annual_industry_demand.py)

Tasks involved:

1. Get demand for non-steel/chemical industry from JRC IDEES
  a. Get electricity demand
  b. Get methane demand for high temperature process (assumed decarbonisable)
  c. Get diesel demand for backup generators (assumed decarbonisable)
2. Update steel industry demand
 a. Get demand for steel industry from JRC-IDEES
 b. Replace energy consumption in reducing iron and blast furnaces with pre-computed hydrogen and electricity requirements for direct reduction of iron (DRI)
  c. update energy consumption for processing steel based on electric arc processes applied to all steel, not only recycled steel
  d. Add all demands together
3. Update chemical industry demand
  a. Get demand from JRC-IDEES
  b. calculate quantity of ethylene, propylene, BTX, ammonia, and methanol (i.e. base chemicals) needed by country. 
  c. calculate hydrogen, CO2 and electricity demands to produce these base chemicals, based on molar masses
4. combine all industry demands together, and fill data from countries not in JRC-IDEES with averages
5. Verify that this processing hasn't vastly increased total energy demand from industry in any country.

This issue is linked to #63

PR #133, opened 2021-07-16
Add JRC-IDEES data processing pipeline

brynpickering:
First step to fixing #17 & #102 and blocking #119.

Now that the JRC IDEES database is available via FTP, we can roll the processing of the data into Euro-Calliope. This allows the transport, industry, and tertiary (commercial) sector data to be processed for the components that we need down the line. The processing takes the largely machine-unreadable data from JRC-IDEES and turns into tidy dataframes for use elsewhere.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,127 @@
...
+        df = pd.read_excel(file, sheet_name=sheet_name, index_col=0)
+        start = df.filter(regex='Physical output', axis=0)
+        end = df.filter(regex='Installed capacity', axis=0)
+
+        df_prod = (
+            df
+            .loc[start.index[0]:end.index[0]]
+            .iloc[1:-1]
+            .dropna(how='all')
+            .assign(country_code=df.index.name.split(':')[0], cat_name=df.index.name.split(': ')[1], unit='kt')
```
timtroendle, line 112: as above. consider making this a four-liner

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,127 @@
...
+            ]
+        df.loc[:idx_end, 'indent'] = [int(i) for i in style_df[col].style.indent.loc[:idx_end]]
+
+        # To check at the end that we haven't lost some data
+        tot = df[df.indent == 1].sum()
+
+        df = df.dropna(subset=['section', 'subsection', 'end_use', 'carrier_name'], how='all')
+        df['section'] = df.section.ffill()
+        df['subsection'] = df.subsection.ffill()
+        df.loc[
```
timtroendle, line 63: You are doing four times the same here. Can you shorten this?

This is not super important. But the CARRIER_NAMES mapping in tertiary.py is more or less what's lacking here, and I find it very clear.

scripts/jrc-idees/tertiary.py
```diff
@@ -0,0 +1,93 @@
...
+    'Gases incl. biogas': 'gas',
+    'Geothermal energy': 'renewable_heat',
+    'Liquified petroleum gas (LPG)': 'oil',
+    'Solar': 'renewable_heat',
+    'Solids': 'solid_fossil'
+}
+
+
+def process_jrc_tertiary_data(data_dir, out_path):
+    data_filepaths = glob.glob(os.path.join(data_dir, "*.xlsx"))
```
timtroendle, line 37: this can also be 

```
import Path from pathlib

Path(data_dir).glob("*.xlsx")
``` 

it doesn't really matter of course, just a fyi in case you didn't know. I personally like pathlib's API much better.

scripts/jrc-idees/tertiary.py
```diff
@@ -0,0 +1,93 @@
...
+    processed_data = get_tertiary_sector_data(data_filepaths)
```
timtroendle, line 38: As above, this could be `pd.concat([get_tertiary_sector_data(file) for file in files])` and you'd get rid of the append-to-list-pattern and you'd replace it by a list comprehension.

scripts/jrc-idees/tertiary.py
```diff
@@ -0,0 +1,93 @@
...
+    processed_data.index = processed_data.index.set_levels(['twh'], level='unit')
+    processed_data.stack('year').to_csv(out_path)
+
+
+def get_tertiary_sector_data(files):
+    dfs = []
+
+    for file in files:
+        df_consumption = pd.read_excel(file, sheet_name='SER_hh_fec', index_col=0)
+        df_demand = pd.read_excel(file, sheet_name='SER_hh_tes', index_col=0)
```
timtroendle, line 49: Hmm, what ecaxtly is the difference between consumption and demand?

scripts/jrc-idees/transport.py
```diff
@@ -0,0 +1,168 @@
...
+        # After all this hardcoded cleanup, make sure numbers match up
+        assert np.allclose(df.sum(), tot.loc[df.columns].astype(float))
+
+        dfs.append(df)
+    return pd.concat(dfs)
+
+
+def remove_of_which(df, main_carrier, of_which_carrier):
+    """
+    Subfuels (e.g. biodiesel) are given as 'of which ...', meaning the main fuel consumption
```
timtroendle, line 150: So annoying when they do that. Nice for humans of course, terrible for automation.

scripts/jrc-idees/transport.py
```diff
@@ -0,0 +1,168 @@
...
+    "rail-distance": {
+        "sheet_name": "TrRail_act",
+        "idx_start_str": "Vehicle-km (mio km)",
+        "idx_end_str": "Stock of vehicles",
+        "unit": "mio_km"
+    },
+}
+
+
+def process_jrc_transport_data(data_dir, dataset, out_path):
```
timtroendle, line 46: A lot of the comments in `industry.py` apply here too.

timtroendle (changes requested):
Only very few comments about functionality, but a lot about structuring the code and readability.

I still think readability is important here. These kinds of data processing scripts are in most cases super difficult to read but structure and descriptions can ease the pain to some degree. You may not agree with all my comments and that's fine of course.

environment.yaml
```diff
@@ -5,4 +5,4 @@ channels:
 dependencies:
     - python=3.8
     - pycountry=18.12.8
-    - snakemake-minimal=5.8.2
+    - snakemake-minimal=6.1.1
```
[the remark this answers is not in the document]
brynpickering: To allow a wildcard in a 'script' call. I can't find the exact version where they made this fix in the changelog, but it doesn't work in 5.8.2 and does in 6.1.1... This may be an unnecessary update if I re-write some of the rules as per your suggestions later in this review.

lib/eurocalliopelib/utils.py
```diff
@@ -51,7 +51,7 @@ def to_numeric(series):
     Returns a numeric pandas.Series.
 
     """
-    series = series.astype(str).str.extract("(\-*\d+\.*\d*)")[0]
+    series = series.astype(str).str.extract("(\\-*\\d+\\.*\\d*)")[0]
```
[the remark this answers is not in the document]
brynpickering: It believe it did, I was fixing it due to a futurewarning only. I'll still add a test to be certain.

rules/data-processing.smk
```diff
@@ -36,3 +42,37 @@ rule annual_energy_balances:
...
+            if country_code in EU28
+        ]
+    params: sector_search = lambda wildcards: wildcards.sector.title()
+    output: temp(directory("build/data/jrc-idees/{sector}/unprocessed"))
+    conda: "../envs/shell.yaml"
+    shell: "unzip 'data/automatic/jrc-idees/*.zip' '*{params.sector_search}*' -d {output}"
+
+
+rule jrc_idees_processed:
+    message: "Process {wildcards.dataset} {wildcards.sector} data from JRC-IDEES to be used in understanding current and future {wildcards.sector} demand"
```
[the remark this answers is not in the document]
brynpickering: I don't need to, but thought it worked quite nicely to keep the number of rules down :) I will split it out to a rule per sector since that will also obviate the need for an update to >=6.0.0.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,127 @@
...
+from styleframe import StyleFrame
+
+from eurocalliopelib import utils
+
+idx = pd.IndexSlice
+
+SHEETS = ['ISI', 'NFM', 'CHI', 'NMM', 'PPA', 'FBT', 'TRE', 'MAE', 'TEL', 'WWP', 'OIS']
+
+
+def process_jrc_industry_data(data_dir, dataset, out_path):
```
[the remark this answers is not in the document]
brynpickering: I wouldn't say 'not a fan', but rather that when I wrote most of this code in a Jupyter notebook I was not thinking too much about the final structure, so didn't think to parcel everything up into subfunctions as I was going along. A lot of this is just a copy-paste job from the notebook.

I can improve legibility, why not ;)

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,127 @@
...
+        idx_end = [int(i) for i in style_df.index if 'Market shares' in str(style_df.loc[i, col])][0]
+        colors = {
+            'section': ['FFC00000'],
+            'subsection': ['FF0070C0', '4f6228', '953735'],
+            'end_use': ['984807'],
+            'carrier_name': ['808080', 'e46c0a', '000000', 'dc9e9c']
+        }
+
+        for k, v in colors.items():
+            idx = style_df[col].style.font_color.isin(v).loc[:idx_end]
```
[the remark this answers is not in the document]
brynpickering: yup.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,127 @@
...
+                str(i) if str(i) != 'nan' else np.nan
+                for i in style_df.loc[idx[idx == True].index, col]
+            ]
+        df.loc[:idx_end, 'indent'] = [int(i) for i in style_df[col].style.indent.loc[:idx_end]]
+
+        # To check at the end that we haven't lost some data
+        tot = df[df.indent == 1].sum()
+
+        df = df.dropna(subset=['section', 'subsection', 'end_use', 'carrier_name'], how='all')
+        df['section'] = df.section.ffill()
```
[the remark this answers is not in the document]
brynpickering: you mean with "ffill"?

scripts/jrc-idees/tertiary.py
brynpickering: Agreed, that is nicer! There had been another subdir in this list, which made it more worthwhile to join everything.

scripts/jrc-idees/tertiary.py
brynpickering: The efficiency losses in converting from a commodity (gas, electricity, coal, etc.) to an end-use (steam heat, space heat, machine operation, distance travelled, etc.).

scripts/jrc-idees/industry.py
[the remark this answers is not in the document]
timtroendle: Yes. The reason for this comment is the following: While most other things are just restructuring data, this line is manipulating data. It replaces nan with the last seen value, right? I don't really understand why this is fine, or what this is for. But if it's any important, it would be good to add a comment. If it's unimportant, just ignore my comment.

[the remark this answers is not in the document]
brynpickering: fair enough, although it still fits with restructuring data. I've left a comment nonetheless.

lib/tests/test_utils.py
```diff
@@ -114,4 +115,24 @@ def test_country_code_conversion_no_output_specified(input_country, iso3166):
...
+
+    def test_dash_to_NaN(self):
+        new_series = to_numeric(pd.Series(["-", 1, "-"]))
+        assert new_series.isna().sum() == 2
+        assert is_numeric_dtype(new_series)
+
+    @pytest.mark.parametrize("numeric_values", ([1, 1, 2], [10, 100, 1000], [1.0, 1.1, 1.2], [1, -1, 2.0]))
+    def test_remove_non_numeric_characters(self, numeric_values):
+        new_series = to_numeric(pd.Series([
+            "-", numeric_values[0], f"{numeric_values[1]}a", f"{numeric_values[2]}-"
```
timtroendle, line 130: I'd personally make this more explicit rather than computed. Something like

```Python
@pytest.mark.parametrize("raw", "expected (
     (["-", 1, "1a", "2-"], [np.nan, 1, 1, 2]
     (["-", 10, "100a", "1000-"], [np.nan, 10, 100, 1000]
))
def test(self, raw, expected):
     assert np.allclose(raw, expected)
```

I think that way it's much clearer what the intended functionality is.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,176 @@
...
+    df.loc[:last_index_of_data, 'indent'] = (
+        style_df[column_names].style.indent.loc[:last_index_of_data].astype(int)
+    )
+
+    # indent of 1 tab == high-level summed data.
+    # We keep this to check at the end that we haven't lost some data
+    total_to_check = df[df.indent == 1].sum()
+
+    df = df.dropna(subset=['section', 'subsection', 'end_use', 'carrier_name'], how='all')
+    # All data between section/subsection names correspond to the preceding section/subsection name
```
timtroendle, line 112: Aha! That comment really helped me in understanding this.

scripts/jrc-idees/transport.py
```diff
@@ -0,0 +1,188 @@
...
+def read_transport_excel(file, sheet_name, idx_start_str, idx_end_str, unit):
+    style_df = StyleFrame.read_excel(file, read_style=True, sheet_name=sheet_name)
+    df = pd.read_excel(file, sheet_name=sheet_name)
+    column_names = str(style_df.data_df.columns[0])
+    idx_start = int(style_df[style_df[column_names].str.find(idx_start_str) > -1].item())
+    idx_end = int(style_df[style_df[column_names].str.find(idx_end_str) > -1].item())
+    df = df.assign(indent=style_df[column_names].style.indent.astype(int)).loc[idx_start:idx_end]
+
+    total_to_check = df.iloc[0]
+    df['section'] = df.where(df.indent == 1).iloc[:, 0].ffill()
```
timtroendle, line 82: Maybe add that same comment from industry.py here?

scripts/jrc-idees/transport.py
```diff
@@ -0,0 +1,188 @@
...
+    df['vehicle_type'] = df.where(df.indent == 2).iloc[:, 0].ffill()
+
+    if sheet_name == 'TrRoad_act':
```
timtroendle, line 85: This if clause is redundant as you are already modelling the relationship between sheet name and dataset in DATSET_PARAMS. Maybe add the function to DATSET_PARAMS as a function object? The disadvantage is that you have to define the functions first and only then you can define DATSET_PARAMS. So nvm, not too important.

```Python
def fun1:
    pass

def fun2:
    pass

DATSET_PARAMS = {
     "road-vehicles": {
          "function": fun1
     }
}
```

timtroendle: But maybe DATSET_PARAMS = DATASET_PARAMS

timtroendle (approved):
Just a few more minor changes.

brynpickering:
Need to fix some bugs found on running the script before it's ready.

PR #136, opened 2021-07-20
Update source of energy balance translation CSVs

brynpickering:
Fixes lack of these expected CSVs in the existing workflow. 

This is linked to https://github.com/calliope-project/euro-calliope-datasets/pull/2, where the data being linked to here would be stored.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


config/default.yaml
```diff
@@ -4,8 +4,8 @@ data-sources:
     biofuel-costs: data/euro-calliope-datasets/biofuels/costs/{feedstock}.csv
     capacity-factors: https://zenodo.org/record/3899687/files/{filename}?download=1
     data-repository: https://raw.githubusercontent.com/calliope-project/euro-calliope-datasets/develop/{dataset}
```
timtroendle, line 6: You'll need to update the branch here.

timtroendle (approved):

PR #137, opened 2021-07-21
Add option to shed load

### LINKED ISSUE #131: Add load_shedding option
`load_shedding` is a technology that is included in the [default set of technologies](https://github.com/calliope-project/euro-calliope/blob/59e0951d89a4dbe60ad2c4aa39bf0e2967e25216/templates/renewable-techs.yaml#L89). This is error-prone and may be unnecessary.

It's error-prone because a user may not expect this technology to exist, leading to unintended results. Instead, it makes more sense to have this as an optional technology.

It may be unnecessary because it overlaps with [a built-in mechanism of Calliope](https://calliope.readthedocs.io/en/stable/user/building.html#allowing-for-unmet-demand).

Definition of done: It's either optional, or it's removed entirely.

timtroendle:
Fixes #131.

This option has existed before as a Calliope tech and override, but the override wasn't part of the pre-built, wasn't documented, wasn't tested.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * **ADD** installation of `curl` and `unzip` from conda-forge, to increase portability (#59).
 * **ADD** sync infrastructure to easily send and receive files to and from a cluster (#74).
 * **ADD** optional email notifications whenever builds fail or succeed (#92).
+* **ADD** option to shed load at all locations (#131).
 
 * **UPDATE** EEZ updated from v10 to v11 (difference in offshore area is < 1% for all relevant countries) (#99).
 * **UPDATE** ENTSOE national electricity load data gap filling priority order included in config (#42).
--- a/templates/README.md
+++ b/templates/README.md
@@ -53,6 +53,10 @@ When using the `directional-rooftop-pv` override, there are three instead of jus
 
 By default, euro-calliope allows capacities of run-of-river hydro, reservoir hydro, and pumped storage hydro capacities up to today's levels. Alternatively, it's possible to freeze these capacities to today's levels using the `freeze-hydro-capacities` override.
 
+> load-shedding
+
+Adds an option to shed load to each location. Shedding load comes with high a variable cost (see `tech-cost.yaml` parameter file) and will only be used when no other, less costly, option is available. Load shedding is not restricted in its total power or energy and has no fixed cost. If no load must be shed in your model (in cost-minimisation terms: if load shedding is economically unattractive), using this override will not change your results.
+
 ## Model components
 
 The models contain the following files. All files in the root directory are independent of the spatial resolution. All files that depend on the spatial resolution are within subfolders named by the resolution.
@@ -64,6 +68,7 @@ The models contain the following files. All files in the root directory are inde
 │   ├── electricity-demand.csv             <- Timeseries of electricity demand on each node.
 │   ├── example-model.yaml                 <- Calliope model definition.
 │   ├── link-all-neighbours.yaml           <- Connects neighbouring locations with transmission.
+│   ├── load-shedding.yaml                 <- Override adding option to shed load.
 │   ├── locations.csv                      <- Map from Calliope location id to name of location.
 │   └── locations.yaml                     <- Defines all locations and their max capacities.
 ├── build-metadata.yaml                    <- Metadata of the build process.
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * **ADD** installation of `curl` and `unzip` from conda-forge, to increase portability (#59).
 * **ADD** sync infrastructure to easily send and receive files to and from a cluster (#74).
 * **ADD** optional email notifications whenever builds fail or succeed (#92).
+* **ADD** option to shed load at all locations (#131).
```
brynpickering, line 15: Are we adding new items at the top or bottom of these lists? Some convention would be good (since I'm adding them at the top, usually :P)

templates/README.md
```diff
@@ -53,6 +53,10 @@ When using the `directional-rooftop-pv` override, there are three instead of jus
 
 By default, euro-calliope allows capacities of run-of-river hydro, reservoir hydro, and pumped storage hydro capacities up to today's levels. Alternatively, it's possible to freeze these capacities to today's levels using the `freeze-hydro-capacities` override.
 
+> load-shedding
+
+Adds an option to shed load to each location. Shedding load comes with high a variable cost (see `tech-cost.yaml` parameter file) and will only be used when no other, less costly, option is available. Load shedding is not restricted in its total power or energy and has no fixed cost. If no load must be shed in your model (in cost-minimisation terms: if load shedding is economically unattractive), using this override will not change your results.
```
brynpickering, line 58: I would be more descriptive about what a. load shedding is, and b. how this is a proxy for that, and c. why someone might want to use this override. 
a. Load shedding is effectively what causes a power system blackout/brownout.
b. We approximate this by a 'virtual supply' of electricity which can be injected into the system when no other electricity production is possible. 
c. It is useful when a user wants to capture the impact of low capacities in the system, which would otherwise lead to model infeasibility. It is a more explicit power-system take on Calliope's existing `ensure_feasibility` configuration option.

templates/README.md
```diff
@@ -64,6 +68,7 @@ The models contain the following files. All files in the root directory are inde
 │   ├── electricity-demand.csv             <- Timeseries of electricity demand on each node.
 │   ├── example-model.yaml                 <- Calliope model definition.
 │   ├── link-all-neighbours.yaml           <- Connects neighbouring locations with transmission.
+│   ├── load-shedding.yaml                 <- Override adding option to shed load.
```
brynpickering, line 71: I start to wonder about the structure of the YAMLs and whether *all* overrides/scenarios should be in their own file with the same name as the override/scenario. This could then even be inside a sub-dir of 'overrides'.

tests/resources/regional/model.yaml
```diff
@@ -18,6 +19,7 @@ scenarios:
     directional-pv: ["directional-rooftop-pv"]
     frozen-hydro: ["freeze-hydro-capacities", "no-hydro-fixed-cost"]
     alternative-cost: ["dea-renewable-cost", "schroeder-hydro-cost"]
+    shed-load: ["load-shedding"]
```
brynpickering, line 22: Does this include load shedding in the model in a particular model run? IIRC, the model tests are also solving the model for each of these scenarios. If yes, could there be a test to check that `load-shedding` has made its way into the model as a technology at every location?

brynpickering (approved):
Looks fine, just a few minor comments. RE the load-shedding rule itself, I think it would make much more sense to not have it depend on the `geo.yaml` env nor on `units.geojson` - it's quite overkill for what it needs (a list of model locations)! Maybe just depend on `locations.csv` from `rules.locations`?

templates/README.md
timtroendle: We'll discuss this in #145.

tests/resources/regional/model.yaml
timtroendle: All scenarios are tested, yes. But it's the simplest possible test only: the test makes sure the model runs and returns an optimal result. I'll see if I manage to write a test you are describing.

brynpickering: I think a very simple test would work:
1. if adding a new tech, check that it can be found associated with all locations in `model._model_data.loc_techs`. 
2. If adding a new constraint, check that the associated data is loaded in and the relevant technology can be found in it, e.g. `model._model_data.energy_cap_equals` should have hydro techs in it when using `freeze-hydro-capacities`.

brynpickering (approved):

PR #138, opened 2021-07-21
Fix moving of hydro stations by transforming to projection first

### LINKED ISSUE #55: Incorrect distance calculation between hydro stations and hydro basins
When a hydro station is not within any of the existing hydro basins we move it into the closest one. The rationale is that the station in fact _is_ within the basin, but that the data is imprecise.

There are two issues with the current approach:
1. We use a geographic CRS (WGS84) which does not lead to meaningful distances.
2. We do not test the assumption that distances are small.

We should fix it the following way:
1. Move to a projected CRS (EPSG3035).
2. Test that the distance is small. This should be controlled by a parameter from config with default 10km(or smaller?).

timtroendle:
Fixes #55.

This was not exactly broken, but let to warnings by geopandas. Further, the buffer size is now the same for all stations that must be moved and the buffer size is easier to understand.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -64,6 +64,7 @@ quality-control:
             - actual_entsoe_power_statistics
     hydro:
         scale-phs-according-to-geth-et-al: false
+        stations-buffer-size-km: 1
 sea-connections:
     continental: []
     national: # Source: https://www.entsoe.eu/data/map/
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -64,7 +64,7 @@ quality-control:
             - actual_entsoe_power_statistics
     hydro:
         scale-phs-according-to-geth-et-al: false
-        stations-buffer-size-km: 1
+        station-nearest-basin-max-km: 1
 sea-connections:
     continental: []
     national: # Source: https://www.entsoe.eu/data/map/
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -265,6 +265,9 @@ properties:
                     scale-phs-according-to-geth-et-al:
                         type: boolean
                         description: Whether or not to use pumped hydro capacities derived from Geth et al. (2015) to scale pumped hydro capacities givven by the JRC hydro database.
+                    stations-buffer-size-km:
+                        type: number
+                        description: Move stations outside of hydrobasins into closest basin by max this amount (fails otherwise).
     capacity_factors:
         type: object
         properties:
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -265,8 +265,9 @@ properties:
                     scale-phs-according-to-geth-et-al:
                         type: boolean
                         description: Whether or not to use pumped hydro capacities derived from Geth et al. (2015) to scale pumped hydro capacities givven by the JRC hydro database.
-                    stations-buffer-size-km:
+                    station-nearest-basin-max-km:
                         type: number
+                        minimum: 0
                         description: Move stations outside of hydrobasins into closest basin by max this amount (fails otherwise).
     capacity_factors:
         type: object
--- a/scripts/hydro/preprocess_hydro_stations.py
+++ b/scripts/hydro/preprocess_hydro_stations.py
@@ -4,12 +4,13 @@ from shapely.geometry import Point
 import pycountry
 
 from eurocalliopelib import utils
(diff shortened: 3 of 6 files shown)


config/default.yaml
```diff
@@ -64,6 +64,7 @@ quality-control:
             - actual_entsoe_power_statistics
     hydro:
         scale-phs-according-to-geth-et-al: false
+        stations-buffer-size-km: 1
```
brynpickering, line 67: `stations-buffer-radius-km`?

config/schema.yaml
```diff
@@ -265,6 +265,9 @@ properties:
                     scale-phs-according-to-geth-et-al:
                         type: boolean
                         description: Whether or not to use pumped hydro capacities derived from Geth et al. (2015) to scale pumped hydro capacities givven by the JRC hydro database.
+                    stations-buffer-size-km:
+                        type: number
```
brynpickering, line 269: Set a min value to zero.

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -118,28 +119,32 @@ def fix_station_country_code(stations):
...
 
 
-def ensure_stations_within_basins(stations, path_to_basins, buffer_size):
-    hydrobasins = gpd.read_file(path_to_basins)
-    is_in_basin = stations.apply(station_in_any_basin(hydrobasins), axis="columns")
-    ill_placed_stations = stations[~is_in_basin]
+def ensure_stations_within_basins(stations, path_to_basins, buffer_size_m):
+    geo_stations = gpd.GeoDataFrame(
+        stations.copy(),
+        geometry=[Point(station["lon"], station["lat"]) for _, station in stations.iterrows()],
```
brynpickering, line 129: `gpd.points_from_xy(stations["lon"], stations["lat"])`

scripts/hydro/preprocess_hydro_stations.py
```diff
@@ -118,28 +119,32 @@ def fix_station_country_code(stations):
...
+    is_in_basin = geo_stations.apply(station_in_any_basin(hydrobasins), axis="columns")
+    ill_placed_stations = geo_stations[~is_in_basin]
     for station_id, station in ill_placed_stations.iterrows():
-        new_point = new_coords(station, buffer_size, hydrobasins)
-        stations.loc[station_id, "lon"] = new_point.coords[0][0]
-        stations.loc[station_id, "lat"] = new_point.coords[0][1]
-    assert stations.apply(station_in_any_basin(hydrobasins), axis="columns").all()
+        geo_stations.loc[station_id, "geometry"] = new_coords(station, buffer_size_m, hydrobasins)
+    assert geo_stations.apply(station_in_any_basin(hydrobasins), axis="columns").all()
+    stations["lon"] = geo_stations.to_crs(WGS84).geometry.apply(lambda point: point.coords[0][0])
```
brynpickering, line 138: I may be misremembering, but does `point.x` or `point.coords.x` not exist? Would be more understandable than `[0][0]`

brynpickering: In fact, maybe it is just `geo_stations.to_crs(WGS84).geometry.x` or `geo_stations.to_crs(WGS84).geometry.centroid.x`...

config/default.yaml
brynpickering: Maybe the word 'buffer' isn't even needed, since it's a geopandas thing and not really informative to the user. `station-nearest-basin-max-km` or something?

brynpickering (approved):
Some minor comments on naming and making the code more clear. 

 I wonder if a test for the main geometric manipulation would be worthwhile?

timtroendle:
I've tested this with the full configuration.

brynpickering:
The whole thing builds a model, but do we know for sure that the stations are being moved into the expected basins based on the buffer?

timtroendle:
Hm I think we do. The buffer is 1km which means stations can be moved by 1km only. If there's no basin 1km away, the workflow fails. (I've tested this through setting the buffer to 0.1km).

Does that answer your question?

brynpickering:
> Does that answer your question?

Well enough!

Could you add a CHANGELOG entry for the config option? Then I'm happy for it to be merged.

PR #139, opened 2021-07-21
Add override constraining e to p ratios

### LINKED ISSUE #130: Add `stylised-storage` override
Previous studies constrained the energy to power ratios of battery and hydrogen. This constraint should be added as an optional override.

In [previous studies](https://github.com/timtroendle/money-land/blob/13c7caa6da34262461d158fcd85b82c069c772c7/src/template/model.yaml#L41), it looked like the following:

```yaml
overrides:
    stylised-storage:
        # Transform battery and hydrogen to stylised short term
        # and long term storage solutions.
        techs:
            battery:
                constraints:
                    energy_cap_per_storage_cap_min: 0.25 # max 4h full-power discharge
                costs:
                    monetary:
                        om_prod: 0 # ASSUME 0 instead of 0.000378 €/kWh, as the tiny number causes numerical issues
            hydrogen:
                constraints:
                    energy_cap_per_storage_cap_max: 0.25 # min 4h full-power discharge
```

timtroendle:
Fixes #130.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * **ADD** automatic download of EEZ (#99).
 * **ADD** Danish Energy Agency and Schroeder et al (2013) cost data as well as no hydro fixed costs as optional overrides (#18, #129).
+* **ADD** override to constrain energy to power ratios of battery and hydrogen storage (#130).
 * **ADD** Add basic documentation to ReadTheDocs (#114).
 * **ADD** ability to move working directory (#45).
 * **ADD** schema that automatically validates configuration files (#45).
--- a/templates/README.md
+++ b/templates/README.md
@@ -49,6 +49,11 @@ By default, euro-calliope contains a single technology for rooftop PV. This tech
 
 When using the `directional-rooftop-pv` override, there are three instead of just one technologies for rooftop PV. The three technologies comprise (1) south-facing and flat rooftops, (2) east- and west-facing rooftops, and (3) north-facing rooftops. This leads to higher capacity factors of rooftop PV as long as the potential of rooftop PV is not fully exploited. However, this also increases the complexity of the model.
 
+> exclusive-e-to-p-ratios
+
+Constrains the energy to power ratios of battery and hydrogen storage in a way that they do not overlap. Battery storage is constrained to a ratio of ≤4h while hydrogen is constrained to a ratio of ≥4h. The ratio is derived from typical values of commercial lithium-ion batteries available today (2021). Constraining hydrogen storage as well ensures  it does not directly compete with battery storage but is used for longer durations of fours hours and longer.
+
+
 > freeze-hydro-capacities
 
 By default, euro-calliope allows capacities of run-of-river hydro, reservoir hydro, and pumped storage hydro capacities up to today's levels. Alternatively, it's possible to freeze these capacities to today's levels using the `freeze-hydro-capacities` override.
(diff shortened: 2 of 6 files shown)


templates/README.md
```diff
@@ -49,6 +49,11 @@ By default, euro-calliope contains a single technology for rooftop PV. This tech
 
 When using the `directional-rooftop-pv` override, there are three instead of just one technologies for rooftop PV. The three technologies comprise (1) south-facing and flat rooftops, (2) east- and west-facing rooftops, and (3) north-facing rooftops. This leads to higher capacity factors of rooftop PV as long as the potential of rooftop PV is not fully exploited. However, this also increases the complexity of the model.
 
+> exclusive-e-to-p-ratios
```
brynpickering, line 52: `energy-to-power` isn't a massive increase in length, but makes the whole thing much more readable.

templates/README.md
```diff
@@ -49,6 +49,11 @@ By default, euro-calliope contains a single technology for rooftop PV. This tech
...
+
+Constrains the energy to power ratios of battery and hydrogen storage in a way that they do not overlap. Battery storage is constrained to a ratio of ≤4h while hydrogen is constrained to a ratio of ≥4h. The ratio is derived from typical values of commercial lithium-ion batteries available today (2021). Constraining hydrogen storage as well ensures  it does not directly compete with battery storage but is used for longer durations of fours hours and longer.
```
brynpickering, line 54: maybe mention that `energy` here refers to stored energy and `power` to charge/discharge capacity. Final sentence: `Constraining hydrogen storage as well ensures it does not directly compete with battery storage, but is used instead for durations of fours hours and longer.`.

templates/storage-techs.yaml
```diff
@@ -23,3 +23,13 @@ techs:
         constraints:
             energy_cap_max: inf
             storage_cap_max: inf
+overrides:
+    exclusive-e-to-p-ratios:
+        # Ensures exclusive energy to power ratios of battery (≤4h) and hydrogen (≥4h) storage.
+        techs:
+            battery:
```
brynpickering, line 30: These could be one-liners: `battery.constraints.energy_cap_per_storage_cap_min: 0.25`

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * **ADD** automatic download of EEZ (#99).
 * **ADD** Danish Energy Agency and Schroeder et al (2013) cost data as well as no hydro fixed costs as optional overrides (#18, #129).
+* **ADD** override to constrain energy to power ratios of battery and hydrogen storage (#130).
```
brynpickering, line 7: Now it's directly in the middle of the others! Also, capitalisation needed.

brynpickering (approved):
Simple enough; all makes sense. Do you see this being a problem in the model, that batteries operate on cycles > a few days and hydrogen < 1 week? I would have thought that the self-discharge of batteries as well as their lower ratio of energy capacity to storage capacity costs would make it unattractive, without the need for an explicit constraint. The existence of this override would lead me (and I guess others) to believe that the model rarely behaves this way...

timtroendle:
It's true, it may not be needed. But that depends on cost parameters of storage and other options.

I've simply always used it in the past and I always copy pasted it. That's why I am including it here. It doesn't really do harm, does it?

brynpickering:
No, certainly no harm, except that we begin to creep up in the number of possible overrides and thus the complexity of the contents across the various YAML files. Moving forward we should probably consider whether to bake some of these things in with config params (i.e. the `battery` and `hydrogen` tech templates have this added to them) or to better structure the YAML files (as I mention in #137) so that it's easier to inspect the list of overrides and differentiate them from the base Calliope config.

PR #140, opened 2021-07-21
Fix syncignore to ignore data/euro-calliope-datasets

timtroendle:
This tiny bug has been introduced in #112 which implemented #104.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.syncignore-send
+++ b/.syncignore-send
@@ -8,4 +8,5 @@ __pycache__
 .DS_Store
 build
 data/automatic
+data/euro-calliope-datasets
 notebooks


brynpickering (commented):
I have all sorts of problems with the data folder already. Large datasets can only really exist on the remote instance, but not having `data` completely ignored leads to them being deleted unless they are in one of these two sub-dirs. I would just make this `data`, unless there is some major reason not to...

timtroendle:
I understand that this can be a problem, but the opposite can be true as well: for small datasets, you'd usually want to do exactly that. You want to put them into `data` and then sync them to the cluster. 

Would you be fine with leaving this up to the user? For `data/automatic` and `data/euro-calliope-datasets` I guess we both agree they should be ignored, so they should be mentioned explicitly. For other data that you have on Euler only and that you don't want to sync you can always ignore `data` entirely, or these files or subfolders specifically `data/my-folder-with-large-files`.

brynpickering (approved):

PR #141, opened 2021-07-22
Fix download of EEZ to happen on local (login) node

timtroendle:
Fixes an issue introduced in #99.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/rules/shapes.smk
+++ b/rules/shapes.smk
@@ -14,6 +14,7 @@ SCHEMA_UNITS = {
 
 configfile: "config/default.yaml"
 localrules: download_raw_gadm_administrative_borders, raw_gadm_administrative_borders, download_raw_nuts_units
+localrules: download_eez
 root_dir = config["root-directory"] + "/" if config["root-directory"] not in ["", "."] else ""
 script_dir = f"{root_dir}scripts/"
 


brynpickering (commented):
Hmm, this has never been a problem for me. Do you get errors when using job nodes for downloads or is this just your preference? I just wonder whether it's a good design choice to have downloads on the login node. Parallelising them would speed things up.

timtroendle:
Yes, this fails in my setup as compute nodes have no internet access. I believe you said once it's possible to change that, but I'd rather not if not necessary.

It's true that downloads on compute nodes could be faster as they could be parallelised. But you can also parallelise downloads on the login node using `--local-cores = [number greater 1]`.

In any case, all download rules right now are local. So we should be consistent either way.

rules/shapes.smk
```diff
@@ -14,6 +14,7 @@ SCHEMA_UNITS = {
 
 configfile: "config/default.yaml"
 localrules: download_raw_gadm_administrative_borders, raw_gadm_administrative_borders, download_raw_nuts_units
+localrules: download_eez
```
brynpickering, line 17: Is there a reason this is on a new line? I would find it easier to understand if the first line were to become a multi-line listing of localrules.

brynpickering (approved):

brynpickering:
OK, would be good to think of a section in the documentation on how best to leverage snakemake args in Euro-Calliope (since their documentation is difficult to navigate, I find). Things like `--local-cores` could then be mentioned there.

rules/shapes.smk
timtroendle: Hmm the first line is too long. There may be other ways of splitting this into several lines with just a single `localrules` call, but I am not exactly sure how.

timtroendle:
Good idea. I've opened issue #144 for that.

PR #142, opened 2021-07-22
Update Snakemake to use mamba by default

timtroendle:
## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,9 @@
 * **UPDATE** JRC hydro database v4 -> v7 (#48). This entails one patch being removed:
     1. Romanian PHS data is no longer manually added from Geth et al. (2015).
 * **UPDATE** Improve gap-filling method for national electricity load data (#3).
-* **UPDATE** dependencies (#44):
+* **UPDATE** dependencies (#44, #142):
     * Python 3.7 -> 3.8
+    * Snakemake 5.8.2 -> 6.1.1 (uses mamba by default)
     * atlite -> 0.2.1
     * geo packages from gdal 2.4 -> 3.2.1
     * Updates to NumPy, Pandas, xarray, and others
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ snakemake --use-conda --conda-frontend conda --list # test your installation
 
 Because input data is large, the actual model including its data is not part of this repository. To use the model, you first need to build it from input data and scripts. Running the build step will build the model in the `./model` folder.
 
-    snakemake --use-conda --cores all
+    snakemake --use-conda --cores <N_CORES>
+
+`N_CORES` is the number of cores of your machine you want to use. It can be anything between `1` and `all`. Please have a look at [Snakemake's documentation](https://snakemake.readthedocs.io) for more information.
 
 ## Build the model on a cluster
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,8 +20,9 @@
 * **UPDATE** JRC hydro database v4 -> v7 (#48). This entails one patch being removed:
     1. Romanian PHS data is no longer manually added from Geth et al. (2015).
 * **UPDATE** Improve gap-filling method for national electricity load data (#3).
-* **UPDATE** dependencies (#44):
+* **UPDATE** dependencies (#44, #142):
     * Python 3.7 -> 3.8
+    * Snakemake 5.8.2 -> 6.1.1 (uses mamba by default)
```
brynpickering, line 25: Is this a dangerous jump in version, as mentioned in #133? What are the implications of this change? Also, why not move even further than this, since snakemake is now version `6.6.1`? Would be good to have an understanding of what functionality we gain overall in changing version.

timtroendle: Haha fair. Well, it mentions the main functionality we gain and the rationale for this PR, right? There is a whole lot more of course, but we cannot list everything here.

I've ran the workflow on Euler yesterday and it just took way too long to create the environments, that's why I made this change. 

We _could_ of course move to the newest `6.6.1` version right away. I don't know which one is better. Moving to `6.1.1` is me being conservative with Snakemake's rather chaotic release management. I've been testing and using `6.1.1` for a while, so it seems to work fine. I cannot say anything about `6.6.1`.

brynpickering: If you're happy that it doesn't break anything then I think it's fine. Does the minimal/full workflow run just fine with 6.1.1? Looking at the [snakemake changelog](https://snakemake.readthedocs.io/en/stable/project_info/history.html) I would say that testing >6.4.0 might be worthwhile for some additional functionality it could provide, but that's obviously outside the scope of this PR.

brynpickering (approved):

brynpickering (changes requested):
Just thought about this: could you add mamba as a dependency in the `environment.yaml` or mention in the README that a user should have it installed? I assume that many will run the whole thing without realising they could realise improvements by just having mamba available.

CHANGELOG.md
timtroendle: Yes. Locally, I've been using 6.1.1. since its release in April.

timtroendle:
Good point. Especially because Snakemake now requires to define cores (WHY?) and so I also updated all Snakemake calls in the README.

brynpickering (commented):
Is `--cores all` a sensible thing to suggest, or should we say e.g. `--cores 1` so that people don't accidentally hang their local machines by snakemake taking over all possible processing power? I don't really mind either way, but I tend to use `--cores 1` for `send/receive` operations.

timtroendle:
I agree. My rationale was that I don't won't to confuse reader into thinking they _must_ use 1 core, not more. It's absolutely possible to always use all available cores, and it doesn't do much harm when run on a local machine.

Do you have an idea how to communicate better that `--cores` can be anything the user wants?

brynpickering:
Would replacing `all` with `N_CORES` work? Then just having a sentence to say that `N_CORES` can be anything between 1 and 'all'?

brynpickering (approved):

PR #147, opened 2021-07-22
Fix determination of location centroids

timtroendle:
## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/scripts/locations.py
+++ b/scripts/locations.py
@@ -4,6 +4,7 @@ import pandas as pd
 import geopandas as gpd
 
 from eurocalliopelib import filters
+from eurocalliopelib.geo import WGS84, EPSG3035
 
 TEMPLATE = """locations:
     {% for id, location in locations.iterrows() %}
@@ -71,7 +72,7 @@ def construct_locations(path_to_shapes, path_to_land_eligibility_km2, path_to_hy
     )
     locations = (
         locations
-        .assign(centroid=locations.centroid.rename("centroid"))
+        .assign(centroid=locations.to_crs(EPSG3035).centroid.to_crs(WGS84).rename("centroid"))
         .loc[:, ["name", "centroid"]]
     )
     capacities = _from_area_to_installed_capacity(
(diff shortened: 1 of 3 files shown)


scripts/locations.py
```diff
@@ -71,7 +72,7 @@ def construct_locations(path_to_shapes, path_to_land_eligibility_km2, path_to_hy
     )
     locations = (
         locations
-        .assign(centroid=locations.centroid.rename("centroid"))
+        .assign(centroid=locations.to_crs(EPSG3035).centroid.to_crs(WGS84).rename("centroid"))
```
brynpickering, line 75: Is there a reason to reproject to `WGS84` afterwards? Seems we have more use for it in `EPSG3035` in general...

tests/model/test_model.py
```diff
@@ -12,6 +12,7 @@
     "default": DEFAULT_TECHNOLOGIES,
     "connected": DEFAULT_TECHNOLOGIES,
     "directional-pv": (DEFAULT_TECHNOLOGIES | DIRECTIONAL_PV) - set(["roof_mounted_pv"]),
+    "e-to-p-ratios": DEFAULT_TECHNOLOGIES,
```
brynpickering, line 15: just sneak that in ;)

brynpickering (approved):

scripts/locations.py
timtroendle: Thought so too, but Calliope requires WGS84, doesn't it?

tests/model/test_model.py
timtroendle: Haha, yes. It was because the test and the override were on parallel branches and didn't know anything about each other. (If we had automatic tests that would have been detected automatically 😒).

scripts/locations.py
brynpickering: No, you can provide info in `lat`/`lon` or `x`/`y`

timtroendle: As discussed just now, we'll leave it in lat/lon. Moving to x/y may have unintended side-effects and will bring not much benefit -- if at all.

Issue #145, opened 2021-07-22
Re-structure final YAML files


brynpickering:
As evidenced by recent PRs (#131, #139), there are many possible Calliope overrides being built in the workflow. These sometimes end up in their own file, sometimes in a file with technology definitions. We should consider re-structuring the YAML files with two key aims:

1. Modularisation. It should be possible to build the set of YAML files with any future combination of technologies/carriers in the system, without the content of the YAML files themselves changing much. 
2. Ease of tweaking in a pre-built model. There will be a set of users working directly with the pre-built model and never with this workflow. It should be intuitive for users to navigate the YAML files to update/add/remove components of the model as they see fit. This also fits for those elements that are not configurable in the Euro-Calliope Snakemake workflow, but a user could change once the model is built (e.g. in #139, a user might want to change from 4hrs being the threshold between the technology types).

Some thoughts on this:
1. Technologies and their overrides and allocation to locations could be packaged on a per-YAML file basis.
2. Overrides could be in a file of the same name as the override, all kept in an override directory.

timtroendle:
I added this to 1.2 because this is a massive change and it gets worse the longer we wait.

brynpickering:
This is what the file structure I propose in #212 looks like with the current Euro-Calliope:

![image](https://user-images.githubusercontent.com/17178478/151200548-752f48be-165e-4645-bc68-4cb15ab61fb8.png)

Each technology YAML file contains the technology definitions, any overrides, and the allocation of the technologies to locations, e.g.:
<details>
<summary>supply-hydro.yaml</summary>
<br>

```yaml
techs:
    hydro_run_of_river:  # from [@JRC:2014] Table 14
        essentials:
            name: Run of river hydro electricity
            parent: supply
            carrier: electricity
        constraints:
            resource: file=supply/capacityfactors-hydro-ror.csv
            resource_unit: energy_per_cap
            lifetime: 60
        costs.monetary:
            energy_cap: 562.0 # (10,000 EUR2013/MW)
            om_annual: 16.86 # (10,000 EUR2013/MW) 3% of CAPEX
            om_prod: 0.0005 # (10,000 EUR2013/MW)
    hydro_reservoir:  # from [@JRC:2014] Table 12
        essentials:
            name: Hydro electricity with a reservoir.
            parent: supply_plus
            carrier: electricity
        constraints:
            resource: file=supply/capacityfactors-hydro-reservoir-inflow.csv
            resource_unit: energy_per_cap
[...]

PR #149, opened 2021-07-26
Add TYNDP data processing & link generation

### LINKED ISSUE #61: Optionally set transmission capacities based on ENTSOE TYNDP
A possible dataset to depend on for transmission capacities is the ENTSOE TYNDP scenarios. The [2020 scenario data](https://2020.entsos-tyndp-scenarios.eu/wp-content/uploads/2020/06/TYNDP-2020-Scenario-Datafile.xlsx.zip) includes export and import capacities between countries / sub-national transmission markets for the scenario years 2025, 2030, and 2040. It should be possible to extract this data and optionally create international transmission capacity limits.

brynpickering:
Fixes #61.

As discussed @timtroendle, this has not been added as an override, but rather a 'hanging' file. That is, it will be generated but not pointed to by `national/model.yaml`. A user has to update `model.yaml` to enforce these links.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -25,7 +25,7 @@ include: "./rules/shapes.smk"
 include: "./rules/hydro.smk"
 include: "./rules/sync.smk"
 localrules: all, download_raw_load, model, clean, parameterise_template, download_potentials, download_eurocalliope_dataset
-localrules: download_capacity_factors_wind_and_solar
+localrules: download_capacity_factors_wind_and_solar, download_enstoe_tyndp_zip
 configfile: "config/default.yaml"
 validate(config, "config/schema.yaml")
 wildcard_constraints:
@@ -275,6 +275,41 @@ rule electricity_load:
     script: "scripts/load.py"
 
 
+rule download_enstoe_tyndp_zip:
+    message: "Download ENSTOE ten-year network development plan (TYNDP) 2020 scenario dataset"
+    params: url = config["data-sources"]["entsoe-tyndp"]
+    output: protected("data/automatic/raw-entsoe-tyndp.xlsx.zip")
+    conda: "envs/shell.yaml"
+    shell: "curl -sLo {output} '{params.url}'"
+
+
+rule entsoe_tyndp_xlsx:
+    message: "Unzip ENTSO-E TYNDP 2020 scenario dataset."
+    input: rules.download_enstoe_tyndp_zip.output[0]
+    shadow: "minimal"
+    output: "build/data/TYNDP-2020-Scenario-Datafile.xlsx",
+    conda: "envs/shell.yaml"
+    shell: "unzip -o {input} 'TYNDP-2020-Scenario-Datafile.xlsx' -d build/data"
+
+
+rule entsoe_tyndp_links:
+    message: "Create YAML file of national-scale links with ENTSO-E TYNDP net-transfer capacities"
+    input:
+        script = script_dir + "link_entsoe_tyndp.py",
+        units = rules.units_without_shape.output[0],
+        entsoe_tyndp = rules.entsoe_tyndp_xlsx.output[0]
+    params:
+        entsoe_scenario = config["parameters"]["entsoe-tyndp"]["scenario"],
+        entsoe_grid = config["parameters"]["entsoe-tyndp"]["grid"],
+        entsoe_ntc_limit = config["parameters"]["entsoe-tyndp"]["entsoe_ntc_limit"],
+        energy_cap_limit = config["parameters"]["entsoe-tyndp"]["energy_cap_limit"],
+        entsoe_year = config["parameters"]["entsoe-tyndp"]["projection-year"],
+        scaling_factor = config["scaling-factors"]["power"]
+    output: "build/model/{resolution}/enstoe-tyndp-links.yaml"
+    conda: "envs/default.yaml"
+    script: "scripts/link_entsoe_tyndp.py"
+
+
 rule link_neighbours:
     message: "Create links between all direct neighbours on {wildcards.resolution} resolution."
     input:
@@ -322,6 +357,7 @@ rule model:
             technology=ALL_WIND_AND_SOLAR_TECHNOLOGIES
         ),
         rules.build_metadata.output,
+        lambda wildcards: rules.entsoe_tyndp_links.output if wildcards.resolution == "national" else [],
         example_model = template_dir + "example-model.yaml"
     output:
         log = "build/logs/{resolution}/model.done",
--- a/scripts/link_entsoe_tyndp.py
+++ b/scripts/link_entsoe_tyndp.py
@@ -1,0 +1,99 @@
+
+
+import pandas as pd
+import jinja2
+
+from eurocalliopelib import filters, utils
+
+TEMPLATE = """
+# {{ entsoe_ntc_limit.title() }} net transfer capacity between countries according to the
+# following ENTSO-E ten-year network development plan 2020 scenario:
+# Scenario: {{ entsoe_scenario.title() }}
+# Case: {{ entsoe_grid.title() }} Grid
+# Year: {{ entsoe_year }}
+# Climate Year: 2007
+
+links:
+    {% for link in ntcs.index %}
+    {{ link[0] }},{{ link[1] }}:
+        techs.ac_transmission.constraints:
+            energy_cap_{{ energy_cap_limit }}: {{ ntcs.loc[link] * scaling_factor }}  # {{ (1 / scaling_factor) | unit("MW") }}
+    {% endfor %}
+"""
+
+def link_tyndp(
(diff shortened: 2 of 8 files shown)


timtroendle (changes requested):
Very nice! Some minor comments below.

One major comment: there should be tests. There should be an integration test testing that the model runs and Gurobi finds an optimum (very easy to implement, see my comment) but there should also be a few tests testing for actual capacity constraint numbers. Those tests could be on the integration test level (i.e. the Calliope model object)  or on the unit-test level. I'd prefer the latter (or both).

Snakefile
```diff
@@ -25,7 +25,7 @@ include: "./rules/shapes.smk"
 include: "./rules/hydro.smk"
 include: "./rules/sync.smk"
 localrules: all, download_raw_load, model, clean, parameterise_template, download_potentials, download_eurocalliope_dataset
-localrules: download_capacity_factors_wind_and_solar
+localrules: download_capacity_factors_wind_and_solar, download_enstoe_tyndp_zip
```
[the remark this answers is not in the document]
brynpickering: Agreed. As in #145 for YAMLs, this should probably follow some clear modularisation effort.

scripts/link_entsoe_tyndp.py
```diff
@@ -0,0 +1,99 @@
...
+    # Capacity is in MW
+    tyndp_import = _split_links_in_index(tyndp_scenarios[tyndp_scenarios.Parameter == "Import Capacity"])
+    tyndp_export = _split_links_in_index(tyndp_scenarios[tyndp_scenarios.Parameter == "Export Capacity"])
+
+    # Some NTCs are different depending on whether it is import or export between countries.
+    # Here, we take either the minimum or maximum NTC of a link, depending on what a user defines for `entsoe_ntc_limit`
+    average_ntc = getattr(pd.concat([tyndp_import, tyndp_export]).abs(), entsoe_ntc_limit)(level=["loc_from", "loc_to"])
+
+    # Only keep countries found in the model
+    return average_ntc[
```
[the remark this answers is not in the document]
brynpickering: Sadly, `xs` doesn't work with multiple values per level accessed. But I can probably use the pandas `IndexSlice`.

scripts/link_entsoe_tyndp.py
```diff
@@ -0,0 +1,99 @@
+
+
```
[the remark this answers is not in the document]
brynpickering: How much is the fix offset by the energy consumption of your comment?

brynpickering:
@timtroendle I've addressed your issues, but now the minimal workflow is failing due to 'electrical islands' in the generation of `build/model/national/link-all-neighbours.yaml`. I did change the pointer to the CRS in this file, but having tested it with reverting the CRS string, the same issue exists. Any ideas on why this might be happening?

Snakefile
```diff
@@ -275,6 +275,43 @@ rule electricity_load:
...
+    params:
+        scenario = config["parameters"]["entsoe-tyndp"]["scenario"],
+        grid = config["parameters"]["entsoe-tyndp"]["grid"],
+        ntc_limit = config["parameters"]["entsoe-tyndp"]["ntc_limit"],
+        energy_cap_limit = config["parameters"]["entsoe-tyndp"]["energy_cap_limit"],
+        year = config["parameters"]["entsoe-tyndp"]["projection-year"],
+        scaling_factor = config["scaling-factors"]["power"]
+    output: "build/model/{resolution}/entsoe-tyndp-links.yaml"
+    wildcard_constraints:
+        resolution = "national"
```
timtroendle, line 310: Why do you make it a wildcard if there is anyway just a single value? Why not 

`output: "build/model/national/entsoe-tyndp-links.yaml"`?

templates/README.md
```diff
@@ -66,6 +66,14 @@ In euro-calliope, we model load shedding not as actual reduction of demand but a
 
 Calliope provides a built-in mechanism that is similiar: [`ensure-feasibility`](https://calliope.readthedocs.io/en/stable/user/building.html#allowing-for-unmet-demand). The benefit of using the `load-shedding` override over Calliope's built-in mechanism is that it is more targeted towards modelling shedding of electrical load and provides more flexibility -- for example in terms of the cost of shed load.
 
+## Manipulating the model using file imports
+
+The `example-model.yaml` configuration file in each resolution sub-directory (e.g. `national/model.yaml`) specifies a list of other configuration files to bring together to describe the model. This list can be changed by the modeller to select a combination of different configuration files. The final result is similar to the use of overrides, except that that model is *never* aware of the configuration being overridden.
```
timtroendle, line 71: `national/model.yaml` -> `national/example-model.yaml`

tests/resources/national/model.yaml
```diff
@@ -15,16 +15,20 @@ model:
...
+    connected_all_neighbours: ["connect_all_neighbours"]
+    connected_entsoe_tyndp: ["connect_entsoe_tyndp"]
     directional-pv: ["directional-rooftop-pv"]
     e-to-p-ratios: ["exclusive-energy-to-power-ratios"]
     frozen-hydro: ["freeze-hydro-capacities", "no-hydro-fixed-cost"]
     alternative-cost: ["dea-renewable-cost", "schroeder-hydro-cost"]
     shed-load: ["load-shedding"]
 overrides:
-    connect:
+    connect_all_neighbours:
```
timtroendle, line 26: Ok. For consistency, can you rename it in `tests/resources/regional/model.yaml`, too?

templates/README.md
```diff
@@ -66,6 +66,14 @@ In euro-calliope, we model load shedding not as actual reduction of demand but a
...
+
+> Transmission links
+
+There are no transmission links specified in `example-model.yaml` by default. At all resolutions, `link-all-neighbours.yaml` can be added to link neighbouring regions + a selection of pre-defined sub-sea links, but with no capacity limits. At the national resolution, transmission links can be set based on an ENTSO-E ten-year development plan 2020 scenario (`national/entsoe-tyndp-links.yaml`). The ENTSO-E links define all existing and planned international connections, including their predicted net transfer capacities (NTCs).
```
timtroendle, line 75: Hmm, that's not correct, is it? The example model imports `link-all-neighbours.yaml` and therefore, by default, all neighbours are connected.

tests/scripts/test_link_entsoe_tyndp.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import pandas as pd
+import numpy as np
+
+from scripts.link_entsoe_tyndp import _split_links_in_index
+
+@pytest.fixture
+def sliced_df():
+    def _sliced_df(idx):
+        data = list(range(10, 10 + len(idx)))
```
timtroendle, line 10: You don't need to convert this to a list here, you should be able to hand over data as an iterator.

tests/scripts/test_link_entsoe_tyndp.py
```diff
@@ -0,0 +1,25 @@
...
+        [["UK00-IE00", "UK00-FR00"], [("GBR", "IRL"), ("GBR", "FRA")], [10, 11]],
+        [["ITN0-CH00", "ITN1-CH00"], [("ITA", "CHE")], [21]],
+        [["ITN0-CH00", "ITN1-CH00", "ITN0-ITN1"], [("ITA", "CHE")], [21]],
+    )
+)
+def test_split_links_in_index(sliced_df, input_idx, expected_idx, expected_values):
+    input_df = sliced_df(input_idx)
+    cleaned_series = _split_links_in_index(input_df)
+    assert cleaned_series.index.symmetric_difference(expected_idx).empty
+    assert np.allclose(cleaned_series.loc[expected_idx].values, expected_values)
```
timtroendle, line 25: I thought about this for a while, but I still cannot figure out what you are testing in this second assert statement. I don't get what the expected values are and why.

timtroendle (approved):
Good!

I think there is a mistake in the documentation that needs to be fixed, but apart from that looks good.

scripts/link_entsoe_tyndp.py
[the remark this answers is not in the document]
timtroendle: 0.0000000000001 kg CO₂. The break-even will be in 10 years time. But we _really_ need to end this conversation as we are moving it more and more into the future.

Snakefile
brynpickering: Because snakemake doesn't like it. I can update the pointer from the `model` rule to be to the filepath rather than the rule output, which would solve this. However, it seems good practice to point to the rule output rather than the filepath... Do you prefer I change to the filepath in the `model` rule?

timtroendle: Ah, I see. Well I'd change from rule to filepath and remove the wildcard. But the current way works too of course.

tests/scripts/test_link_entsoe_tyndp.py
brynpickering: Let me know if the latest update is clearer

timtroendle: It is!

PR #150, opened 2021-07-26
Update Calliope to 0.6.6

### LINKED ISSUE #73: Update to Calliope 0.6.7
Changes since 0.6.5:
https://calliope.readthedocs.io/en/stable/history.html#id1

timtroendle:
Fixes #73.

## Checklist

- [ ] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * **ADD** optional email notifications whenever builds fail or succeed (#92).
 * **ADD** option to shed load at all locations (#131).
 
+* **UPDATE** Calliope version from 0.6.5 to [0.6.6.post1](https://calliope.readthedocs.io/en/stable/history.html#id1) (#73).
 * **UPDATE** EEZ updated from v10 to v11 (difference in offshore area is < 1% for all relevant countries) (#99).
 * **UPDATE** ENTSOE national electricity load data gap filling priority order included in config (#42).
 * **UPDATE** IRENA hydro generation data from 2018 to 2020 (#40).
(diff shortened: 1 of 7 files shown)


timtroendle:
While this works, there are staggering 19936 warnings in the test report. Some come from ruamel and some from NumPy 1.20, but most of them come from pyomo:

```
test_model.py: 19652 warnings
  /Users/trtim/Develop/euro-calliope/.snakemake/conda/8c2a862642b61c6eaabb146b1c24d816/lib/python3.8/site-packages/pyomo/solvers/plugins/solvers/direct_or_persistent_solver.py:219: DeprecationWarning:

  Deprecated, pass a TempConstr or use Model.addLConstr
```

brynpickering:
The warning you're coming across is being passed on by Pyomo, since it is related to their interface with Gurobi... Maybe an older `gurobipy` version would deal with this?

brynpickering (commented):
`0.6.6.post1` is the only version of `0.6.6` available on conda-forge (the need for the `post1` was due to an error in publishing the pip wheel, rather than any change in Calliope). I would say there's less confusion by going for `0.6.6`.

If this PR waits 1 week, we could also just leapfrog to v0.6.7...

timtroendle:
I think I'd rather stick to `0.6.6.post1` as it's more specific. Also, I am not exactly sure where it's fine to abbreviate it with `0.6.6` and where not. In conda envs, it should be fine. But is it also fine in the `model.yaml`?

We can ignore this is if we wait for 0.6.7 of course. Will 0.6.7 have any of the breaking changes of 0.7.0 or is it mainly a drop-in replacement? Do you think 0.6.7 will solve the deprecation warning?

brynpickering:
It's mainly a drop-in replacement. The tests won't spit out the deprecation warning you mentioned because that warning is explicitly ignored in the `pytest.ini` of 0.6.7.

timtroendle:
Ok, let's wait for 0.6.7 in that case.

PR #153, opened 2021-07-27
Add config option to choose solar and wind potential scenario

timtroendle:
Adds config option to choose the solar and wind potential scenario.

A major motivation for this change is visibility. While it was always simple to switch scenarios, this change increases the visibility of the default choice (technical-potential).

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


config/schema.yaml
```diff
@@ -155,7 +155,15 @@ properties:
                         enum: ["2010", "2030", "2050"]
             biofuel-efficiency:
                 type: number
+                description: Combustion efficiency of the power plant.
```
brynpickering, line 158:
```suggestion
                description: Combustion efficiency of the `biofuel` technology, which produces electricity from biofuel assuming intermediate anaerobic digestion to biogas followed by combustion.
```

config/schema.yaml
```diff
@@ -155,7 +155,15 @@ properties:
...
+                minimum: 0
                 maximum: 1
+            wind-and-solar-potential-scenario:
+                type: string
+                description: Scenario defining the amount of surfaces eligible for solar and wind power from [@Trondle:2019].
+                enum:
```
brynpickering, line 164: 1. Are these the only two available on Zenodo?
2. I would prefer following the square bracketed convention for lists where possible

brynpickering (approved):

config/schema.yaml
timtroendle: 1. Yes.
2. Ok, will change. We are using both ways in the schema though.

PR #157, opened 2021-07-30
Feature update calliope

### LINKED ISSUE #73: Update to Calliope 0.6.7

timtroendle:
Fixes #73 .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


timtroendle:
Now, there are even more warnings during tests than in #150. 🙈 To be fair, that's very likely only because there are more tests.

As the warning are DeprecationWarnings and as we fixed the versions, that's fine.

I did some research what this is. It's because [the latest Pyomo 6.0.1 calls the deprecated Gurobi method Model.addConstr](https://github.com/Pyomo/pyomo/blob/9172765e3a47c986d12f4390ba4aa457266b3966/pyomo/solvers/plugins/solvers/gurobi_direct.py#L321). Fun fact: [Gurobi's documentation doesn't say that the method is deprecated](https://www.gurobi.com/documentation/9.1/refman/py_model_addconstr.html). What the documentation _does_ say though is that replacing `Model.addConstr` with the recommended `Model.addLConstr` can [increase performance by up to 50%](https://www.gurobi.com/documentation/9.1/refman/py_model_addlconstr.html). So can we hope for some performance increases when pyomo is updated?

timtroendle:
I should check whether all these warnings end up on the Calliope log when running Calliope. If yes, we cannot leave it this way.

CHANGELOG.md
```diff
@@ -18,19 +18,20 @@
 * **ADD** option to shed load at all locations (#131).
 * **ADD** option to choose solar and wind potential scenario, limiting the amount of eligible surfaces (#153).
 
+* **UPDATE** Calliope version from 0.6.5 to [0.6.7](https://calliope.readthedocs.io/en/stable/history.html#id1) (#73).
```
brynpickering, line 21: `#id1` moves to point to the top header with each update to `stable`, so would be out of date by the next calliope release.
```suggestion
* **UPDATE** Calliope version from 0.6.5 to [0.6.7](https://calliope.readthedocs.io/en/v0.6.7/history.html#id1) (#73).
```

envs/test.yaml
```diff
@@ -3,12 +3,13 @@ channels:
...
+    - python=3.8
+    - numpy=1.19.5
+    - pandas=1.2.4
+    - xarray=0.17.0
+    - gurobi=9.1.2
+    - pytest=6.2.4
+    - pytest-html=3.1.1
+    - calliope=0.6.7
+    - pyomo=6.0.1
+    - pyutilib=6.0.0
```
brynpickering, line 15: pyomo no longer depends on pyutilib and, from v0.6.7, neither does Calliope - it can now be purged!

templates/environment.yaml
```diff
@@ -3,8 +3,11 @@ channels:
...
-    - pyomo=5.6.8 # to avoid https://github.com/calliope-project/calliope/issues/309
-    - pyutilib=5.8.0 # older version that works with pyomo 5.6.8
+    - python=3.8
+    - numpy=1.19.5
+    - pandas=1.2.4
+    - xarray=0.17.0
+    - gurobi=9.1.2
+    - calliope=0.6.7
+    - pyomo=6.0.1
+    - pyutilib=6.0.0
```
brynpickering, line 13: see above comment

Snakefile
```diff
@@ -357,7 +358,7 @@ rule model:
             technology=ALL_WIND_AND_SOLAR_TECHNOLOGIES
         ),
         rules.build_metadata.output,
-        lambda wildcards: "build/model/national/entsoe-tyndp-links.yaml" if wildcards.resolution == "national" else [],
+        lambda wildcards: rules.entsoe_tyndp_links.output[0] if wildcards.resolution == "national" else [],
```
brynpickering, line 361: I'm glad you snuck this in here, means I don't have such a public shaming for getting it wrong in my PR ;)

brynpickering (approved):
Just some minor changes then all good to go. Removing pyutilib should be fine, but worth doing a quick check on that in case there's something buggy somewhere....

Snakefile
timtroendle: you are welcome :)

brynpickering (approved):

Issue #159, opened 2021-08-02
Add a FAQ // known issues section


timtroendle:
While this FAQ section can and should be extended and updated whenever appropriate, here's a list of FAQs to start with:

- [x] Snakemake appears frozen while running on the cluster. What can I do? (related to unresolved #93)
- [x] The workflow fails when trying to create a conda environment. How can I resolve this?
- [x] How do you link rules to files?
- [x] How to debug?
- [x] how long should I expect components of the workflow to take to run?
- [x] how do I propagate changes made in the configuration file to update the build?
- [x] Can you run the workflow for several different weather years?

timtroendle:
@brynpickering , @FLomb , @wwwuFei , @FebinKa , @sjpfenninger , @suvayu , @jwohland: Is there anything else that would have helped you?

suvayu:
All my confusions have been about:
- visibility of internal state, while debugging, or
- trying to run a subset of the rules.

Possibly useful FAQs could have been:
- the conditions under which (re)running a subset of the rules is possible;
- for debugging, I resorted to print statements, however it wasn't obvious where to put them, or if it was possible at all for certain situations (e.g. debugging "snakemake globs"), an entry addressing that would help.

timtroendle:
Thanks @suvayu . I've added the debugging question. It came up before and it's certainly an important one.

For the rerunning question, let me ask so that I understand correctly. Do you mean "How to run parts of the workflow only, without executing the entire workflow?"

brynpickering:
Who is this FAQ aimed at? I could imagine ones associated with development (e.g. working with different config files when working on features), but this looks more like it's aimed at running the workflow as-is... 

Another, based on internal conversations: 

* how long should I expect components of the workflow to take to run?

suvayu:
yes, either because I am only rerunning part of the workflow because I changed something (say the year), or because I'm debugging a part of the workflow.

brynpickering:
RE the changing of the year, I think we should address this with the question: "why are parts of the workflow not automatically re-running when I update the configuration file?" since it is really the only part of the whole process that isn't intuitive to the way snakemake is supposed to work... We should then either have a list of rules that are affected by each config option (possibly too hopeful) or recommend people rerun everything to be safe and wait on your corresponding snakemake PR to fix this...

FLomb:
Perhaps this:
- Is there a sensible way to parallelise the workflow when, e.g., one needs to run it for several different weather years?

Issue #108, opened 2021-05-05
Add DAG to doc


sjpfenninger:
Provide a complete overview of repository layout

* All top-level directories and what their purpose is
* Figure(s) of DAG(s)

timtroendle:
The first of the two has always been part of the documentation. Or do you think it requires more work?

timtroendle:
We want this to be automated. First, so that we always generate the same graph. Second, so that we don't forget to update it. This will get easily out of sync if we do not automate it. In the best case, we would not need to check the result in to the Git repository (similar to the built documentation or the built model which are not checked in).

Automating this locally isn't too difficult. I cannot come up with a solution that automates this for Read the Docs. Somehow we needed a hook that generates the DAG just before readthedocs generates the documentation. I fear this hook does not exist. Any ideas? @brynpickering maybe?

Here's the prototype for a solution that works locally. Because Snakemake does not support this natively, it's a bit clunky.

```Python
rule graph_text:
    output: "build/some-file.txt"
    run:
        import io
        from contextlib import redirect_stdout

        graph = io.StringIO()
        with redirect_stdout(graph):
            snakemake.snakemake("/Users/trtim/Develop/euro-calliope/Snakefile", printrulegraph=True)
        with open(output[0], "w") as f_graph:
            f_graph.write(graph.getvalue())


rule graph:
    input: rules.graph_text.output[0]
    conda: "envs/docs.yaml"
    output: "build/some-file.png"
    shell: "cat {input} | dot -Tpng > {output}"
```

timtroendle:
A second best solution would be to keep the rules above so that we can automatically generate it locally. We check the graph in to the Git repository. Then, this must be updated manually as part of our release process (#109).

timtroendle:
The right entry point seems to be the plugin system of mkdocs: https://www.mkdocs.org/dev-guide/plugins/

PR #162, opened 2021-08-04
Add and split documentation into basic and advanced use

### LINKED ISSUE #154: Split the documentation based on user roles
The current documentation is all over the place -- and the new readthedocs documentation (introduced in #106) made that only worse than better I believe (because it exposes a lot of information targeted at developers and advanced users to the public which mainly want to use the pre-builts and don't need this information).

I believe there are three user roles and we should split the documentation based on these roles.

There are **basic users** (no, some, or expert knowledge in Calliope; no or little knowledge in running workflows; some or expert knowledge in energy system modelling). These users want to use the pre-built and they must know about model components, available overrides, and available modules they can import (yaml files).

There are **advanced users** (some or expert knowledge in Calliope; some or expert knowledge in running workflows; some or expert knowledge in energy system modelling). These users want to run the workflow and they must know (in addition) about repo layout, how to run the workflow, how to configure the workflow. They may be interested in using euro-calliope as a subworkflow (and thus a Cookiecutter template https://github.com/timtroendle/cookiecutter-reproducible-research/issues/4).

There are **developers** (expert knowledge in Calliope; some or expert knowledge in workflows; some or expert knowledge in Python development; no, some, or expert knowledge in energy system modelling). These _users_ will send pull requests and they must know about our branching model, testing strategy, and documentation policy.

I may have forgotten things, but this is roughly how I think we should structure the documentation -- in particular the one that is facing the public right now on read the docs.

timtroendle:
* Fix #154
* Split documentation into basic and advanced use (pre-builts vs workflow)
* Move all documentation to doc folder
* Restructure documentation
* Document all customisation options
* Move from Sphinx to mkdocs

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


docs/index.md
```diff
@@ -0,0 +1,17 @@
+# Euro Calliope
+
+Models of the European electricity system for Calliope.
```
brynpickering, line 3: `for the Calliope energy system modelling framework`?

docs/index.md
```diff
@@ -0,0 +1,17 @@
...
+
+Euro Calliope is a suite of pre-built models and routines to build these models from raw data. You can configure, adapt, and extend both models and routines in many ways to build your own model that is fit for your purpose.
+
+![](./img/ec-in-modelling-process.png)
+
+## At a glance
```
brynpickering, line 9: Figures here would be nice.

docs/sources.md
```diff
@@ -1,47 +1,47 @@
...
 
 JRC. (2014). <i>ETRI 2014 -- Energy Technology Reference Indicator projections for 2010-2050</i>. JRC. https://setis.ec.europa.eu/related-jrc-activities/jrc-setis-reports/etri-2014
 
-> [@EuropeanEnvironmentAgency:2009]
+### @EuropeanEnvironmentAgency:2009
 
 European Environment Agency. (2009). <i>Europe’s onshore and offshore wind energy potential</i> [Publication]. https://www.eea.europa.eu/publications/europes-onshore-and-offshore-wind-energy-potential
 
-> [@Ruiz Catello et al:2015]
+### @Ruiz Catello et al:2015
```
brynpickering, line 37: should just be Ruiz?

templates/README.md
```diff
@@ -2,114 +2,35 @@
 
 Ready to use models of the European electricity system built using Calliope. Models are available on three different spatial resolutions: continental, national, and regional.
 
-In addition, euro-calliope models can be built manually which adds more configuration options. To build euro-calliope manually, head over to [GitHub](https://github.com/calliope-project/euro-calliope).
+In addition, euro-calliope models can be built manually which adds more configuration options. To learn how to build euro-calliope manually, head over to [euro-calliope's documentation](https://euro-calliope.readthedocs.io).
```
brynpickering, line 5: I would keep the GH link and mention that it was built from raw data using this workflow, with a pointer to the repo with appropriate tag.

templates/README.md
```diff
@@ -2,114 +2,35 @@
...
+euro-calliope models the European electricity system with each location representing an administrative unit. It is built on three spatial resolutions: on the continental level as a single location, on the national level with 34 locations, and on the regional level with 497 locations. At each location, renewable generation capacities (wind, solar, bioenergy) and balancing capacities (battery, hydrogen) can be built. In addition, hydro electricity and pumped hydro storage capacities can be built up to the extent to which they exist today. All capacities are used to satisfy electricity demand on all locations where demand is based on historic data. Locations are connected through transmission lines of either unrestricted capacity or projections. Using [Calliope](https://www.callio.pe), the model is formulated as a linear optimisation problem with total monetary cost of all capacities as the minimisation objective. Due to the flexibility of Calliope and the availability of the routines building the model all components can be adapted to the user's needs.
```
brynpickering, line 9: This is the third instance of this paragraph, which might be a pain to deal with in the long run. Any way to depend on one instance of this description to fill in the rest...?

templates/README.md
```diff
@@ -2,114 +2,35 @@
...
-The units of quantities within the models are the following:
-
-* power: {{ (1 / scaling_factors.power) | unit("MW", parenthesis=False) }}
-* energy: {{ (1 / scaling_factors.power) | unit("MWh", parenthesis=False) }}
-* area: {{ (1 / scaling_factors.area) | unit("km2", parenthesis=False) }}
-* monetary cost: {{ (1 / scaling_factors.monetary) | unit("EUR", parenthesis=False) }}
+## More information
 
-These units were chosen in order to minimise numerical issues within the optimisation algorithm.
+For more information on euro-calliope and how to use and modify the models, see [euro-calliope's](https://euro-calliope.readthedocs.io) documentation.
```
brynpickering, line 33: I would have the hyperlink also covering 'documentation'

brynpickering (commented):
Nice rapid implementation.

I'm not keen on the structure though. `basic` and `advanced` are unnecessarily abstract. I would stick with more explicit sections like `Running the pre-built model`, `Customising the pre-built model`, `Building the model from raw data`, `Customising the model building workflow`. `Customising` could be a subset of the `running` and `building` sections, e.g.
```
Running the pre-built model
-- Prepare
---- The models
-- Run
-- Customise
Building the model from raw data
-- Prepare
---- The repository
-- Build
---- Building on a cluster
-- Run (maybe refer to the run phase from the other section)
-- Customise
```

Other general comments:
1. We need more pictures! We could have the network and region image in the index when describing what euro-calliope is and should have the rule graph in the `building the model` section.

2. We should have a `troubleshooting` section for running the euro-calliope model and building it. I guess this fits with the FAQs in #159. Ideally we want to use the documentation not as just a quick-start, but as the first line of defence for being inundated with support requests! This could come in on another PR. 

3. For general maintainability via GH PRs, it would be good to have one sentence per line. Then changes made to the documentation in future can be tracked a lot more easily. Provided there isn't an empty line, it will be rendered as one paragraph in the end.

templates/README.md
timtroendle: Yeah. But from there users would be pointed back to the documentation anyway. Plus, it's one more link for which we must maintain the correct version. I am already worried that we will mess this up.

templates/README.md
timtroendle: True. I tried to keep text duplicates to a minimum but found this paragraph important enough to copy it. If you have a technical solution, I am happy to hear about it. Otherwise we simply need to maintain the copies.

timtroendle:
I have addressed most of your comments.

The ones I haven't addressed are parts of other issues: network graph (#165), DAG visualisation (#108), troubleshooting (#159).

mkdocs.yaml
```diff
@@ -0,0 +1,23 @@
+site_name: Euro-Calliope
+nav:
+    - Home: index.md
+    - Access:
+        - Use pre-builts: model/pre-built.md
```
brynpickering, line 5:
```suggestion
        - Use pre-built models: model/pre-built.md
```

brynpickering (approved):
One minor change request, otherwise looks great!

Issue #158, opened 2021-08-02
Does `eurocalliopelib` have a public API?


timtroendle:
A while ago, we introduced `eurocalliopelib`(in #58) as a second place for Python code next to the `scripts` (formerly: `src`) directory. There is a single reason for that: We wanted to make code reusable across euro-calliope scripts so that we can reduce code duplication. (We also wanted to make code easier to test but with #120 this second goal does not require `eurocalliopelib` anymore.)

An important question is whether we see `eurocalliopelib` as an internal thing, or whether we want to make it public. Should it be used within euro-calliope scripts only, or also beyond that? I personally started using it in workflows that use euro-calliope as a subworkflow. Amongst others, this is important for API stability (see also [this comment](https://github.com/calliope-project/euro-calliope/pull/113#commitcomment-53666876)).

If we consider `eurocalliopelib` to be facing the public we should:

- [ ] Add it to the documention.
- [ ] Include it to the CHANGELOG.md.
- [ ] Include an appropriate versioning scheme and update the version.

If we consider `eurocalliopelib` to be internal only we should:

- [ ] Include an appropriate versioning scheme and update the version (likely to follow the version of the entire repository).

timtroendle:
@brynpickering, @suvayu, any thoughts?

brynpickering:
Hmm, I see `eurocalliopelib` being a pain to maintain alongside the main workflow since it is a versioned package within a separately versioned repository... 

I also use it in scripts that rely on euro-calliope as a sub-workflow so do see the advantage of a public API. Adding it to the documentation (for parts `3.` and `4.` I mention in #154) would make a lot of sense and having a sub-section in the CHANGELOG for it also sounds good.

RE versioning, following the version of the entire repository would be my preference. We could even have the repo version in a top-level file that `eurocalliope` can read so we don't need to remember to change this in several places when it comes to it.

timtroendle:
I see your points, but I don't understand how they match. I also think maintaining `eurocalliopelib` can be difficult, but wouldn't that mean we should rather not make it public?

Using the same version for `eurocalliopelib` and euro-calliope AND making `eurocalliopelib` public would either mean versioning will be tricky, or we will need to forget the whole idea of semantic versioning -- which I'd rather not.

We can always keep `eurocalliopelib` internal for now and publish it at a later stage if we want to. We can reuse it in subworkflows either way.

brynpickering:
[...]

PR #168, opened 2021-08-05
Version update to 1.2.dev

brynpickering:
Based on discussion in #158, I've preliminarily updated eurocalliopelib to be the same version as the main repo version. I've updated both of these to reflect our release plan for 1.2.

This PR replaces #167 (now closed) with the correct version number. 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


timtroendle (approved):

timtroendle:
I think I'd still go for 1.2.0 internally (i.e. with the micro version set to `0`), but that doesn't matter much here.

PR #170, opened 2021-08-06
Generalise unit conversion function

brynpickering:
In adding new features, I've come across the need to convert between units of energy on multiple occasions. Before the 'from' and 'to' units had to be explicitly defined, as well as the 'unit' index level added/updated manually. This rolls it all into a tested function.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/lib/eurocalliopelib/utils.py
+++ b/lib/eurocalliopelib/utils.py
@@ -79,13 +79,15 @@ def convert_unit(df, output_unit, input_unit=None, unit_in_output_idx=True):
             raise ValueError("Cannot infer unit for data")
         else:
             mask = df.index
-    df.loc[mask] *= UNIT_CONVERSION_MAPPING[(input_unit.lower(), output_unit.lower())]
+
+    if input_unit.lower() != output_unit.lower():
+        df.loc[mask] *= UNIT_CONVERSION_MAPPING[(input_unit.lower(), output_unit.lower())]
 
     if units is None and unit_in_output_idx is True:
-        df = add_idx_level(df, unit=output_unit)
+        df = add_idx_level(df, unit=output_unit.lower())
     elif units is not None:
         if unit_in_output_idx is True:
-            idx_renamer = lambda x: output_unit if x == input_unit else x
+            idx_renamer = lambda x: output_unit.lower() if x == input_unit else x
             df = df.rename(idx_renamer, level="unit")
         else:
             assert len(units) == 1, f"Cannot drop the index level `unit` with multiple units {units}"
--- a/lib/eurocalliopelib/utils.py
+++ b/lib/eurocalliopelib/utils.py
@@ -66,6 +66,29 @@ def to_numeric(series):
 
 
 def convert_unit(df, output_unit, input_unit=None, unit_in_output_idx=True):
+    """
+    Convert between units in a dataframe, e.g. from "TJ" to "TWh" in energy data
+
+    Parameters
+    ----------
+    df : pd.DataFrame or pd.Series with single-/multi-index
+    output_unit : str
+        Unit to which data must be converted.
+    input_unit : str, optional, default None
+        Unit in which data is currently stored.
+        If not given, will be inferred from the "unit" index level in `df`.
+    unit_in_output_idx : bool, default True
+        If True, include an index level "unit" with `output_unit` as the value.
+        If "unit" index level already exists, will rename the index level values
+        corresponding to `input_unit` to match `output_unit`.
+
+    Returns
+    -------
+    df : pd.DataFrame or pd.Series with single-/multi-index
+        If `unit_in_output_idx` is True, `df` will have an index level `unit` with value of
+        `output_unit` ALWAYS in lower case.
+        If `unit_in_output_idx` is False, there will be no index level `unit` in `df`.
+    """
     df = df.copy()
     if isinstance(df.index, pd.MultiIndex) and "unit" in df.index.names:
         units = df.index.get_level_values("unit").unique()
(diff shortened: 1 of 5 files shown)


tests/lib/test_utils.py
```diff
@@ -139,4 +139,124 @@ def test_remove_non_numeric_characters(self, raw, expected):
...
+    @pytest.fixture
+    def create_df(self):
+        def _create_df(input_units):
+            index = pd.MultiIndex.from_product([["foo", "bar"], input_units], names=["baz", "unit"])
+            return pd.Series(index=index, data=1)
+        return _create_df
+
+    @pytest.fixture
+    def create_idx(self):
+        def _create_idx(multi_or_single):
```
timtroendle, line 155: Looking at your tests, you don't need the complicated factory pattern here. Do this instead:

```Python
@pytest.fixture(parameters=["single", "multi"])
def idx(self, request):
    multi_or_single = request.param
    if multi_or_single == "multi":
        return pd.MultiIndex.from_tuples([("foo", "bar")], names=["baz", "foobar"])
    elif multi_or_single == "single":
        return pd.Index(["foo"], name="baz")
```

All tests asking for `idx` will be run twice automatically. Syntax in my example is likely not 100% correct.

tests/lib/test_utils.py
```diff
@@ -139,4 +139,124 @@ def test_remove_non_numeric_characters(self, raw, expected):
...
+            if multi_or_single == "multi":
+                return pd.MultiIndex.from_tuples([("foo", "bar")], names=["baz", "foobar"])
+            elif multi_or_single == "single":
+                return pd.Index(["foo"], name="baz")
+        return _create_idx
+
+    def get_conversions():
```
timtroendle, line 162: This is cheating! You are testing the implementation by comparing it to an important part of the implementation. You are still testing something, but certainly not everything and especially not the important bit: the numbers! I think there should be explicit numbers in this test.

timtroendle: In addition, your tests will be much easier to understand. Currently there are all these variables `convert_from` and `convert_to` which are difficult to graps. Another reason why they should be explicit.

lib/eurocalliopelib/utils.py
```diff
@@ -51,18 +60,63 @@ def to_numeric(series):
...
+        If True, include an index level "unit" with `output_unit` as the value.
+        If "unit" index level already exists, will rename the index level values
+        corresponding to `input_unit` to match `output_unit`.
+
+    Returns
+    -------
+    df : pd.DataFrame or pd.Series with single-/multi-index
+        If `unit_in_output_idx` is True, `df` will have an index level `unit` with value of
+        `output_unit` ALWAYS in lower case.
+        If `unit_in_output_idx` is False, there will be no index level `unit` in `df`.
```
timtroendle, line 90: But it will exist if it existed before?

lib/eurocalliopelib/utils.py
```diff
@@ -51,18 +60,63 @@ def to_numeric(series):
...
 
-def ktoe_to_twh(array):
-    """Convert KTOE to TWH"""
-    return array * 1.163e-2
+    Parameters
+    ----------
+    df : pd.DataFrame or pd.Series with single-/multi-index
+    output_unit : str
+        Unit to which data must be converted.
+    input_unit : str, optional, default None
```
timtroendle, line 77: Is this unit inference on a row by row basis? Otherwise I would not do unit inference -- it's just too brittle and complicated for its usefulness.

lib/eurocalliopelib/utils.py
```diff
@@ -51,18 +60,63 @@ def to_numeric(series):
...
+    df : pd.DataFrame or pd.Series with single-/multi-index
+        If `unit_in_output_idx` is True, `df` will have an index level `unit` with value of
+        `output_unit` ALWAYS in lower case.
+        If `unit_in_output_idx` is False, there will be no index level `unit` in `df`.
+    """
+    df = df.copy()
+    if isinstance(df.index, pd.MultiIndex) and "unit" in df.index.names:
+        units = df.index.get_level_values("unit").unique()
+        if input_unit is None:
+            assert len(units) == 1, f"Cannot infer unit for data with multiple available units {units}"
```
timtroendle, line 98: Ok so this answers my question from above. Do you think this unit inference is really worth its potential problems?

lib/eurocalliopelib/utils.py
```diff
@@ -51,18 +60,64 @@ def to_numeric(series):
...
 
-def gwh_to_tj(array):
-    """Convert GWh to TJ"""
-    return array * 3.6
 
+def convert_unit(df, output_unit, input_unit=None, unit_in_output_idx=True):
+    """
+    Convert between units in a dataframe, e.g. from "TJ" to "TWh" in energy data
 
-def ktoe_to_twh(array):
```
timtroendle, line 63: I actually think this was a very clear function. 🙈 I guess you have good reasons to make this more complex.

Here's a compromise solution to consider:

First, I wouldn't do unit inference. It's just so much more brittle and implicit and its added value is close to zero, isn't it? That's why I'd consider getting rid of unit inference. If you do, you could also consider to keep these functions as convenience functions like so:

```Python
def convert_unit(df, output_unit, input_unit, unit_in_output_idx=True):
   # as before but input_unit is mandatory
   return

def gwh_to_tj(df, **kwargs):
     return convert_unit(df, output_unit="twh", input_unit="ktoe", **kwargs)

def ...
```

timtroendle (changes requested):
Generally fine. But I am not a big fan of the unit inference -- see my comments. I would only keep it if absolutely necessary.

I think you need to work on the tests. It was difficult for me to make sense out of them. I see the rationale behind your test design: they are brief, smart, and there are LOADS of them.

But they are also very implicit and that makes them difficult to understand. There should be at least a few tests with explicit input and output -- those are usually the easiest to understand the purpose of a function.

You also don't need so many tests. For example, many tests are testing each unit conversion in UNIT_CONVERSION_MAPPING. That's not necessary. Test them once, and then you know they work, right? I mean if you have tested KTOE -> KWH once, there is no reason to believe it doesn't work in another test.

lib/eurocalliopelib/utils.py
brynpickering: No, not sure the best way to name this to cover both. `keep_unit_index_level`?

lib/eurocalliopelib/utils.py
brynpickering: I think yes, if (as I have been doing) there is almost always an index level 'unit' in datasets. It's a lot *less* brittle than variable/column names being appended with the unit name...

brynpickering:
@timtroendle I can see the concern RE the inference. The reason I chose this approach was because I was finding so many instance of doing e.g. 

```
df = df.xs("TJ", level="unit").apply(tj_to_twh)
df = utils.add_idx_level(df, unit="twh")
```

This is good, as it means the data always has the unit information attached to it. However, I believe it's more error-prone and likely to go wrong with future iterations. 

I can see the issue of the current implementation though and have come to dislike my approach over the last few hours ;) 

I wonder whether an alternative would be to store all this data as xarrays with a "unit" attribute. This would then be a lot easier to track, less error-prone to infer, and just as easy to update data...

timtroendle:
Regarding adding a unit attribute to processed data: of course, that makes sense. We don't have a consistent way for that right now. 

Generally though, I would suggest to stick to the default unit for all processed data: MW / MWh. If all data are in the same unit, units should probably still be documented, but can in practise be ignored (or assumed) when working with the data.

timtroendle:
Closing this as it became stale and there is no conclusion.

Issue #109, opened 2021-05-05
Developer docs


sjpfenninger:
Add more info on how to contribute

* development workflow
* testing
* git branch model

timtroendle:
Importantly, this should include a release process. There are actually more and more elements to this:

* changelog
* version of euro-calliope
* version of eurocalliopelib
* version of euro-calliope-datasets
* ...

timtroendle:
Also needed: how to build the documentation.

timtroendle:
Also: remove the changelog to the top level and copy it into docs using a mkdocs plugin.

PR #171, opened 2021-08-06
Add DAG to doc

### LINKED ISSUE #108: Add DAG to doc

timtroendle:
Fix #108.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 build
 data
+docs/img/dag.png
 
 notebooks
 
--- a/docs/workflow/overview.md
+++ b/docs/workflow/overview.md
@@ -29,6 +29,20 @@ This way, everyone can ensure that the workflow does indeed build the models it 
 
 Automation makes the models **adaptable**, as one can not only reproduce the exact same model but one can also [customise and configure](./customisation.md) the workflow so that it builds a different variant of the model that fits better for ones purpose.
 
+## Workflow steps
+
+The image below shows all steps in Euro-Calliope's workflow and their connections.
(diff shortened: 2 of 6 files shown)


timtroendle:
The plugin system of mkdocs is great: simple and powerful!

The disadvantage of this approach is that building the documentation becomes more complicated. The environment is just much heavier than before.

Also, the image is too small. Actually, the column of the theme is too small to handle the image. @brynpickering any ideas? Would be nice to be able to click to enlarge the image, but I'm not sure how to do that. Otherwise could also see whether the rendering of the image could be constrained in width, but that would likely make the visualisation more difficult to read.

brynpickering:
> Also, the image is too small. Actually, the column of the theme is too small to handle the image. @brynpickering any ideas? 

No great ideas, sadly. You could make the image clickable pointing to itself? `[](img.jpg)` 
We could also change theme to something that uses more width.

timtroendle:
@brynpickering have a look at this implementation of a Zoom. What do you think?

timtroendle:
@brynpickering sorry for all the noise. This works and is eventually ready for review.

brynpickering (approved):
Ultimately quite complex... Hopefully we can use some of this functionality for more than just this one graphic....

timtroendle:
You think so? Do you find the plugin mechanism complex? I thought it's rather elegant. Either way, we can certainly reuse this. First in #177 and then likely again in #109.

PR #172, opened 2021-08-09
Feature validate external links

### LINKED ISSUE #164: Validate external links in documentation
We should have a test that ensures that external links in all files in `docs` (and maybe the two readmes) actually work.

timtroendle:
Fixes #164  .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.github/workflows/externallinks.yaml
+++ b/.github/workflows/externallinks.yaml
@@ -1,0 +1,14 @@
+name: Check Markdown links
+
+on:
+    push:
+    pull_request:
+    schedule:
+        - cron: "0 3 * * *" # Runs every day at 3am.
+
+jobs:
+  markdown-link-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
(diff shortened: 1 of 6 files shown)


timtroendle:
This caught 7 broken links. Seems very valuable.

.github/workflows/externallinks.yaml
```diff
@@ -0,0 +1,14 @@
+name: Check Markdown links
+
+on:
+    push:
+    pull_request:
+    schedule:
+        - cron: "0 3 * * MON" # Runs Mondays at 3am.
```
brynpickering, line 7: How would we be actually notified of link issues when running weekly?

brynpickering (approved):
Looks good!

.github/workflows/externallinks.yaml
timtroendle: By email. I guess it's send to all admins or whatever the right role is. For my fork, I received the email this afternoon.

PR #177, opened 2021-08-11
Feature config in doc

### LINKED ISSUE #68: Parse schema and default config to a table and include in documentation
The schema includes long descriptions for each config parameter, which is possibly useful for us core developers when trying to update things, but is invisible to most users. To overcome this, the schame should really be parsed to a nice table and included in the README or some more in-depth documentation. Default values for config items should also be in there, which is currently found only in the default config file (but could move to the schema, and be printed into a config file on validation).

timtroendle:
Fixes #68.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -283,16 +283,15 @@ properties:
                         description: Move stations outside of hydrobasins into closest basin by max this amount (fails otherwise).
             load:
                 type: object
-                description: >
+                description: >-
                     Data quality enforcing methods for open-power-system-data.org-derived national load data. These cover, in order:
-                        1. remove outlier data hours, based on full dataset
-                        2. select only data for year of interest
-                        3. interpolate empty data when the data gap is small
-                        4. fill larger data gaps with data from nearby years
-                        5. If 29th of February is empty and couldn't previously be filled, use data from the 28th of February
-                        6. select most complete dataset for each country from list of data sources, based on a priority order
-
-                        This process will fail if the 'most complete dataset' for any country has any remaining gaps.
+                    1. remove outlier data hours, based on full dataset;
+                    2. select only data for year of interest;
+                    3. interpolate empty data when the data gap is small;
+                    4. fill larger data gaps with data from nearby years;
+                    5. if 29th of February is empty and couldn't previously be filled, use data from the 28th of February;
+                    6. select most complete dataset for each country from list of data sources, based on a priority order.
+                    This process will fail if the 'most complete dataset' for any country has any remaining gaps.
                 properties:
                     outlier-data-thresholds:
                         type: object
@@ -316,6 +315,7 @@ properties:
                         type: array
                         uniqueItems: true
                         minItems: 1
+                        description: Order in which to prioritise available electricity load data from open-power-system-data.org. The first source in the list will be used to initialise the data; the remaining data sources in the list will be use to gap fill in order from first to last.
                         items:
                             type: string
                             enum:
@@ -326,7 +326,7 @@ properties:
                                 - actual_gross_generation_tso
                                 - forecast_entsoe_transparency
                                 - forecast_tso
-                        description: Order in which to prioritise available electricity load data from open-power-system-data.org. The first source in the list will be used to initialise the data; the remaining data sources in the list will be use to gap fill in order from first to last.
+
     capacity_factors:
         type: object
         description: Parameters related to average or time-dependent capacity factors of renewables.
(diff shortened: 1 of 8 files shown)


timtroendle:
This is blocked by #171. Once merged, need to merge develop into this branch.

timtroendle:
The result is not pretty, but it's complete and the time invest was justified.

config/schema.yaml
```diff
@@ -255,17 +268,29 @@ properties:
...
-                        - interpolate empty data when the data gap is small
-                        - fill larger data gaps with data from nearby years
-                        - If 29th of February is empty and couldn't previously be filled, use data from the 28th of February
-                        - select most complete dataset for each country from list of data sources, based on a priority order
+                        1. remove outlier data hours, based on full dataset
+                        2. select only data for year of interest
+                        3. interpolate empty data when the data gap is small
+                        4. fill larger data gaps with data from nearby years
+                        5. If 29th of February is empty and couldn't previously be filled, use data from the 28th of February
+                        6. select most complete dataset for each country from list of data sources, based on a priority order
```
brynpickering, line 293: Currently renders with the numbering continuing through to the parameters in `load`...

brynpickering (changes requested):
Agreed that it is worthwhile to have, but this package might be more appropriate for a good result, and possibly less involved implementation? https://github.com/coveooss/json-schema-for-humans

You could output as html or as md. Running both locally on the schema works nicely. It comes down to our preferred aesthetics.

timtroendle:
Ok, will look into `json-schema-for-humans`.

timtroendle:
Hm do you have a suggested configuration? I am creating a Markdown and I find the result pretty bloated -- I don't understand everything in that file and I wouldn't say it provides an easy overview. There is loads of stuff that's too bold and just unnecessary. It basically looks like their own documentation, which I find bloated and difficult to navigate, too:

https://github.com/coveooss/json-schema-for-humans/blob/master/docs/examples/examples_md_default/Configuration.md#deprecated_from_description

If we find a configuration generating a better result it could still be a good tool for its ease of use.

brynpickering:
@timtroendle I've looked into this a bit more and agree with you. Your current implementation seems the best of the options out there. I have fixed the 'data_quality' section locally so that it doesn't need special treatment. Let me know if I can push it to your branch. Two other issues that I see:

1. schema items using 'propertyNames' don't render ('shapes' and 'sea-connections'), which is an issue with the dependency not supporting it.
2. it would be good to include default values in here too. Would this be easy to add to each line in 'lines' after turning it to markdown...?

timtroendle:
1. Sure, push. I actually did not want to change `schema.yaml` more than necessary for this, but maybe that's only my neurosis calling here.
2. I can give it a try, but wouldn't spend much more time on this. So if there is no quick solution, I suggest to omit default values and instead point the reader to the `default.yaml` file for default values. (Thinking about it: it may be quicker to "fake" default values in the schema by adding them from `default.yaml` and then let the third-party library render it to Markdown. Default values are a feature of schemas, we are just not using it.)

brynpickering:
1. Don't worry the change is pretty minimal. 
2. OK, maybe update the schema preamble to link directly to the default YAML in this repo?

brynpickering:
btw, markdown link check fails on https://open-power-system-data.org which it shouldn't....

lib/eurocalliopelib/docs/schema.py
```diff
@@ -50,12 +51,10 @@ def on_files(self, files, config, **kwargs):
...
-        lines = ['    '.join(line.split('  ')) for line in lines]
-        # 2. Change headline
+        # 1. Change headline
         assert lines[0] == '# JSON Schema\n\n'
         lines[0] = "# Configuration parameters of Euro-Calliope's workflow\n\n"
-        # 3. Remove main description and subheadline
+        # 2. Remove main description and subheadline
         assert lines[2] == '## Properties\n\n'
         del lines[2]
         #del lines[1]
```
brynpickering, line 60: delete this commented line before merging.

timtroendle:
Nice changes! The failing Markdown link check may simply be a hickup, let's see whether it fails again.

Re default values: I decided to go for the simple solution. While adding default values for simple parameters would be easy, it's complicated for more complicated ones like the arrays.

brynpickering (approved):
Looks good to me. The only issue is with `shapes` and `sea-connections` schema not rendering, but I think we can aim to deal with that at a later date via #180.

PR #181, opened 2021-08-18
Move from tidy dataframes to xarray dataarrays

brynpickering:
This is a change that I've deliberated over for some time, but I think is ultimately the better choice for processing the multi-dimensional data that occurs so frequently in the workflow. The main benefits of relying on xarray are:

1. We can select single/multiple items from a named dimension, which is only possible with a single item in pandas (`.xs()`).
2. `groupby` only needs the specific dimension one cares to group to be defined. In pandas, any multiindex levels not in the groupby will be aggregated out.
3. Similarly, `sum` sums over the dimension defined in xarray but over the dimensions not defined in pandas (e.g. `da.sum("foo")` would leave a dataarray without the dimension `foo`, `df.sum(level="foo")` would leave *only* the dimension `foo`). This means you don't always *have* to know all the dimensions your working with in an array.

And yet, pandas tidy dfs (a pandas `Series` with a `MultiIndex`) have got a few advantages that might be sorely missed:

1. They are very easy to concatenate and are dense (no NaNs)
2. It is easy to align two datasets that have only a subset of shared dimensions (maybe this is easy in xarray too...?)
3. It's easy to read them as text files and to share the data with others.

Anyway, I think the downsides are outweighed by the benefits!

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


lib/eurocalliopelib/utils.py
```diff
@@ -60,6 +81,21 @@ def gwh_to_tj(array):
     return array * 3.6
 
 
+def gwh_to_twh(array):
+    """Convert GWh to TWh"""
+    return array / 1000
+
+
+def pj_to_twh(array):
+    """Convert GWh to TWh"""
```
timtroendle, line 90: When you copy 18 characters to avoid repeating 7 of them. :P

lib/eurocalliopelib/utils.py
```diff
@@ -60,6 +81,21 @@ def gwh_to_tj(array):
...
+    return array / 3.6
+
+
+def tj_to_twh(array):
+    """Convert GWh to TWh"""
```
timtroendle, line 95: When you copy 18 characters to avoid repeating 7 of them. :P

lib/eurocalliopelib/utils.py
```diff
@@ -40,6 +41,26 @@ def convert_country_code(input_country, output="alpha3"):
         return pycountry.countries.lookup(input_country).alpha_3
 
 
+def rename_and_groupby(da, rename_dict, dim, new_group_dim_name=None, dropna=False):
```
timtroendle, line 44: I think all functions in the lib should have docstrings, even if short ones.

lib/eurocalliopelib/utils.py
```diff
@@ -40,6 +41,26 @@ def convert_country_code(input_country, output="alpha3"):
...
+        .reindex({dim: rename_da[dim]})
+        .groupby(rename_da)
+        .sum(new_group_dim_name, skipna=True, min_count=1, keep_attrs=True)
+    )
+    if dropna:
+        da = da.dropna(new_group_dim_name, how="all")
+    return da
+
+
+def merge_da(da_list):
```
timtroendle, line 59: Same here, should have a docstring in my opinion.

lib/eurocalliopelib/utils.py
```diff
@@ -40,6 +41,26 @@ def convert_country_code(input_country, output="alpha3"):
...
+    datasets = [da.to_dataset(name="var") for da in da_list]
```
timtroendle, line 60: No need to turn them into a dataset. They only need a name, which `rename("var")` will do.

lib/eurocalliopelib/utils.py
```diff
@@ -40,6 +41,26 @@ def convert_country_code(input_country, output="alpha3"):
...
+    return xr.merge(datasets, combine_attrs="no_conflicts")["var"]
```
timtroendle, line 61: The side-effect of this is that you will have a "var"-named dataarray as an output. Is that fine?

lib/eurocalliopelib/utils.py
timtroendle: The docstring must include what the types of these things are. For example, I don't get what `new_group_dim_name` is or what it does.

timtroendle: Isn't this just a `new_dim_name`? Maybe rename `dim` to `dim_name` to make this more obvious?

scripts/annual_energy_balance.py
```diff
@@ -258,22 +257,47 @@ def get_ch_industry_energy_balance(path_to_excel):
...
     )
-    df = (
-        df.apply(utils.to_numeric)
+    df_twh = (
+        df
         .set_index("Unnamed: 0", append=True)
+        .apply(utils.to_numeric)
+        .apply(utils.tj_to_twh)
         .rename_axis(index=["carrier_code", "year"], columns="cat_code")
+        # ASSUME: we take the 'new' 2013 values, rather than the 'old' ones.
```
timtroendle, line 266: Can you add a little more context so that this assumptions can be understood without looking at the code? Something along these lines: `# ASSUME: we take the 'new' 2013 values from the Swiss industry energy balance, rather than the 'old' ones.`

Even better: say WHY we do that.

scripts/annual_energy_balance.py
```diff
@@ -258,22 +257,47 @@ def get_ch_industry_energy_balance(path_to_excel):
...
+        .groupby(level=["carrier_code", "year"]).sum()
+    )
+
+    da = df_twh.stack().to_xarray()
+
+    return da.assign_attrs({"unit": "twh"})
+
 
-    return df.stack()
+def valid_countries_to_alpha3(country_codes):
```
timtroendle, line 292: Maybe move to lib.util?

timtroendle (changes requested):
Sorry for the delay. I cannot say that I understood every change in detail, but all changes look rather harmless.

A few comments mostly on documentation below.

PR #182, opened 2021-08-20
Update model to enable multi-year config

### LINKED ISSUE #152: Allow to run multiple years
Euro-Calliope is built to run a single year. While it is possible to run multiple years, this requires manual work. Instead, there should be a built-in mechanism that automates this.

brynpickering:
Fixes #152. 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


scripts/hydro/inflow_mwh.py
```diff
@@ -4,12 +4,14 @@
...
-def determine_energy_inflow(path_to_stations_with_water_inflow, path_to_generation, year,
-                            max_capacity_factor, path_to_output):
+def determine_energy_inflow(path_to_stations_with_water_inflow, path_to_generation, start_year,
+                            end_year, max_capacity_factor, path_to_output):
     plants_with_inflow_m3 = xr.open_dataset(path_to_stations_with_water_inflow)
-    annual_national_generation_mwh = read_generation(path_to_generation, year)
-
-    inflow_MWh = energy_inflow(plants_with_inflow_m3, annual_national_generation_mwh, max_capacity_factor)
+    inflows = []
+    for year in range(start_year, end_year + 1):
```
timtroendle, line 11:
```Python
inflows = [
                      energy_inflow(
                          plants_with_inflow_m3.sel(time=str(year)), 
                          read_generation(path_to_generation, year), 
                          max_capacity_factor)
                      ) for year in range(start_year, end_year + 1)
]
```

tests/model/test_model.py
```diff
@@ -17,6 +17,7 @@
     "frozen-hydro": DEFAULT_TECHNOLOGIES,
     "alternative-cost": DEFAULT_TECHNOLOGIES,
     "shed-load": DEFAULT_TECHNOLOGIES | set(["load_shedding"]),
+    "multi-year": DEFAULT_TECHNOLOGIES,
```
timtroendle, line 20: This doesn't scale. Disclaimer: I am the original author.

tests/resources/national/model.yaml
```diff
@@ -22,13 +22,16 @@ scenarios:
     frozen-hydro: ["freeze-hydro-capacities", "no-hydro-fixed-cost"]
     alternative-cost: ["dea-renewable-cost", "schroeder-hydro-cost"]
     shed-load: ["load-shedding"]
+    multi-year: ["two-years"]
```
timtroendle, line 25: The same scenario should exist for continental and regional resolutions.

timtroendle (changes requested):
Nice! A few detail comments below.

Do you need this in 1.1? Otherwise I suggest to release this with 1.2 and allow us more time to test it thoroughly ourselves. If we bring it into 1.2 we can either keep this on hold or merge it into the long-lived feature branch.

timtroendle:
Oh and: what about CHANGELOG and documentation?

rules/hydro.smk
```diff
@@ -31,13 +31,14 @@ rule download_runoff_data:
...
-        y_max = config["scope"]["bounds"]["y_max"]
+        first_year = config["scope"]["temporal"]["first-year"],
+        final_year = config["scope"]["temporal"]["final-year"],
+        x_min = config["scope"]["spatial"]["bounds"]["x_min"],
+        x_max = config["scope"]["spatial"]["bounds"]["x_max"],
+        y_min = config["scope"]["spatial"]["bounds"]["y_min"],
+        y_max = config["scope"]["spatial"]["bounds"]["y_max"]
     output:
-        protected("data/automatic/europe-cutout-{}.nc".format(config["year"]))
+        protected("data/automatic/europe-cutout.nc")
```
timtroendle, line 41: Would be better to include years like this: `"data/automatic/europe-cutout-{first-year}-{final-year}.nc"`

config/minimal.yaml
```diff
@@ -9,14 +9,18 @@ sea-connections:
...
+            - "Ireland"
+            - "United Kingdom"
+        bounds:
+            x_min: -10.62
+            x_max: 1.8
+            y_min: 49.8
+            y_max: 60.9
+    temporal:
+        first-year: 2015
+        final-year: 2016
```
timtroendle, line 23: Is it a good idea to run this for two years? Will this make the execution much longer?

tests/resources/continental/model.yaml
```diff
@@ -20,6 +20,10 @@ scenarios:
     frozen-hydro: ["freeze-hydro-capacities", "no-hydro-fixed-cost"]
     alternative-cost: ["dea-renewable-cost", "schroeder-hydro-cost"]
     shed-load: ["load-shedding"]
+    multi-year: ["two-years"]
+overrides:
+    two-years:
+        model.subset_time: ['2015-12-31', '2016-01-01']
```
timtroendle, line 26: Wait, this must fail in the normal workflow, because the `default` config includes one year only. Here and in the two other models.

timtroendle (changes requested):
Thanks for updating this. Seemed to be quite some effort.

There are a few tiny things to fix -- in particular, I'd expect the default workflow execution to fail.

rules/hydro.smk
brynpickering: Yes, I had been thinking this too. The issue would be that you don't want to have to re-download this data if you rerun with a set of different years that overlaps with a previous run (e.g. you previously ran with 2010-2015 and now you want to run with just 2014-2016), or is a subset of a previous run (e.g. 2011-2012). Perhaps a cutout per year would suffice, running this rule with the year as a wildcard. Then the relevant years can be merged in 'inflow_m3'?

config/minimal.yaml
brynpickering: I wouldn't say *much* longer. The optimisation part of the workflow run just goes from december 31st 2015 to january 1st 2016. So, it crosses the two years, but only runs for two days.

tests/resources/continental/model.yaml
brynpickering: Yes, I didn't think of the need to have these tests working for both the minimal and default runs (I was considering only the minimal workflow). Since the default workflow tests are only run for each release, it seems reasonable to have it run over more than one year. Is there anything wrong with the pre-built model having more than one year of data, other than size?

rules/hydro.smk
timtroendle: I personally would not bother with caring for the case of redownloads. I don't see this as an important use-case. A usual use-case may be to use a subset, but that's something you'd do in the Calliope model rather than in the workflow.

tests/resources/continental/model.yaml
timtroendle: Mainly size I'd say. And that it is slightly odd to have two years in a model that by default runs for one year. Don't you think that testing the two-year mechanism in the minimal workflow suffices?

brynpickering:
@timtroendle: Minimal workflow builds with the latest commit.

config/euler/cluster-config.yaml
```diff
@@ -8,3 +8,16 @@ test:
     runtime: 240
     cores: 1
     memory: 16000
+download_runoff_data:
```
timtroendle, line 11: This is configured as a localrule which won't be send to a compute node. You mentioned that this is problematic because of the memory footprint.

How did you manage to run this on a compute node? Shouldn't you remote the localrule statement? And, if that's necessary, can you document what to do to be able to reach the internet from a compute node?

timtroendle (approved):
One last comment.

Regarding runtime of the minimal workflow, that included downloads, right? With data available, it should take a few minutes only.

config/euler/cluster-config.yaml
brynpickering: It's no longer a localrule. I don't know if I do anything special to make it work on a compute node. I have no special environment modules loaded....

brynpickering:
RE runtime, it still takes a while even without downloads. Does it really only take a few minutes for the minimal workflow currently, following `snakemake --cores 1 clean`?

timtroendle:
Are you running locally or on Euler? I gave it a try locally just now and it ran slightly longer than 9 minutes, using 3 cores and running from develop.

brynpickering:
Locally, it takes 14 minutes with 3 cores on this branch. It takes longer on the cluster, probably due to the time penalty of sending things off to job nodes. Is that permissible?

timtroendle:
I'd say it's not great as 14 minutes is quite long for a test that is supposed to run often. But I guess there is not much we can do.

brynpickering:
Does this mean we need an intermediate testing level, which happens just before PR merging or release? In most cases, running over multiple years shouldn't be broken by features/fixes - it's just looping over the same functions, after all.

timtroendle:
I think I wouldn't bother at this point -- also to not make the testing infrastructure too complex (it is already rather complex I think). But we need to keep this in mind, as the runtime of the minimal workflow will only become longer with more features.

PR #192, opened 2021-10-21
Update ts weighting to handle NaN values

brynpickering:
Just stumbled on this issue when running the workflow: Italy had an offshore wind capacityfactor of zero. This was caused by a single gridcell in the CF timeseries having only NaN values. 

This commit drops those gridcells with possible non-zero/NaN weight but NaN values. It then scales weights to 1, since there is one gridcell with a non-zero weight that has been removed. 

On testing, the weight of the removed gridcell in offshore wind was ~1e-8, so the weight correction wasn't really necessary...

Another option to fix this would be to "skipna" in the summation over all gridcells. The current implementation doesn't skip them.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -79,8 +79,14 @@ def weighted_time_series(weights_and_values):
     ds = (
         weights_and_values
         .where(weights_and_values.weight > 0)
-        .dropna(subset=["weight"], dim="xy") # drop all locations with weight == 0
+        .dropna(subset=["weight", "value"], dim="xy") # drop all locations with weight == 0 or value == np.nan
     )
+    if ds.weight.sum() != 1:
+        print(
```
timtroendle, line 85: Instead of a print message, would it make sense to assert a configurable threshold (must be very high, like 0.99 or so) is not exceeded? These printed messages can get overlooked too easily.

timtroendle (changes requested):
Nice catch! This solves the regression I found in #202.

Can you add a changelog entry?

lib/eurocalliopelib/geo/spatiotemporal.py
timtroendle: I think this could be even as high as 0.999 (99.9%).

brynpickering: Yeah, that would be fine by me. Maybe 0.995?

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -79,8 +79,14 @@ def weighted_time_series(weights_and_values):
     ds = (
         weights_and_values
         .where(weights_and_values.weight > 0)
-        .dropna(subset=["weight"], dim="xy") # drop all locations with weight == 0
+        .dropna(subset=["weight", "value"], dim="xy") # drop all locations with weight == 0 or value == np.nan
```
timtroendle, line 82: Maybe add `how="any"` to the function call? This happens implicitly, but it seems important enough to make it explicit here.

lib/eurocalliopelib/geo/spatiotemporal.py
```diff
@@ -75,12 +77,19 @@ def weights_between_shape_and_xy(shapes, stacked_spatiotemporal):
...
 
 
-def weighted_time_series(weights_and_values):
-    ds = (
+def weighted_time_series(weights_and_values, gridcell_overlap_threshold):
+    ds = (  # drop all locations with weight == 0 or value == np.nan
         weights_and_values
         .where(weights_and_values.weight > 0)
-        .dropna(subset=["weight"], dim="xy") # drop all locations with weight == 0
+        .dropna(subset=["weight", "value"], dim="xy", how="all")
```
timtroendle, line 84: This must be "any", otherwise it doesn't work.

tests/lib/test_spatiotemporal.py
```diff
@@ -11,6 +11,7 @@
 WGS84 = "EPSG:4326"
 EPSG3035 = "EPSG:3035"
 EPSG27700 = "EPSG:27700"
+DEFAULT_THRESHOLD = 0.999
```
timtroendle, line 14: There should be a test. The "bug" above wasn't captured by this.

timtroendle (changes requested):
Looks almost good! I'll try to come up with a test myself.

timtroendle (approved):

timtroendle:
A changelog entry is not necessary as this is a regression introduced after the last release. I ran the minimal workflow with all tests and it works beautifully. Therefore, I am going to merge this PR.

PR #196, opened 2021-12-08
Add automatic download of bioenergy potentials and cost

### LINKED ISSUE #193: Cost of municipal solid waste is too high
The cost of municipal solid waste is too high by factor 1000. In the source, cost values of all biomass feedstocks are given in €/GWh -- all but municipal solid waste, where they are given in €/MWh.

As euro-calliope forms a weighted average over all countries and feedstocks, the impact is less severe but it is still large: Instead of ~28 € / MWh the cost should be 17 € / MWh. That's 40% lower than the 1.0.0 value.

Variable cost of biofuel is also impacted by its efficiency (default 45%) and O&M cost (3.1 € / MWh within ETRI). The old value this was 64.83 €/MWh, the new value is 40.88. That's 37% lower than the 1.0.0 value.

In addition to fixing this issue, we should align the years of cost and potential #195. The value then will be 18.47 €/MWh, leading to a total variable cost of 44.14 €/MWh. That's 32% below the 1.0.0 value.

### Version

1.0.0

### LINKED ISSUE #194: Add automatic download of bioenergy potentials and costs
Euro-Calliope uses bioenergy potential and cost data from @RuizCastello:2015. As that data were published only in PDF form, I extracted the data manually from the PDF and added it to the repo (now separated in the dataset repo).

In the meantime, JRC published ENSPRESO including a downloadable dataset as spreadsheet in which they include those data. We should replace the manually extracted data with the new publication.

Ruiz Castello, P., Sgobbi, A., Nijs, W., Thiel, C., Dalla Longa, F., Kober, T., Elbersen, B., &#38; Hengeveld, G. (2015). <i>The JRC-EU-TIMES model. Bioenergy potentials for EU and neighbouring countries.</i> [https://ec.europa.eu/jrc/en/publication/eur-scientific-and-technical-research-reports/jrc-eu-times-model-bioenergy-potentials-eu-and-neighbouring-countries](https://ec.europa.eu/jrc/en/publication/eur-scientific-and-technical-research-reports/jrc-eu-times-model-bioenergy-potentials-eu-and-neighbouring-countries)

### LINKED ISSUE #195: Align biofuel potential and cost
Because the old data source of bioenergy cost and potentials did not include data for the same years, Euro-Calliope 1.0.0 uses potentials from 2020 and costs from 2030. With updated data (#194), we can use the same year and should align it.

The year should be 2020.

timtroendle:
* Add automatic download, Fixes #194.
* Update default scenario years to align them, Fixes #195.
* Fix cost of municipal solid waste, Fixes #193.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


scripts/biofuels.py
```diff
@@ -68,81 +59,95 @@ class GlobCover(Enum):
...
+    weighted_cost = (
+        (national_costs * national_potentials / national_potentials.sum(["country_code", "feedstock"]))
+        .sum(["country_code", "feedstock"])
+        .item()
+    )
+    with open(paths_to_output.costs, "w") as f_cost:
         f_cost.write(str(weighted_cost))
 
 
 def allocate_potentials(national_potentials, units, population, land_cover):
```
brynpickering, line 104: I suspect all of this would be best done in pandas, rather than xarray:

```
shares = pd.concat([population, land_cover[FOREST].sum(axis=1), land_cover[FARM].sum(axis=1)], axis=1, keys=['population_share', 'forest_share', "farm_share"]).groupby(units.country_code).transform(lambda x: x / x.sum())
regional_potentials = pd.merge(national_potentials.reset_index(), units["country_code"].reset_index(), on="country_code").set_index(["id", "feedstock"]).unstack("feedstock")
regional_potentials_sum = regional_potentials.groupby(PROXIES, axis=1).sum()
regional_potentials_weighted = regional_potentials_sum.mul(shares).sum(axis=1)
```

scripts/biofuels.py
```diff
@@ -68,81 +59,95 @@ class GlobCover(Enum):
...
         national_potentials=national_potentials,
         units=units,
         population=pd.read_csv(path_to_population, index_col=0).reindex(index=units.index)["population_sum"],
         land_cover=pd.read_csv(path_to_land_cover, index_col=0).reindex(index=units.index)
     )
-    total_potential.to_csv(path_to_potentials, index=True, header=True)
-    weighted_cost = (costs * national_potentials / national_potentials.sum().sum()).sum().sum()
-    with open(path_to_costs, "w") as f_cost:
+    total_potential.to_csv(paths_to_output.potentials, index=True, header=True)
+    weighted_cost = (
```
brynpickering, line 95: If we have costs per region, why not keep them as regional values? They'll be the same for every subnational region, but will be different between countries.

brynpickering (changes requested):
Looks fine. Some comments to address and perhaps some new issues to create:

1. PROXIES -> config
2. retain regional distribution of costs
3. verify that your PROXIES for subnational regionalisation are similar to the NUTS2 data given in the Excel

scripts/biofuels.py
```diff
@@ -1,22 +1,13 @@
...
 GJ_TO_MWH = 1 / 3600 * 1e3
 NAME = "biofuel_potential_mwh_per_year"
 
-SCENARIOS = {
-    "low": "Low availability scenario",
-    "medium": "Medium availability scenario",
-    "high": "High availability scenario"
-}
-
 PROXIES = {
```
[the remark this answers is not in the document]
timtroendle: Yes!

scripts/biofuels.py
```diff
@@ -68,81 +59,95 @@ class GlobCover(Enum):
...
+        .sel(year=potential_year, scenario=scenario)
     )
-    costs = pd.concat(
-        [pd.read_csv(path, index_col=0)
-           .rename(index=utils.eu_country_code_to_iso3)
-           .loc[:, cost_year]
-           .rename(path.stem) / GJ_TO_MWH
-         for path in paths_to_costs],
-        axis=1
+    national_costs = (
```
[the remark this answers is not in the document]
timtroendle: Hm, most tech costs are given in Eur2013, so I'd assume both are equally good.

scripts/biofuels/extract_biofuels.py
```diff
@@ -0,0 +1,79 @@
...
+    # Sets 1 missing value (NOR, high cost 2030) to 0.
+    df = pd.read_excel(path_to_raw_data, sheet_name=sheet_name, na_values="")
+    df = df[df[COL_NAME_FEEDSTOCK].isin(FEEDSTOCK_NAME_MAP.keys())]
+    df = df[df[COL_NAME_SCENARIO].isin(SCENARIO_NAME_MAP.keys())]
+    df = df.where(df[COL_NAME_COUNTRY] != "KS").dropna(subset=[COL_NAME_COUNTRY])
+    df[COL_NAME_SCENARIO] = df[COL_NAME_SCENARIO].map(SCENARIO_NAME_MAP)
+    df[COL_NAME_FEEDSTOCK] = df[COL_NAME_FEEDSTOCK].map(FEEDSTOCK_NAME_MAP)
+    df[COL_NAME_COUNTRY] = df[COL_NAME_COUNTRY].map(utils.eu_country_code_to_iso3)
+    df.rename(columns=COL_NAME_MAP, inplace=True)
+    df.value = pd.to_numeric(df.value, errors='coerce')
```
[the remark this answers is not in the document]
timtroendle: Ok but is there a specific reason?

scripts/biofuels.py
timtroendle: Thought about the same when implementing. OK, but in a separate issue #200.

scripts/biofuels/extract_biofuels.py
```diff
@@ -0,0 +1,79 @@
...
+
+def read_filter_and_map_names(path_to_raw_data, sheet_name):
+    # ASSUME no energy crop feedstock available.
+    # Ignores scenarios other than high/med/low.
+    # Removes "KS" NUTS0 from costs, as this is an unknown abbreviation.
+    # Renames scenarios and feedstocks.
+    # Renames countries to ISO3.
+    # Renames column names
+    # Sets 1 missing value (BIH med potential 2010) to 0.
+    # Sets 1 missing value (NOR, high cost 2030) to 0.
```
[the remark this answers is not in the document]
timtroendle: ok. I am using another scenario's value as values barely change between scenarios (and are tiny anyways)

scripts/biofuels/extract_biofuels.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def read_filter_and_map_names(path_to_raw_data, sheet_name):
+    # ASSUME no energy crop feedstock available.
+    # Ignores scenarios other than high/med/low.
+    # Removes "KS" NUTS0 from costs, as this is an unknown abbreviation.
+    # Renames scenarios and feedstocks.
+    # Renames countries to ISO3.
+    # Renames column names
+    # Sets 1 missing value (BIH med potential 2010) to 0.
```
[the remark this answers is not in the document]
timtroendle: ok. But both are 0.

scripts/biofuels.py
timtroendle: Maybe. I gave this a try, but it didn't work right away. A pure Pandas approach would require more changes, as the inputs are xarray objects.

timtroendle:
Ok. I implemented most of your recommendations and I moved PROXIES (and actually much more) to the config. The other two points should be handled in other issues in my opinion (I opened #200).

scripts/biofuels.py
```diff
@@ -68,81 +45,97 @@ class GlobCover(Enum):
...
+        .transform(lambda x: x / x.sum())
+        .rename("forest")
+    )
+    ds["farmland"] = (
+        land_cover[FARM]
+        .sum(axis=1)
+        .groupby(units.country_code)
+        .transform(lambda x: x / x.sum())
+        .rename("farmland")
+    )
```
brynpickering, line 118:
```suggestion
    shares = (
        pd.concat(
            [population, land_cover[FOREST].sum(axis=1), land_cover[FARM].sum(axis=1)], 
            axis=1, 
            keys=['population', 'forest', "farmland"]
        )
        .groupby(units.country_code)
        .transform(lambda x: x / x.sum())
    )
```

scripts/biofuels.py
```diff
@@ -68,81 +45,97 @@ class GlobCover(Enum):
...
+    ds["proxy"] = pd.Series(proxies).rename("proxy").rename_axis(index="feedstock").to_xarray()
+    proxy_value = xr.ones_like(ds.value)
+    for feedstock in ds.feedstock:
+        for id in ds.id:
+            proxy = ds.sel(id=id, feedstock=feedstock).proxy.item()
+            proxy_value.loc[dict(id=id, feedstock=feedstock)] = ds.sel(id=id)[proxy].item()
+    ds["value"] = ds.value * proxy_value
+    return ds.value.sum("feedstock").rename(NAME).to_series()
```
brynpickering, line 126:
```suggestion
    regional_potentials_sum = regional_potentials.groupby(proxies, axis=1).sum()
    regional_potentials_weighted = regional_potentials_sum.mul(shares).sum(axis=1)
    return regional_potentials_weighted.rename(NAME)
```

scripts/biofuels.py
```diff
@@ -68,81 +45,97 @@ class GlobCover(Enum):
...
-        land_cover=pd.read_csv(path_to_land_cover, index_col=0).reindex(index=units.index)
+        land_cover=pd.read_csv(path_to_land_cover, index_col=0).reindex(index=units.index),
+        proxies=proxies
+    )
+    total_potential.to_csv(paths_to_output.potentials, index=True, header=True)
+    weighted_cost = (
+        (national_costs * national_potentials / national_potentials.sum(["country_code", "feedstock"]))
+        .sum(["country_code", "feedstock"])
+        .item()
     )
```
brynpickering, line 86:
```suggestion
    weighted_cost = (
          national_costs
          .mul(national_potentials)
          .div(national_potentials.sum())
          .sum()
    )
```

scripts/biofuels.py
```diff
@@ -68,81 +45,97 @@ class GlobCover(Enum):
...
+    units = pd.read_csv(path_to_units).set_index("id")
     if (len(units.index) == 1) and (units.index[0] == "EUR"): # special case for continental level
-        national_potentials = pd.DataFrame(index=["EUR"], data=national_potentials.sum(axis=0).to_dict())
-        costs = pd.DataFrame(index=["EUR"], data=costs.mean(axis=0).to_dict())
+        national_costs = (
+            (national_costs * national_potentials / national_potentials.sum(["country_code", "feedstock"]))
+            .sum(["country_code", "feedstock"])
+            .expand_dims({"country_code": ["EUR"]})
+        )
+        national_potentials = national_potentials.sum("country_code").expand_dims({"country_code": ["EUR"]})
```
brynpickering, line 72:
```suggestion
        national_potentials = (
            national_potentials
            .rename(lambda x: "EUR", level="country_code")
            .sum(level=["country_code", "feedstock"])
        )
```

scripts/biofuels.py
brynpickering: and move this above `if (len(units.index) == 1) and (units.index[0] == "EUR"):`

scripts/biofuels.py
```diff
@@ -68,81 +45,97 @@ class GlobCover(Enum):
...
-    units = gpd.read_file(path_to_units).set_index("id")
+    units = pd.read_csv(path_to_units).set_index("id")
     if (len(units.index) == 1) and (units.index[0] == "EUR"): # special case for continental level
-        national_potentials = pd.DataFrame(index=["EUR"], data=national_potentials.sum(axis=0).to_dict())
-        costs = pd.DataFrame(index=["EUR"], data=costs.mean(axis=0).to_dict())
+        national_costs = (
+            (national_costs * national_potentials / national_potentials.sum(["country_code", "feedstock"]))
+            .sum(["country_code", "feedstock"])
+            .expand_dims({"country_code": ["EUR"]})
+        )
```
brynpickering, line 71:
```suggestion
```

In preference of bringing 'weighted_costs' up from below, which is a duplication of this if undertaken before changing 'national_potentials'.

brynpickering (changes requested):
OK, I made suggestions to have everything remain in pandas the whole way through (rather than the current approach of pandas -> xarray -> pandas). I think it's cleaner and also quicker.

It's my first attempt at using 'suggestions' and I think it ultimately made it more confusing than I had hoped. Sorry. Let me know if you'd prefer that I just push a commit with the necessary updates.

On another note. Should `biofuels.py` be in the biofuels subdir?

timtroendle:
It's indeed a little complicated. Plus, you couldn't know whether it worked (it didn't), you had no code autocorrect, and you appear as commiter only, not as author. So my conclusion would be: worth the try, but not a terribly good solution. :P

Thanks in any case. This is cleaner now. Will merge once you give your ok.

brynpickering:
Haha, ok, I won't try making such lengthy suggestions again ;)

> On another note. Should biofuels.py be in the biofuels subdir?

Did you consider this?

timtroendle:
Yes, I did. Sorry, I was running the minimal workflow first before pushing. Now it's updated.

brynpickering (approved):

PR #201, opened 2021-12-22
Update retrieval of euro-calliope-datasets

timtroendle:
euro-calliope-datasets are now retrieved from Zenodo only. This has three advantages:

* We need to maintain the Zenodo repo only -- we can abandon the GitHub repo.
* We can download each dataset separately.
* We can (if needed) download datasets from different versions on Zenodo and therefore we can be more flexible with versioning.

All together should reduce complexity and make the development and all releases easier.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.syncignore-send
+++ b/.syncignore-send
@@ -8,5 +8,4 @@ __pycache__
 .DS_Store
 build
 data/automatic
-data/euro-calliope-datasets
 notebooks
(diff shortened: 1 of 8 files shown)


brynpickering (approved):
Just need to update changelog

PR #203, opened 2021-12-24
Fix minor issues in release guidelines

timtroendle:
## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -103,7 +103,7 @@ Take the existing fixtures as a starting point.
 ## 4) Release
 
 Whenever you want to publish a new release of Euro-Calliope, you need to perform several manual steps.
-Be aware that you can publish a release only if you are a maintainer of the two central GitHub repositories and if you have edit access to the three Zenodo archives linked below.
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -113,15 +113,16 @@ Be aware that you can publish a release only if you are a maintainer of the two
         * [.github/ISSUE_TEMPLATE/FEATURE-REQUEST.yml](.github/ISSUE_TEMPLATE/FEATURE-REQUEST.yml)
         * [./lib/setup.py](./lib/setup.py)
         * [./lib/eurocalliopelib/__init__.py](./lib/eurocalliopelib/__init__.py)
+        * [./templates/README.md](./templates/README.md)
```
brynpickering, line 116: This is different to the others, since it is already the 'correct' version for the pre-built (no `.dev`). Instead, it needs to be updated *after* release.

timtroendle: You are right, it's different. Would you remove it though? Even if one may not need to change it, one should still verify the correct number is mentioned. Point 5 involves the changes _after_ release, including this one.

Issue #205, opened 2022-01-22
Store and publish workflow logs


timtroendle:
We have valuable information in the logs of the workflow, e.g. we show which parts of countries are removed and which exact data sources are used for electricity load data. This information is only shown during workflow execution. In addition, it should be stored and published with the models.

We may be able to use Snakemake's `log` mechanism for that. In any case, Snakemake's linter is complaining about logs missing in all our rules.

timtroendle:
Potential duplicate of #94.

timtroendle:
Closing in favour of #94

PR #211, opened 2022-01-25
Update Snakemake profiles

timtroendle:
* add profile for local execution
* update Euler profile to match Snakemake best practises

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


profiles/euler/config.yaml
```diff
@@ -0,0 +1,13 @@
+cluster: bsub -oo {resources.log-output} -W {resources.runtime} -n {resources.cores} -R "rusage[mem={resources.memory}]" -J {resources.job-name} <
+jobs: 999
+cluster-config: "config/euler/cluster-config.yaml"
```
brynpickering, line 3:
```suggestion
```

brynpickering (commented):
Should the profiles include anything related to using mamba? Or is that the default now with the version of snakemake being used?

+ should there be a "minimal" profile, for running the minimal workflow (maybe with more cores than 1)?

timtroendle:
Mamba is the default, yes. The workflow doc tells users how to use conda instead.

I don't think there should be a `minimal` profile, as these are two different things: One is the configuration of the workflow behaviour, the other is the configuration of Snakemake. [Snakemake's doc](https://snakemake.readthedocs.io/en/stable/snakefiles/best_practices.html) itself warns to not mix the two.

What we should do though is to tell the users 

- [ ] they can use more cores by running `snakemake --profile profiles/default -j <N_CORES>`
- [ ] they can simplify the command by doing x/y (creating an alias most probably).

brynpickering:
> Snakemake's doc itself warns to not mix the two.

I'm scared to try linting the workflow at this point... Nevertheless, good to know they have some advice, I guess we should look to follow that advice beyond the specifics of this PR.

timtroendle:
The issue coming up most when linting is #205.

brynpickering:
@timtroendle what's the status here? Do you still need to update the docs the two bullet point items from your comment in this thread?

timtroendle:
Thanks for bumping this. I have to admit it's at the bottom of my priority list right now and I'd leave it there if not you convince me of changing that.

timtroendle:
@brynpickering I continued working on this but hadn't seen your commit above. I've tried to rebase onto your commit but this is causing a very long list of conflicts. Instead, I will now remove your commit to "resolve" the issue. I've fixed your fix in my incoming changes anyways.

timtroendle:
I've updated it and successfully ran the full workflow.

brynpickering (approved):

Issue #209, opened 2022-01-25
Reference all overrides in one file of scenarios


brynpickering:
### What can be improved?

Although all overrides are documented in the Euro-Calliope docs, it would make sense for ease of navigating the YAMLs to have a single file which points to all overrides. To do this, overrides could be turned into scenarios, e.g. `load-shedding: [load-shedding-override]`. In some cases, this is a chance to group overrides into scenarios.

Since Calliope v0.6.7, scenarios of scenarios are possible, so this would not remove the possibility of grouping overrides together, but would enable a user to quickly see what overrides are available without having to trawl through large YAMLs, where the overrides might be mixed with base tech/location definitions.

### Version

1.2.0.dev

timtroendle:
I am not sure this is a good idea. See #236

timtroendle:
The changelog says this is implemented. Close @brynpickering ?

timtroendle:
This would break the modularity and is therefore not implemented. See as well #236 and #241.

Issue #132, opened 2021-07-06
Add DEA bioenergy model and cost


timtroendle:
The existing model of bioenergy does not fit any of the models of DEA. We should provide an override for the bioenergy tech that follows a model within the DEA catalogue.

timtroendle:
See discussion in #128.

Issue #200, opened 2021-12-21
Regionally differentiated cost of biofuel


timtroendle:
### What can be improved?

The JRC input data for cost of biofuel is disaggregated by country. Right now, we form an average. Instead, we should feed the disaggregated forward into the model.

### Version

1.0.0

Issue #199, opened 2021-12-21
Change proxy of landscape-care-residues biofuel feedstock


timtroendle:
According to Ruiz et al (2015): "The JRC EU-TIMES model" the feedstock of "landscape-care-residues" (MINBIOFRSR1a) is biomass from forestry. Hence, the proxy should be forest, not population.

Version: 1.0.0

PR #212, opened 2022-01-25
Update script and final model file structure

### LINKED ISSUE #145: Re-structure final YAML files

### LINKED ISSUE #146: Re-structure final YAML files
As evidenced by recent PRs (#131, #139), there are many possible Calliope overrides being built in the workflow. These sometimes end up in their own file, sometimes in a file with technology definitions. We should consider re-structuring the YAML files with two key aims:

1. Modularisation. It should be possible to build the set of YAML files with any future combination of technologies/carriers in the system, without the content of the YAML files themselves changing much. 
2. Ease of tweaking in a pre-built model. There will be a set of users working directly with the pre-built model and never with this workflow. It should be intuitive for users to navigate the YAML files to update/add/remove components of the model as they see fit. This also fits for those elements that are not configurable in the Euro-Calliope Snakemake workflow, but a user could change once the model is built (e.g. in #139, a user might want to change from 4hrs being the threshold between the technology types).

Some thoughts on this:
1. Technologies and their overrides and allocation to locations could be packaged on a per-YAML file basis.
2. Overrides could be in a file of the same name as the override, all kept in an override directory.

brynpickering:
Fixes #145 .

Large restructure of final model files following discussions with various euro-calliope users and developers. 

The idea is to make things a bit more intuitive for a non-workflow user, while making the addition of model components easier and more modular for workflow users and developers.

This is necessarily the perfect final structure or naming convention, so I would welcome comments on how to improve!

Some other things changed in the process of this restructure:

1. ENTSOE TYNDP transmission links changed to an override (should 'nearest neighbour' links also be an override?)
2. model.yaml file import lists are created automatically from the template file structure
3. a 'scenarios.yaml' file consolidates overrides that have been split between files in the new structure, but were one override previously. This leaves open the option to add all overrides here, as discussed in #209.
4. a dedicated tech cost file (introduced in #132 ?) has been changed back to costs being defined with the rest of the tech definition.
5. 'load-shedding' and 'direction-rooftop-pv' are no longer independent rules. The 'direction-rooftop-pv' override is rolled into the 'wind-and-solar' template.
6. parametrising a template with Jinja2 has moved to eurocalliopelib.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


Snakefile
```diff
@@ -188,7 +124,7 @@ rule electricity_load:
         national_load = rules.electricity_load_national.output[0]
     params:
         scaling_factor = config["scaling-factors"]["power"]
-    output: "build/model/{resolution}/electricity-demand.csv"
+    output: "build/models/{resolution}/timeseries/demand/electricity-demand.csv"
```
brynpickering, line 127: Could rename this to `demand-electricity.csv` here, or leave for a follow-up PR. If follow-up, I would also update `capacityfactors...` filenames to more closely match the YAML filename counterparts.

Snakefile
```diff
@@ -304,14 +194,15 @@ rule test:
         "build/logs/{resolution}/model.done",
         test_dir = model_test_dir,
         tests = map(str, Path(model_test_dir).glob("**/test_*.py")),
-        model = test_dir + "resources/{resolution}/model.yaml",
-        example_model = "build/model/{resolution}/example-model.yaml",
+        example_model = "build/models/{resolution}/example-model.yaml",
```
brynpickering, line 197: Removed test-specific model YAMLs as they were repeating most of what was already in the example model YAML. Now, test+resolution-specfic config is defined in the workflow config and brought in here as parameters.

config/schema.yaml
```diff
@@ -244,6 +244,46 @@ properties:
...
+                minimum: 2010
+                maximum: 2016
+
+            test:
+                type: object
+                description: Resolution-specific parameters for running the model tests
+                properties:
+                    scenarios:
+                        type: object
+                        propertyNames: {"$ref": "#/properties/shapes/propertyNames"}
```
brynpickering, line 259: Can only take on a name matching a resolution defined in 'shapes' ('continental', etc.).

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+
+root_dir = config["root-directory"] + "/" if config["root-directory"] not in ["", "."] else ""
+script_dir = f"{root_dir}scripts/"
+template_dir = f"{root_dir}templates/"
+model_template_dir = f"{template_dir}model/"
+techs_template_dir = f"{model_template_dir}techs/"
+
+NON_MODEL_CONFIG_FILES = ["environment.yaml", "README.md"]
+
+MISSING_LOCATION_SPECIFIC_FILES = {
```
brynpickering, line 13: Not necessarily ideal implementation, but there didn't seem an easier way to handle this specific YAML file not being generated for non-national resolutions...

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+import glob
+from pathlib import Path
+
+root_dir = config["root-directory"] + "/" if config["root-directory"] not in ["", "."] else ""
+script_dir = f"{root_dir}scripts/"
+template_dir = f"{root_dir}templates/"
+model_template_dir = f"{template_dir}model/"
+techs_template_dir = f"{model_template_dir}techs/"
+
+NON_MODEL_CONFIG_FILES = ["environment.yaml", "README.md"]
```
brynpickering, line 11: could also read this from a `glob` of the `templates` top-level dir.

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+
+rule locations_template:
+    message: "Generate locations configuration file for {wildcards.resolution} resolution from template."
+    input:
+        script = script_dir + "template_locations.py",
+        template = model_template_dir + "locations.yaml",
+        shapes = rules.units.output[0]
+    output:
+        yaml = "build/models/{resolution}/locations.yaml",
+        csv = "build/data/{resolution}/locations.csv"
```
brynpickering, line 27: Moved from `model` to `data`, since it isn't necessary for the model to run, and the `locations.yaml` file is now slim enough to communicate the same info.

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+rule simple_techs_and_locations_template:
+    message: "Create {wildcards.resolution} tech definition file `{wildcards.template}` from template."
+    input:
+        script = script_dir + "template_techs_and_locations.py",
+        template = techs_template_dir + "{template}",
+        locations = rules.locations_template.output.csv
+    params:
+        scaling_factors = config["scaling-factors"],
+    wildcard_constraints:
+        template = "demand/demand-electricity.yaml|supply/supply-load-shedding.yaml|storage/storage-electricity.yaml"
```
brynpickering, line 41: could also avoid this wildcard constraint by setting a ruleorder so that this rule comes last. Then all remaining template files would be parametrised with this rule.

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+    output: "build/models/{resolution}/techs/{template}"
+    script: "../scripts/template_techs_and_locations.py"
+
+
+rule bio_techs_and_locations_template:
+    message: "Create biofuel tech definition file from template."
+    input:
+        script = script_dir + "biofuels/template_bio.py",
+        template = techs_template_dir + "supply/supply-biofuel.yaml",
+        biofuel_cost = "build/data/regional/biofuel/{scenario}/costs-eur-per-mwh.csv".format(
```
brynpickering, line 52: If this was in the same file as 'potential-mwh-per-year.csv', then we wouldn't need the `template_bio.py` script. This fix relates to #200.

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+    script: "../scripts/template_techs_and_locations.py"
+
+
+rule transmission_entsoe_tyndp_template:
+    message: "Create YAML file of national-scale links with ENTSO-E TYNDP net-transfer capacities"
+    input:
+        script = script_dir + "transmission/template_transmission_entsoe_tyndp.py",
+        template = techs_template_dir + "transmission/transmission-electricity-entsoe.yaml",
+        locations = rules.locations_template.output.csv,
+        entsoe_tyndp = "build/data/national/TYNDP-2020-Scenario-Datafile.xlsx"
```
brynpickering, line 112: This has to be explicitly named (not only referencing the rule output) since the rule is in the main Snakefile. TODO in follow-up PR: move all rules from the main Snakefile into sector/tech-specific snakefiles in `rules`.

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+    input:
+        template = template_dir + "{template}",
+    output: "build/models/{template}"
+    wildcard_constraints:
+        template = "|".join(NON_MODEL_CONFIG_FILES)
+    conda: "../envs/shell.yaml"
+    shell: "cp {input.template} {output}"
+
+
+def get_all_config_files(resolution):
```
brynpickering, line 162: [Snakemake best practices](https://snakemake.readthedocs.io/en/stable/snakefiles/best_practices.html) would suggest that this help function should be in a separate file. Where? Or can this helper be completely avoided?

rules/model-template.smk
```diff
@@ -0,0 +1,187 @@
...
+        template = model_template_dir + "example-model.yaml",
+        non_model_files = expand(
+            "build/models/{template}", template=NON_MODEL_CONFIG_FILES
+        ),
+        config_files = lambda wildcards: expand(
+            f"build/models/{wildcards.resolution}/{{filepath}}",
+            filepath=get_all_config_files(wildcards.resolution)
+        )
+    params:
+        model_year = config["parameters"]["model-year"]
```
brynpickering, line 184: Since a user can now generate multiple weather years of data, it makes sense to enable them to define a 'default' year within those years.

rules/wind-and-solar.smk
```diff
@@ -38,6 +38,20 @@ rule download_capacity_factors_wind_and_solar:
     shell: "curl -sLo {output} '{params.url}'"
 
 
+rule area_to_capacity_limits:
```
brynpickering, line 41: Replaces the sub-function found in the former `locations.py` script.

scripts/template_locations.py
```diff
@@ -0,0 +1,38 @@
...
+    locations = gpd.GeoDataFrame(
+        gpd.read_file(path_to_shapes).set_index("id")
+    )
+    locations = (
+        locations
+        .assign(centroid=locations.to_crs(EPSG3035).centroid.to_crs(WGS84).rename("centroid"))
+        .loc[:, ["name", "centroid"]]
+    )
+
+    locations = locations.assign(id=locations.index.str.replace(".", "-", regex=False)).set_index("id")
```
brynpickering, line 21: I wonder whether `.`->`-` should happen much sooner in the process...

scripts/template_model.py
```diff
@@ -0,0 +1,25 @@
+"""Creates Calliope top-level model file."""
+from eurocalliopelib.parametrise_template import parametrise_template
+
+
+def construct_model(path_to_template, path_to_output, config_files, resolution, model_year):
+    config_files = sorted([
```
brynpickering, line 6: Possible not a good idea to use string replace here; might create errors. Does `pathlib.Path` have an equivalent string replacement that might be more robust?

scripts/template_techs_and_locations.py
```diff
@@ -0,0 +1,25 @@
...
+from eurocalliopelib.parametrise_template import parametrise_template
+
+
+def construct_techs_and_locations(path_to_template, path_to_output, path_to_locations, params):
+
+    locations = pd.read_csv(path_to_locations, index_col=0)
+    parametrise_template(
+        path_to_template, path_to_output,
+        locations=locations,
+        **params
```
brynpickering, line 15: Just passes on all `params` that are defined in the rule.

scripts/transmission/template_transmission.py
```diff
@@ -1,38 +1,26 @@
 """Link all those locations that are neighbours."""
 import geopandas as gpd
 import shapely
-import jinja2
 import networkx as nx
 
 from eurocalliopelib.geo import EPSG3035
+from eurocalliopelib.parametrise_template import parametrise_template
```
brynpickering, line 7: TODO: add tests for this.

templates/model/example-model.yaml
```diff
@@ -0,0 +1,17 @@
+import:
+    {% for file in config_files %}
```
brynpickering, line 2: list of files is now auto-generated.

templates/model/scenarios.yaml
```diff
@@ -0,0 +1,4 @@
+scenarios:
```
brynpickering, line 1: The only scenarios I've defined here so far are those which require stitching together newly split overrides back into scenarios with the same name as former 'overrides'.

templates/model/techs/storage/storage-electricity.yaml
```diff
@@ -0,0 +1,48 @@
...
+            parent: storage
+            carrier: electricity
+        constraints:
+            energy_cap_max: inf
+            storage_cap_max: inf
+            energy_eff: 0.9273  # 0.86 round trip efficiency
+            storage_loss: 0  # No loss over time assumed
+            lifetime: 10
+        costs.monetary:
+            storage_cap: {{ 723130 * 0.14 * scaling_factors.specific_costs }} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MWh") }}, 802 $2015
```
brynpickering, line 14: I have moved costs back into where the technologies themselves are defined. This seems more intuitive.

templates/model/techs/supply/supply-hydro.yaml
```diff
@@ -0,0 +1,77 @@
...
+        hydro_run_of_river:
+            constraints:
+                energy_cap_max: {{ location.installed_capacity_hror_MW * scaling_factors.power }} # {{ (1 / scaling_factors.power) | unit("MW") }}
+        hydro_reservoir:
+            constraints:
+                energy_cap_max: {{ location.installed_capacity_hdam_MW * scaling_factors.power }} # {{ (1 / scaling_factors.power) | unit("MW") }}
+                storage_cap_max: {{ location.storage_capacity_hdam_MWh * scaling_factors.power }} # {{ (1 / scaling_factors.power) | unit("MWh") }}
+    {% endfor %}
+
+overrides:
```
brynpickering, line 41: Overrides are found alongside base tech definitions. This can make the YAML files very long and could ultimately 'hide' these overrides. Hence why I would create a scenario which encapsulates each override and place it in the top-level `scenarios.yaml` file.

templates/model/techs/supply/supply-wind-and-solar.yaml
```diff
@@ -0,0 +1,173 @@
...
+                om_annual: {{ (11340 - 1 * 8760 * capacity_factors.onshore) * scaling_factors.specific_costs }} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW/year") }}
+                om_prod: {{ 1 * scaling_factors.specific_costs }} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW") }} stolen from om_annual for technical reasons
+        techs.wind_offshore: # Sheet 21 Offshore Turbines (last updated 2019)
+            constraints.lifetime: 30
+            costs.monetary:
+                energy_cap: {{ 1777000 * scaling_factors.specific_costs}} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW") }}
+                om_annual: {{ (32448 - 1 * 8760 * capacity_factors.offshore) * scaling_factors.specific_costs }} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW/year") }}
+                om_prod: {{ 1 * scaling_factors.specific_costs }} # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW") }} stolen from om_annual for technical reasons
+
+    directional-rooftop-pv:
```
brynpickering, line 135: Moved from its own file to an override in `supply-wind-and-solar.csv`, which fits with the approach taken with overrides elsewhere.

templates/model/techs/supply/supply-wind-and-solar.yaml
```diff
@@ -0,0 +1,173 @@
+tech_groups:
```
brynpickering, line 1: I could imagine this file being split into `wind` and `solar`. The only reason I can't easily do it right now is due to the shared `available_area` of onshore wind and open field PV, which is placed at the location level. Putting this bit in 'locations.yaml' would require that rule to include the wind and PV available area CSV as an input, which seems less clean.

templates/model/techs/transmission/transmission-electricity-entsoe.yaml
```diff
@@ -0,0 +1,15 @@
+# {{ ntc_limit.title() }} net transfer capacity between countries according to the
```
brynpickering, line 1: This could be in `transmission-electricity.yaml` as an override (it would match the implementation of overrides in other files), however, I don't know how to only stitch this in for the national resolution.

brynpickering:
@timtroendle this is now ready to review following the OK from others on the structure in #145. I have left comments here and there to provide some clarity and to open discussion where I think it would be useful. Hopefully these act to help (rather than hinder) you in the review process!

rules/model-template.smk
timtroendle: Then again, you have all sorts of hard coded file names in the rules below. As long as we cannot get rid of these, maybe simply inline NON_MODEL_CONFIG_FILES in the rule below? That would be consistent.

timtroendle: Hello, it's me again. Now I understand: you need them at more than one location. What about this: These files can be differentiated by being top-level. So why don't you add a wildcard constraint to the specific rule forbidding a path separator. In the model rule, you could list them explicitly.

rules/model-template.smk
timtroendle: Not sure I agree. See my comment in the config.

rules/model-template.smk
timtroendle: My personal opinion is to be explicit here. This is _the_ place where we stitch together the entire models. I think it's fine if we are explicit. It also solves some of the issues you mention above.

We could then consider moving this rule back to the main Snakfile, just because it's so important. There is another model rule over there but isn't one of them superfluous?

rules/model-template.smk
timtroendle: If you ditch the `get_config_files` function as I propose below and if you go back to the old being-explicit approach, you do not need this but you can reuse the approach we use in v1.1

rules/model-template.smk
timtroendle: I'd go for a ruleorder, especially if there are only a few exceptions (which I think we should aim for, see my main comment).

timtroendle (changes requested):
This is great and looks like it will make both the workflow and the models cleaner. 

I left many comments inline and I have two main comments.

1. We should try to get rid of tech-specific template rules as they are making the workflow more complicated than need be (I think). Can’t we have a structure such that we need a `template/models/{class}/{tech}.yaml` and a `data/{class}/{tech}.csv` which will automatically build a `build/models/{class}/{tech}.yaml` ? Then, a workflow dev would only need to create the former two to build in a new tech. Only in special cases would one need to deviate from this. Wouldn’t that be simpler to grasp?
2. There is a lot of copy-paste going on in this PR. We need to heavily test the result of this PR to avoid regression. Maybe we can run a diff on a Calliope/netcdf/xarray object to avoid doing that by hand?

And finally: please try to bring in such a PR in smaller chunks next time. This is a huge PR and while it may seem slower to build it in chunks, I am convinced that eventually it is faster. It certainly will be less brain fuck.

rules/model-template.smk
brynpickering: OK, will move it back. It just seems an odd addition to the model files when it is the only file that isn't used to build the model. It might cause confusion with those using the pre-built model... 

Maybe there needs to be a 'name' option for locations. Something to add to the v0.7.0 to-do list...

scripts/template_locations.py
brynpickering: hmm, yes and no. The official naming according to NUTS doesn't include `.`. This is purely a GADM naming convention.

templates/model/scenarios.yaml
brynpickering: I was trying to keep this PR as small as possible ;) Maybe, given it's a monster already, I should add all overrides in here.

templates/model/techs/supply/supply-hydro.yaml
brynpickering: Except that overrides are sometimes so long that locations are hidden...

templates/model/techs/supply/supply-wind-and-solar.yaml
brynpickering: I'm generally against using a change of file imports as a way of setting up different versions of a model. It isn't Calliope best practice. This is why I would always import all files and use scenarios/overrides to shape the model setup.

Nevertheless, I agree with splitting the YAMLs.

brynpickering: For futureproofing, leaning more towards different file imports might actually be OK. We're intending on changing the file import structure in such a way that would make it more intuitive to change file imports with overrides.

Snakefile
```diff
@@ -16,15 +16,15 @@ include: "./rules/wind-and-solar.smk"
 include: "./rules/biofuels.smk"
 include: "./rules/hydro.smk"
 include: "./rules/sync.smk"
-localrules: all, download_raw_load, model, clean, parameterise_template
+include: "./rules/model-template.smk"
```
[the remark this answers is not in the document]
brynpickering: Yes, agreed. I kept it in one place to make it easier to parse. Moving to technology domain-specific rule rules will mean expanding the file structure, which I was trying to keep to a minimum. Will update to tech 'domain'-specific rule files.

docs/model/overview.md
```diff
@@ -5,29 +5,27 @@ No matter whether you have downloaded Euro-Calliope's pre-builts or you have bui
...
+|   |   |── {technology-class}                              <- Calliope base technology classes (one of `supply`, `demand`, `storage`, `transmission`).
+|   |   |   └── {technology-class}-{technology-group}.csv   <- Definition of a technology (or group of technologies) relevant to the base technology, and the allocation of that technology to nodes in the model.
+│   ├── example-model.yaml                                  <- Calliope model definition.
+|   ├── interest-rate.yaml                                  <- Interest rates of all capacity investments.
+|   ├── scenarios.yaml                                      <- scenario names which can be used to override the base model configuration.
+│   └── locations.yaml                                      <- Defines all nodes in the model, including the coordinates defining their centroids.
+├── build-metadata.yaml                                     <- Metadata of the build process.
+├── environment.yaml                                        <- Conda file defining an environment to run the model in.
+└── README.md                                               <- Basic documentation (pre-builts only).
```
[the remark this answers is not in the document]
brynpickering: Did I add that sentence? I think it was there already...

brynpickering:
Thanks for the comments and sorry (not sorry) again for the monster. Splitting up would have taken a lot of redundant work to make each PR buildable, which I was unwilling to do...

rules/model-template.smk
timtroendle: Does Calliope forbid to add unknown key value pairs? Otherwise, couldn't we simply add the name without it being a supported feature?

brynpickering: Yes, but we'd get a stream of warnings telling us of an unknown key on loading the model. Best to leave it as this CSV for now.

lib/eurocalliopelib/parametrise_template.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+def parametrise_template(path_to_template, path_to_output_yaml, **kwargs):
+    """Applies config parameters to template files."""
+
+    kwargs = _update_kwargs(**kwargs)
+    path_to_template = Path(path_to_template)
+    env = jinja2.Environment(
+        loader=jinja2.FileSystemLoader(path_to_template.parent),
+        lstrip_blocks=True, trim_blocks=True, keep_trailing_newline=True,
```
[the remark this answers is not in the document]
brynpickering: Yep, let's introduce Black, but in a separate PR.

scripts/transmission/template_transmission.py
```diff
@@ -1,38 +1,26 @@
 """Link all those locations that are neighbours."""
```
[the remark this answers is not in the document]
brynpickering: I've implemented a shared template. However, it then relies on if statements. I don't think this is inherently bad, but raises the question of how maintainable it will be if/when other transmission setups are added.

brynpickering:
So the only thing that is still causing me issues to solve is your first point in your main comment, @timtroendle.

To do this there would need to be a single CSV file providing location-specific information for each resulting YAML file which relies on location-specific information. The current setup of having all solar and wind capacity limits in one CSV wouldn't work any more. Likewise for hydro supply/storage. Would you suggest doing that (i.e. splitting the CSV-generation rules to output tech-specific content)?

brynpickering:
@FLomb and @sjpfenninger if you have an opinion on point (3) of my comment above, feel free to chip in. This is about the structure within each of the final model YAML files.

FLomb:
> @FLomb and @sjpfenninger if you have an opinion on point (3) of my comment above, feel free to chip in. This is about the structure within each of the final model YAML files.

I am a bit confused though. At some point it looked like both you and Tim agreed on having overrides in a separate YAML file, which should solve the problem of whether locations or overrides come first in a file, shouldn't it? So, are we now talking of just a few, particular remaining overrides? like directional-rooftop and freeze-hydro-capacities? or am I missing something?

brynpickering:
The agreement was to have a separate `scenarios.yaml` file which effectively lists all available overrides by name. It wouldn't include the configuration component of the overrides, only their names.

FLomb:
> The agreement was to have a separate `scenarios.yaml` file which effectively lists all available overrides by name. It wouldn't include the configuration component of the overrides, only their names.

Ok, got it. Then I agree that there probably is no "good" order. If overrides where always short ones, I would agree that they should be on top, otherwise they'd be obscured. But, as you say, sometimes overrides are even longer than locations themselves. 

However, if we do have to choose, perhaps the less error-prone option might be indeed to have overrides on top. In fact, provided that we can make clear that no file is supposed to contain only overrides, people shall always expect something to _follow_ them. And so they shall always check what's at the bottom. While, if we chose to have overrides at the bottom, I guess one could still legitmately expect a file to contain only techs/locs with no overrides at the bottom, because we don't always need them; and sometimes they might just forget to check.

Does this make sense?

brynpickering:
Sounds very sensible, thanks @FLomb!

brynpickering:
@timtroendle: this PR is now effectively ready. However, since it is a large change, I'm rerunning `all_tests` on the cluster right now and will check the pre-built model contents with that of v1.1 to see if there are any discrepancies.

brynpickering:
@timtroendle I have now completed the full set of model tests, as would be required for a version release. I've checked the models against the version 1.1 prebuilds and found no unexpected differences in the default runs (+ in the regional model run with all overrides at once). Those differences that do exist are:

1. No transmission technologies in the example models by default. This is a feature change that aligns with discussions in #149, that we should be forcing the user to think about what transmission technologies they want to include in their model. Removing all techs by default ensures they don't blindly go ahead with the 'link all neighbours' example without realising.

2. PV and wind technologies have no `energy_cap_max: .inf` defined, since the inclusion of this is related to a now closed issue in the Calliope repo.

3. the "wind" and "pv" tech groups have been removed, since we can't accommodate tech groups between files.

4. Regional-level distributions of biofuel tech storage and energy capacities are different due to changes made in #199, i.e. since the release of v1.1. National/continental-level numbers don't change.

All relevant model subsets (e.g. "loc_techs", "loc_tech_carriers", etc.) are consistent between this PR and v1.1 pre-builds.

timtroendle:
@brynpickering Before I start looking at the implementation, a brief reaction to the four points you mentioned.

2--4: all fine.

1: I tend to disagree to that change. I would like to give new users the ability to run a model _without_ thinking about the model. This change will increase the entry barrier which already is quite high. The model is called `example_model` for a reason: it is just that. I am also happy to make that point clearer.

If we keep this change, we need to document this properly (it isn't right now) and guide new users on what to do exactly.

Snakefile
```diff
@@ -3,30 +3,38 @@ from pathlib import Path
...
 wildcard_constraints:
         resolution = "continental|national|regional"
 
-root_dir = config["root-directory"] + "/" if config["root-directory"] not in ["", "."] else ""
-__version__ = open(f"{root_dir}VERSION").readlines()[0].strip()
-script_dir = f"{root_dir}scripts/"
-template_dir = f"{root_dir}templates/"
-test_dir = f"{root_dir}tests/"
-model_test_dir = f"{test_dir}model"
+ruleorder: area_to_capacity_limits > hydro_capacities > biofuels > dummy_tech_locations_template
```
timtroendle, line 29: This is unfortunate. Is there no way to avoid?

Snakefile
```diff
@@ -43,253 +51,133 @@ rule all:
...
-        flat_roof_share = config["parameters"]["roof-share"]["flat"],
-        maximum_installable_power_density = config["parameters"]["maximum-installable-power-density"],
-        scaling_factors = config["scaling-factors"],
-        biofuel_efficiency = config["parameters"]["biofuel-efficiency"]
-    output:
-        yaml = "build/model/{resolution}/locations.yaml",
-        csv = "build/model/{resolution}/locations.csv"
-    conda: "envs/geo.yaml"
-    script: "scripts/locations.py"
+rule dummy_tech_locations_template:
```
timtroendle, line 72: What is this for? Can you add an explanation?

Snakefile
timtroendle: Maybe using wildcard constraints?

Snakefile
```diff
@@ -43,253 +51,133 @@ rule all:
...
 
-rule directional_rooftop_pv:
-    message: "Generate override for directional rooftop PV in {wildcards.resolution} resolution."
+rule techs_and_locations_template:
+    message: "Create {wildcards.resolution} definition file for the {wildcards.tech_group} tech `{wildcards.tech}`."
     input:
-        script = script_dir + "directional_rooftop.py",
-        units = rules.units_without_shape.output[0],
-        land_eligibility_km2 = rules.potentials.output.land_eligibility_km2,
+        script = script_dir + "template_techs.py",
```
timtroendle, line 83: Is the indentation correct here? Could by my device.

Snakefile
```diff
@@ -43,253 +51,133 @@ rule all:
...
+                "techs/supply/hydro.yaml",
+                "techs/supply/load-shedding.yaml",
+                "techs/supply/open-field-solar-and-wind-onshore.yaml",
+                "techs/supply/rooftop-solar.yaml",
+                "techs/supply/wind-offshore.yaml",
+            ]
+        ),
+        capacityfactor_timeseries_data = expand(
+            "build/models/{{resolution}}/timeseries/supply/capacityfactors-{technology}.csv",
+            technology=ALL_CF_TIMESERIES
```
timtroendle, line 146: Call this ALL_CF_TECHNOLOGIES

Snakefile
```diff
@@ -43,253 +51,133 @@ rule all:
...
+        ),
+        demand_timeseries_data = expand(
+            "build/models/{{resolution}}/timeseries/demand/{energy_carrier}.csv",
+            energy_carrier=ALL_DEMAND_TIMESERIES
```
timtroendle, line 150: ALL-DEMAND-CARRIERS

config/default.yaml
```diff
@@ -126,6 +126,49 @@ parameters:
         projection-year: 2040
         ntc_limit: max
         energy_cap_limit: equals
+
+test:
```
timtroendle, line 130: I would not put test config into this file.it will be necessary fer all projects downstream which is not appropriate.

docs/model/overview.md
```diff
@@ -5,29 +5,27 @@ No matter whether you have downloaded Euro-Calliope's pre-builts or you have bui
...
+├── {resolution}                                            <- An individual folder for each spatial resolution.
+│   ├── timeseries                                          <- All timeseries data CSV files.
+|   |   |── supply
+|   |   |   └── capacityfactors-{technology}.csv            <- Timeseries of capacityfactors of all renewables.
+|   |   └── demand
+|   |   |   └── electricity.csv                             <- Timeseries of electricity demand on each node.
+|   ├── techs                                               <- All technology definition YAML files.
+|   |   |── {technology-class}                              <- Calliope base technology classes (one of `supply`, `demand`, `storage`, `transmission`).
+|   |   |   └── {technology-group}.csv                      <- Definition of a technology (or group of technologies) relevant to the base technology, and the allocation of that technology to nodes in the model.
+│   ├── example-model.yaml                                  <- Calliope model definition.
```
timtroendle, line 22: This is an excellent place to emphasize this is an example only.

templates/models/example-model.yaml
```diff
@@ -0,0 +1,17 @@
+import:
+    {% for file in input_files %}
```
timtroendle, line 2: 👍

timtroendle (commented):
It’s not possible for me to judge whether these changes are fine. Certainly not in the time you would need. 

As we are far away from the next release and as you said these changes won’t introduce any model changes, I am fine with you merging this in using your administrator super powers.

Snakefile
brynpickering: There is no way of doing this that I can see. If using wildcards, they would apply to the *non*-location-specific technologies, which means adding them to 'dummy_tech_locations_template'. I would prefer to distinguish all those technologies that have location-specific data to use (hence this ruleorder), but could also apply it inside the main `techs_and_locations_template` rule inputs for `locations` as follows:
```
        locations = lambda wildcards: (
            f"build/data/{wildcards.resolution}/{wildcards.tech_group}/{wildcards.tech}.csv"
            if f"{wildcards.tech_group}/{wildcards.tech}" in TECHNOLOGIES_WITH_LOCATIONAL_OVERRIDES
            else rules.locations_template.output.csv
        )
 ```
which would require specifying the locational outputs explicitly in the snakefile: 

```
TECHNOLOGIES_WITH_LOCATIONAL_OVERRIDES = [
    "supply/hydro", "supply/open-field-solar-and-wind-onshore", "supply/rooftop-solar",
    "supply/wind-offshore", "storage/hydro", "supply/biofuel"
]
```

This would replace ruleorder and `dummy_tech_locations_template`. I don't really know which is better (read: more maintainable)

Snakefile
brynpickering: It is correct.

config/default.yaml
brynpickering: Hmm. How would you separate it? Would you put it in "tests/resources" and have it loaded in as a file from there?

timtroendle: I think that's better, yes. It could be either in `test/resources` or `config/test.yaml`. Not sure which is better.

Snakefile
timtroendle: I agree that this does not look more maintainable. I guess it's fine the way it is -- we can still improve this later.

config/default.yaml
brynpickering: Moved it to resources and moved the schema section there too.

PR #214, opened 2022-03-25
Update sync to use cluster file permissions defaults

timtroendle:
## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


timtroendle:
I fixed the documentation error in #215. We need to merge #215 first.

brynpickering (approved):

PR #219, opened 2022-04-19
Fix links to pre-builts to always point to most recent version

### LINKED ISSUE #218: Do not point to v1.0 pre-builts in more recent versions of the documentation
Right now, we point to v1.0 pre-builts in various places in the documentation. The cleanest would be to replace all these links with version-specific links to Zenodo -- always pointing to the corresponding version. That would increase maintenance effort though and it would be brittle.

Instead, I suggest to always point to the newest version on Zenodo (i.e. using the version-unspecific link of Zenodo).

timtroendle:
Fixes #218  .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/.github/workflows/externallinks-config.json
+++ b/.github/workflows/externallinks-config.json
@@ -7,6 +7,6 @@
         "pattern": "schema.md"
       }
     ],
-    "timeout": "20s",
+    "timeout": "30s",
     "retryOn429": true
 }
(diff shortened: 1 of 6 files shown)


zoltanmaric (approved):

brynpickering (commented):
Does it make sense to also add the version number explicitly as a header / banner in the docs? E.g., we have it at as a subheading in the [Calliope docs](https://calliope.readthedocs.io/en/stable/)

timtroendle:
It probably would make sense. Right now, the version number is not easily identifiable from the docs. In the best case, the version number would be visible on each page. Any ideas how to do that technically? Within readthedocs, mkdocs, or simply manually?

brynpickering:
I think within mkdocs (https://www.mkdocs.org/user-guide/configuration/#extra). In rst with sphinx, the version can be referred to with `|version|` at any point in the docs.

timtroendle:
The readthedocs theme does not support displaying version. So I simply added it manually to the home page. It won't be visible on any other page of course, but it's better than nothing.

brynpickering (approved):

PR #226, opened 2022-05-12
Add nuclear technology and capacity processing.

### LINKED ISSUE #78: Add Nuclear
Nuclear is a possible technology in the future energy system, with some countries having it and some not. See here for more info: https://github.com/calliope-project/euro-calliope-2.0/issues/3

brynpickering:
Feature can accommodate using today's nuclear capacities (`energy_cap_equals`) or future capacity scenarios (`energy_cap_min/max`).

Fixes #78.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


config/default.yaml
```diff
@@ -126,6 +127,29 @@ parameters:
         projection-year: 2040
         ntc_limit: max
         energy_cap_limit: equals
+    nuclear:
```
timtroendle, line 130: It's confusing and slightly inconsistent that this contains config (`scenario`) and data `national_capacitiies`. I think the latter should not be here but rather somewhere else. It's very similar to the Geth et al data and thus, that's where this should be? That data lives on Zenodo right now.

config/default.yaml
```diff
@@ -126,6 +127,29 @@ parameters:
...
+        scenario: "2050"
+        national_capacities:
+            "2030":
```
timtroendle, line 133: This list does not contain all countries with nuclear power, does it? What about others?

config/schema.yaml
```diff
@@ -245,6 +253,40 @@ properties:
                         description: How to use ENTSO-E-based NTCs to constrain link energy capacity in the Calliope model.
                         enum: [min, equals, max]
 
+            nuclear:
+                type: object
+                description: Nuclear technology capacity scenario information
+                properties:
+                    scenario:
+                        type: string
+                        description: Scenario choice. "current" will use today's capacities based on the JRC power plant database (PPDB). All other scenarios will use the regional share of capacities from the PPDB to distribute the defined capacties.
```
timtroendle, line 262: Typo in capacities

config/schema.yaml
```diff
@@ -245,6 +253,40 @@ properties:
...
+                    national_capacities:
+                        type: object
+                        description: Per-country capacity ranges for non-current scenarios
+                        propertyNames:
+                            allOf:
+                                - not:
+                                    pattern: "current"
+                                - pattern: \w+
+                        patternProperties:
```
timtroendle, line 271: Building on my earlier comment: All of this is not really "config", is it?

rules/nuclear.smk
```diff
@@ -0,0 +1,31 @@
...
+    message: "Unzip JRC power plant units from database."
+    input: rules.jrc_power_plant_database_zipped.output[0]
+    output:
+        "data/automatic/JRC_OPEN_UNITS.csv"
+    conda: "../envs/shell.yaml"
+    shell: "unzip -o {input[0]} JRC_OPEN_UNITS.csv -d data/automatic/"
+
+
+rule nuclear_regional_capacity:
+    message: "Calculate proportion of future planned nuclear capacity will be installed in each"
```
timtroendle, line 21: This is incorrect when config value is "current", right?

templates/models/techs/supply/nuclear.yaml
```diff
@@ -0,0 +1,28 @@
+techs:
+    nuclear:  # for sources, see https://github.com/calliope-project/euro-calliope-2.0/issues/3
```
timtroendle, line 2: Please add the sources here instead of referring to this issue.

templates/models/techs/supply/nuclear.yaml
```diff
@@ -0,0 +1,28 @@
...
+        essentials:
+            name: Nuclear power
+            carrier: electricity
+            parent: supply
+        constraints:
+            energy_eff: 0.4
+            lifetime: 50
+            energy_cap_min_use: 0.8  # based on annual average capacity factors in France being approximately 0.75 - 0.85
```
timtroendle, line 10: This seems to be a crucial parameter. Should this move into the config?

timtroendle: Either way, can you add a `# ASSUME` so this can easily be detected as important assumption?

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
+import pandas as pd
+import shapely.geometry
+import geopandas as gpd
+import pycountry
+
+
+def regionalise_nuclear_capacity(
+    path_to_power_plant_database, path_to_units,
```
timtroendle, line 8: Let's go for typing being the default, no?

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
...
+    nuclear_scenario_config, path_to_output
+):
+    """
+    Use current geolocations of nuclear capacity in Europe to assign nuclear
```
timtroendle, line 12: This describes better what happens here than the rule message.

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
...
+    capacity to Euro-Calliope regions.
+
+    Args:
```
timtroendle, line 15: Nice. Properly we should do (should have done) this everywhere.

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
...
+        path_to_power_plant_database (str):
+            location of JRC powerplant database "OPEN_UNITS" CSV file
+        path_to_units (str):
+            location of euro-calliope regions geojson/shapefile
+        nuclear_scenario_config (dict):
```
timtroendle, line 20: I'd split this into two parts that are very different: scenario choice and min/max values.

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
...
+        .capacity_g  # Generating unit capacity, net (MW)
+    )
+    scenario = nuclear_scenario_config["scenario"]
+    if scenario == "current":
+        capacity_per_region = (
+            capacity_current_per_region
+            .to_frame("installed_capacity_nuclear_equals_MW")
+        )
+    else:
+        nuclear_regional_proportion = capacity_current_per_region.div(
```
timtroendle, line 51: `Capacity_current_per_region.group_by(country_code).transform(lambda x: x / x.sum())`

scripts/nuclear/regional_capacity.py
```diff
@@ -0,0 +1,79 @@
...
+        future_capacity = (
+            pd.DataFrame(nuclear_scenario_config["national_capacities"][scenario])
+            .transpose()
+            .rename(index=_iso3, columns=lambda x: f"installed_capacity_nuclear_{x}_MW")
+            .rename_axis(index="country_code")
+        )
+
+        capacity_per_region = (
+            future_capacity
+            .mul(nuclear_regional_proportion, level='country_code', axis=0)
```
timtroendle, line 63: This is an important assumption that we should mark with an `ASSUME`. Nuclear capacity is changed block-wise and therefore a linear scaling is a fair but very likely incorrect assumption.

templates/models/techs/supply/nuclear.yaml
```diff
@@ -0,0 +1,28 @@
...
+        costs:
+            monetary:
+                energy_cap: {{ 3.672e6 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW") }}
+                om_annual: {{ 76116 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW/year") }}
+                om_prod:  {{ 1.75 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MWh") }}
+                om_con:  {{ 8.25 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MWh") }}
```
timtroendle, line 16: Variable cost of 10 €/MWh seems very high to me. Are you sure this is correct? We should really cite the source here.

timtroendle: Actually it's much more than 10 €/MWh_output, right? The om_con must be multiplied by efficiency, so it's more like 20 €/MWh_output, isnt it?

templates/models/techs/supply/nuclear.yaml
```diff
@@ -0,0 +1,28 @@
...
+            name: Nuclear power
+            carrier: electricity
+            parent: supply
+        constraints:
+            energy_eff: 0.4
+            lifetime: 50
+            energy_cap_min_use: 0.8  # based on annual average capacity factors in France being approximately 0.75 - 0.85
+        costs:
+            monetary:
+                energy_cap: {{ 3.672e6 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW") }}
```
timtroendle, line 13: I'd avoid scientific notation for consistency and to avoid problems like the one we've seen last week.

timtroendle (changes requested):
The new modular structure is nice and clean!

I added a few comments. Most importantly this lacks CHANGELOG updates and a test. I think it would be good to have at least one test that tests the regional allocation.

config/default.yaml
brynpickering: It's not exactly like the Geth et al data, since it is my interpretation of possible future capacities based on a variety of sources - this should be readily editable by users if they want to use different sources or assumptions. Nevertheless, I could still move it to a data file that is accessed from Zenodo. I could also move it to a config CSV file so that it doesn't clutter the config YAML so much.

config/default.yaml
brynpickering: As mentioned above, it is a list of countries with expected nuclear capacity in 2030 according to some of my assumptions. The current nuclear capacity countries (scenario: "current") would be inferred directly from the JRC data.

rules/nuclear.smk
brynpickering: True.

templates/models/techs/supply/nuclear.yaml
brynpickering: Well, there is a discussion that takes place (ok, a monologue of mine) in which I set out the reasoning for my numbers from a variety of sources, so it isn't sufficient to just provide the sources.

scripts/nuclear/regional_capacity.py
brynpickering: What do you mean exactly? That the scenario choice and the whole config (all scenarios) are given as two separate input args?

templates/models/techs/supply/nuclear.yaml
brynpickering: [This report](https://www.diw.de/documents/publikationen/73/diw_01.c.698579.de/dp1833.pdf) mentions 10USD/MWh for fuel (+2USD/MWh O&M), but not whether it is MWH_output or input. [this paper](https://www.sciencedirect.com/science/article/pii/S2211467X17300561) gives 3211USD/kg fuel and 3.15e-6kg/kWh, which is equivalent to ~10USD/MWh for fuel.

The number I chose in the end comes from this (translated from USD2018 to EUR2015).

templates/models/techs/supply/nuclear.yaml
brynpickering: For legibility, I would be in favour of scientific notation. It's easy to imagine a zero accidentally being missed if typing very large numbers out in full. I would prefer to try finding out why that issue occurred last week since I have had zero issues using scientific notation up to now.

config/default.yaml
timtroendle: Ok you are right then, it's different and not comparable to anything we already have.

Actually, these are three different configurations. So in some way the cleanest would be to provide only a single configuration -- similar to the way we provide only one configuration to everything else.

We could provide different configurations in different configuration _files_. We'd have 2030 parameters in the `default.yaml` and could keep other configurations in `nuclear-today.yaml` and `nuclear-2050.yaml`. 
 
If you don't like this solution, I am also fine with keeping it as is. Maybe rename `national-capacities` to `scnearios` to limit the confusion.

config/default.yaml
timtroendle: What about Switzerland? Switzerland will likely still have nuclear power in 2030, right?

It's not clear to me whether 2030 and 2050 scenarios amend and overwrite capacities of today or whether they completely replace today's capacities. If it's the former, it's probably all fine.

scripts/nuclear/regional_capacity.py
timtroendle: Yes. I guess it comes back to the unusual structure of the config. This really depends on the first comment about the values in the configuration.

config/default.yaml
timtroendle: So using this approach, Switzerland won’t have nuclear power capacities in 2030 and 2050, is that correct?

config/default.yaml
brynpickering: I've tried implementing something which is somewhat simplified but raises the question of both *whether* and *where* our future nuclear capacity assumption CSVs should be stored.

config/default.yaml
brynpickering: In 2050, no. In 2030, now it does have capacity ;)

templates/models/techs/supply/nuclear.yaml
brynpickering: For now, I've switched to non-scientific notation. We can revisit this at some later point.

brynpickering:
@timtroendle do you think the technology should be in the example model by default?

config/default.yaml
timtroendle: It looks fine to me. It's certainly easy for users to come up with their own config.

CHANGELOG.md
```diff
@@ -5,11 +5,12 @@
 ### Added (models)
 
 * **ADD** top level model configuration file `scenarios.yaml` which can act as the focal point to consolidate all available overrides in the model (#209).
+* **ADD** nuclear power plant technology with capacity limits. Capacity limits can be equals to today or be bound by a minimum and maximum capacity to represent an available range in future. In either case, capacities are allocated at a subnational resolution based on linear scaling from current capacity geolocations, using the JRC power plant database.
```
timtroendle, line 8: Can you add the issue number here?

config/default.yaml
timtroendle: Ok. It was simply not clear to me how exactly this mechanism works. The new documentation makes this clear.

timtroendle:
I would add it to the example model, yes.

timtroendle (commented):
The implementation looks good to me. I still think there should be a test for the regionalisation of the capacity. It could be as simple as testing that the national total stay the same.

By the way, have you tested this with the continental resolution? Doesn't that need special treatment?

brynpickering:
> have you tested this with the continental resolution

I had, but not with loading in a custom config. Updated the script now to handle it.

> I still think there should be a test for the regionalisation of the capacity.

Is the assertion in the script sufficient? Would you like a bespoke test?

scripts/nuclear/regional_capacity.py
```diff
@@ -48,25 +55,23 @@ def regionalise_nuclear_capacity(
...
-            nuclear_scenario_df
-            .rename(index=_iso3, columns=lambda x: f"installed_capacity_nuclear_{x}_MW")
-            .rename_axis(index="country_code")
-        )
-
         capacity_per_region = (
             future_capacity
             .mul(nuclear_regional_proportion, level='country_code', axis=0)
         )
+        print(capacity_per_region, future_capacity)
```
timtroendle, line 74: That's unintended, right?

tests/model/test_model.py
```diff
@@ -39,6 +39,10 @@ def test_technologies_are_available(energy_cap, location, technologies):
             assert pd.notna(
                 energy_cap.where(energy_cap.techs.str.find(technology) > -1).sum(min_count=1).item()
             )
+        elif technology in OPTIONAL_LOCATIONAL_TECHNOLOGIES:
```
timtroendle, line 42: Not very happy about this if statement. But I guess the solution to this is outside the scope of this PR because the test had been weird and complex already before.

timtroendle (approved):
Yes I think the assert statement is pragmatic and sufficient.

Nice! Let's create some virtual nuclear waste!

tests/model/test_model.py
brynpickering: Yeah, neither am I, but the test isn't fit for purpose for these types of technologies since they fail whenever a technology isn't at a location (in the minimal example, it is Ireland).

PR #239, opened 2022-05-16
Add rule to process shared coast from EEZs

### LINKED ISSUE #238: Enable changes to spatial resolution layers and new layers
### What can be improved?

Euro-Calliope depends on some upstream data processing (namely from https://github.com/calliope-project/solar-and-wind-potentials) downloaded from Zenodo which effectively fixes the resolution layers to those defined in the default Euro-Calliope workflow. This means that the config option `shapes` cannot be really be changed. In some instances it *can* be changed since the resulting unit ID will be the same as the downloaded data, but the underlying shapes might be subtly different  (see #237 for identified difference between NUTS and GADM shapes).

To enable a user to *either* change the existing spatial resolution layers or to add their own (e.g. with a mix of national and subnational units), the following datasets need to be made shape agnostic:

- [ ] Fraction of coast per unit (to distribute EEZs).
     Implement a more performant [shared_coast.py](https://github.com/calliope-project/solar-and-wind-potentials/blob/main/scripts/shared_coast.py) from our sister repository.

- [ ] Population (to distribute biofuels and to distribute demand - see next point)
    A high-res raster can be easily accessed from JRC open data portal (as is done in `solar-and-wind-potentials`).

- [ ] Demand per unit (to distribute electricity profiles to subnational units)
    Implement allocation of industry electricity demand using the EU-ETS and Eurostat datasets (where EU-ETS isn't available) (#63).
    Use a direct population raster to distribute non-industry demands.

- [ ] land eligibility (to set capacity limits for wind and solar)
    Upload a raster derived from https://github.com/calliope-project/solar-and-wind-potentials to Zenodo and do the final step of converting to availability per shape in this repository.

- [ ] land cover (to distribute biofuels)
    Same as with land eligibility.

If these steps are achieved, #76 can probably be closed, since a user can handle arbitrary shapes in a downstream workflow (either by injecting their own 'units.geojson' or by aggregating smaller administrative units). This approach would also align with #79.

### Version

1.2.0.dev

brynpickering:
Partially fixes #238 by bringing shared coast processing directly into Euro-Calliope.

For the full regional resolution parallelised over four cores, this rule takes 5 minutes. For the minimal workflow, it takes seconds.

Where I've made changes/improvements from the version found in `solar-and-wind-potentials`:
1. I use 'within' rather than `intersects` to filter out coastal units. This is much quicker.
2. I remove small islands (defaults to any polygons <1% of a unit's area), which were heavily bloating the contribution of certain coastlines in Norway and Sweden. I assumed removing them would also speed up the later intersection but, on testing, the effect is not noticeable.
3. I intersect EEZs with the boundary of units, not the polygons. This is both quicker and (I believe) more accurately reflects the length of the coastline (previous method was to take the 'length' of the resulting intersected polygon which then changes depending on how well the EEZ followed each unit's coastline to begin with).

## Checklist

- [x] CHANGELOG updated
- [ ] Documentation updated


config/default.yaml
```diff
@@ -211,6 +211,8 @@ quality-control:
         scale-phs-according-to-geth-et-al: false
         station-nearest-basin-max-km: 1
     capacity-factor-gridcell-overlap-threshold: 0.97 # Ceuta y Melilla has smallest overlap
+    shared-coast-polygon-area-share-threshold: 0.01
+
```
timtroendle, line 215: Unnecessary newline.

config/schema.yaml
```diff
@@ -178,6 +178,7 @@ properties:
             monetary:
                 type: number
                 description: "Applied to all quantities of monetary cost (base: EUR)."
+
```
timtroendle, line 181: Unnecessary newline.

lib/eurocalliopelib/docs/__init__.py
timtroendle, line 1: Why is this necessary?

brynpickering:
Looking into the changes compared to the old method: it doesn't really matter how we calculate shared coasts. There's currently never a case where a region has more than one EEZ shape linked to it, using the old method or the new one. The percentage per region only matters for eligible area of offshore wind, which requires the results of this approach to be applied to the upstream solar-and-wind-potentials repo (or for eligible area to be calculated in euro-calliope).

Still, this PR allows us to link EEZs to (sub-)national regions without needing the `shared-coast.csv` file from the Zenodo repository without much overhead.

rules/wind-and-solar.smk
```diff
@@ -78,12 +77,26 @@ rule capacity_factors_onshore_wind_and_solar:
     script: "../scripts/wind-and-solar/capacityfactors.py"
 
 
+rule shared_coast:
+    message: "Determine share of coast length between EEZ and {wildcards.resolution} units using {threads} threads."
+    input:
+        units = rules.units.output[0],
+        continental_units = "build/data/continental/units.geojson",
```
timtroendle, line 84: This is somewhat unfortunate as it links resolutions. Is there a way that would avoid that? Could we simply merge the units to the continental resolution or would that create unnecessary high compute needs?

templates/models/techs/supply/wind-offshore.yaml
```diff
@@ -24,9 +24,11 @@ overrides:
 
 locations:
     {% for id, location in locations.iterrows() %}
+    {% if location.eligibility_offshore_wind_mw > 0 %}
```
timtroendle, line 27: Why filter? Isn't it good to be explicit?

scripts/wind-and-solar/shared_coast.py
```diff
@@ -0,0 +1,209 @@
...
+    eez = eez[eez.POL_TYPE.str.lower() != "joint regime"]
+
+    if resolution == "continental":
+        share = pd.Series(
+            index=pd.MultiIndex.from_product(
+                (["EUR"], eez.MRGID.unique()), names=["id", "MRGID"]
+            ),
+            data=1,
+        )
+
```
timtroendle, line 68: unnecessary newline

scripts/wind-and-solar/shared_coast.py
```diff
@@ -0,0 +1,209 @@
...
+    """
+    eez = gpd.read_file(path_to_eez).to_crs(EPSG3035)
+    units = gpd.read_file(path_to_units).to_crs(EPSG3035)
+
+    # ASSUME: we can ignore EEZs shared by two countries
+    # (these are very small far offshore, so removing them does not impact the results)
+    eez = eez[eez.POL_TYPE.str.lower() != "joint regime"]
+
+    if resolution == "continental":
+        share = pd.Series(
```
timtroendle, line 62: Comment could be useful:

    # All EEZ belong to the single "EUR" administrative unit.

timtroendle:
This may also enable us to fix #259.

scripts/wind-and-solar/shared_coast.py
```diff
@@ -0,0 +1,209 @@
...
+    continental_units: gpd.GeoDataFrame,
+    polygon_area_share_threshold: float,
+) -> gpd.GeoDataFrame:
+    """
+    Get the outline of all Euro-Calliope units which sit on the coast
+    (i.e., will have some share of the EEZ assigned to them)
+    """
+    # slightly increase the sub-continental polygon size so that those on the coast
+    # slightly overlap the continent polygon boundary.
+    units["geometry"] = units.geometry.buffer(units.total_bounds.mean() * 1e-6)
```
timtroendle, line 117: I see the point but this will also change the length of the coast. Possibly do this to a copy?

Not too important if that won't work as this whole concept of shared coasts is pretty rough.

scripts/wind-and-solar/shared_coast.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+
+
+def _simplify_geometries(
+    units: gpd.GeoDataFrame, polygon_area_share_threshold: float
+) -> gpd.GeoDataFrame:
+    """
+    Remove tiny islands from units to speed up the later intersection.
+    Any polygons in a multipolygon A with an area below
+    (polygon_area_share_threshold * A) will be removed.
```
timtroendle, line 139: `(polygon_area_share_threshold * size(A))`

rules/wind-and-solar.smk
timtroendle: Over in `units.py` we do it like that:
     
    layer = layer.dissolve(by=[1 for idx in layer.index])

timtroendle (approved):
Nice! Mainly minor comments, but one: this ties the continental resolution to all other three resolutions. I'd try to avoid if compute cost isn't too high. See my comments.

Sorry for the minor delay in my review.

timtroendle:
From a wall-clock perspective, this PR is very mature: it will celebrate its second birthday in 10 days.

scripts/wind-and-solar/shared_coast.py
brynpickering: We only work with proportions of the total length, so it doesn't really matter that the length changes. Will operate on a copy though

templates/models/techs/supply/wind-offshore.yaml
brynpickering: I'd prefer to never have non-coastal regions in the data as it bloats the `shared-coast.csv`, `capacityfactors-wind-offshore.csv` and `wind-offshore.yaml` files with effectively empty data. Saying that, I'll revert to the original method as this breaks some other part of the workflow that is outside the scope of this PR to try and fix

Issue #236, opened 2022-05-16
Scenarios.yaml breaks the new modularity


timtroendle:
### What happened?

Following updates in #145, scenarios in `models/{resolution}/scenarios.yaml` break the new modularity. In a way, this is a bug and a feature. It's a feature because it conveniently makes groups of overrides available to the user. It's a bug because it breaks once on of the linked modules (=techs, for example offshore wind) is removed.

### Version

1.2.0.dev

### Relevant log output

```shell
Error in apply_overrides, /Users/trtim/Develop/envs/snakemake/euro-calliope/x/y/d9f4c35c91fe54819efe2f2470f2f1e7/lib/python3.8/site-packages/calliope/preprocess/model_run.py:200
Scenario definition must be a list of override or other scenario names.
```

timtroendle:
This is fixed in #241.

PR #241, opened 2022-05-18
Update docs and remove scenarios.yaml

### LINKED ISSUE #230: Imports aren't properly documented
Following updates in #145, imports aren't properly documented. File names of the previously existing imports (transmission links) are broken. All new imports aren't mentioned at all.

brynpickering:
Fixes #230 & #236. Reverts #209.

Since we are now modular, it's impossible to have any one file that links to the contents of multiple other files (as `scenarios.yaml` was doing in grouping overrides). So, we have to remove this file and instead document how one would create their own scenarios by grouping overrides.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


timtroendle:
@brynpickering is this ready to be reviewed again?

brynpickering:
Yes!

timtroendle (approved):
I like it now!

One small thing needs to be fixed: all code boxes (yaml and python) in the overrides section of customisation.md are broken (have a look at the readthedocs preview). Not sure what's going on but it's basically unreadable right now.

brynpickering:
Hmm, the code blocks render just fine when I build it locally:

image

timtroendle:
It works for me too. Also, I figured just now, that these code blocks are broken in other parts of the documentation on readthedocs, too. That's weird. Is it possible that this is a glitch in the built?

brynpickering:
I checked the supposed dead markdown link (https://open-power-system-data.org/) and it isn't dead.

timtroendle (approved):
Thanks a lot for following up on this, @brynpickering . It all looks good to me now, apart from the erratic behaviour of the Markdown link check. We need to investigate what's going on or otherwise remove it. It's suspicious that it's consistently open-power-system-data that's failing.

PR #281, opened 2024-03-13
Add rule that generates the DAG

### LINKED ISSUE #208: Add a rule that generates the dag visualisation
### What can be improved?

The DAG visualisation must be generated manually right now. This is not so bad as long as the right software is installed. To ease this process, we should add a rule that uses a conda env to generate the dag.

### Version

1.2.0.dev

timtroendle:
Fixes #208

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Added (workflow)
 
+* **ADD** DAG rule that generates a visualisation of Snakemake's directed acyclic graph (#208).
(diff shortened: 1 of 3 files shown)


adrienmellot (approved):
I've tested the DAG and all seems good. Wondering if we should save as a .png instead of a .pdf ?

timtroendle:
We decided to stay with PDF as it's zoomable, being a vector graphic.

PR #283, opened 2024-03-13
Remove unused data from annual transport demand calculations

timtroendle:
In the rule annual_transport_demand three variables where calculated: distance, vehicles, and efficiency. We do not need the latter two. In addition, there seemed to be a problem with the calculation of vehicles. In fact, it was equal to distance. See: https://github.com/calliope-project/sector-coupled-euro-calliope/blob/4eeff47ec3b285e0280a4131f6c381cd9028cf75/src/construct/annual_transport_demand.py#L204

I now removed vehicles and efficiency where unneeded.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


adrienmellot (approved):
All seems good. I also don't see a use case for this data in the future, and if we do it's an easy fix to bring it back (and fix the vehicles data at the same time).

Issue #315, opened 2024-04-08
Process JRC IDEES, Eurostat, and CH data separately and blend once cleaned but before focussing on subsectors


brynpickering:
### What can be improved?

Rather than a JRC IDEES / Eurostat / CH processing script / function per sector, we'd do better to process that data source at once (since the steps are similar / sometimes identical per sector) and then pull in the cleaned data whenever a sector needs it. In the CH case, this also creates an example of how to bring in a single country's data and then blend it with Eurostat.

### Version

1.2.0.dev

Issue #291, opened 2024-03-21
Data missing in transport timeseries when running the default config


adrienmellot:
### What happened?

While the minimal config runs well for the implementation of the transport sector, the default config fails because:

- the annual timeseries profile data from RAMP is missing CYP. This is an easy fix by adding `CYP: ['ROU']` in the fill-missing-values dictionary in the config file
- some annual demand data for historic-electrification is missing for CHE. The file `build/data/transport/annual-road-transport-historic-electrification.csv` does not come with CHE. See log below. 

### Version

1.2.0.dev

### Relevant log output

```
(euro-calliope) adrienmellot@Adriens-MBP euro-calliope % snakemake --profile profiles/conda --configfile="config/default.yaml" "build/data/transport/timeseries/timeseries-passenger-cars-historic-electrification.csv"

Config file config/default.yaml is extended by additional config specified via the command line.
Building DAG of jobs...
Using shell: /bin/bash
Provided cores: 2
Rules claiming more threads will be scaled down.
Job stats:
job                                                          count    min threads    max threads
---------------------------------------------------------  -------  -------------  -------------
create_road_transport_timeseries_historic_electrification        1              1              1
total                                                            1              1              1

Select jobs to execute...

[Thu Mar 21 15:50:44 2024]
Job 0: Create timeseries for historic electrified road transport demand
Reason: Missing output files: build/data/transport/timeseries/timeseries-passenger-cars-historic-electrification.csv

Activating conda environment: ../envs/snakemake/euro-calliope/532acc2ee043cde2d8eb0d26b6750c89_
Activating conda environment: ../envs/snakemake/euro-calliope/532acc2ee043cde2d8eb0d26b6750c89_
/Users/adrienmellot/Documents/GitHub/envs/snakemake/euro-calliope/532acc2ee043cde2d8eb0d26b6750c89_/lib/python3.9/site-packages/numpy/lib/arraysetops.py:583: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  mask |= (ar1 == a)

Traceback (most recent call last):
  File "/Users/adrienmellot/Documents/GitHub/euro-calliope/.snakemake/scripts/tmp2ebh8nme.road_transport_timeseries.py", line 84, in <module>
    create_road_transport_demand_timeseries(
[...]

PR #317, opened 2024-04-08
Feature ch-stats as a separate folder

### LINKED ISSUE #313: Separate ch-stats downloading and processing
### What can be improved?

The branch 'feature-sector-coupling' had separated Swiss stats download and pre-processing to match the formatting of Eurostat's data processing. We want to merge this back into the develop branch. 

This should also help fixing #291 which fails because of Swiss historical data for transport. It seems that the historical data for Swiss transport is not processed currently.

### Version

1.2.0.dev

adrienmellot:
Fixes #313

Co-authored by: Francesco Sanvito sanvitofrancesco@gmail.com

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -25,6 +25,7 @@ include: "./rules/nuclear.smk"
 include: "./rules/transport.smk"
 include: "./rules/sync.smk"
 include: "./rules/heat.smk"
+include: "./rules/ch-stats.smk"
 min_version("7.8")
 localrules: all, clean
 wildcard_constraints:
--- a/lib/eurocalliopelib/utils.py
+++ b/lib/eurocalliopelib/utils.py
@@ -1,7 +1,9 @@
 """Utility functions."""
-
 import pycountry
+import pandas as pd
+import xarray as xr
 
+from string import digits
 
 def eu_country_code_to_iso3(eu_country_code):
     """Converts EU country code to ISO 3166 alpha 3.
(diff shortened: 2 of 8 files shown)


sjpfenninger:
I've added a fix for the linux verison conflicts issue. The PR now builds for me successfully on linux-x86. @brynpickering @adrienmellot can you maybe do a check if this doesn't break on macOS in turn?

timtroendle:
Thanks for the PR. This is a large PR that doesn't seem to add any feature as far as I can tell. Would it be possible to refactor this change to something that actually adds a feature or fixes a bug? Why not proving a fix for #291? It will be really difficult to review this without a clear purpose.

I understand this may be the wrong time and it may not be reasonable to refactor it now. In that case, let me know and I will start trying to review.

adrienmellot:
Fair point. Initially the point for this PR was to move towards #315 gradually, by first including ch-stats and then including Eurostat processing separately, before finally merging both with the actual workflow. In the meantime other commits have piled up to solve environment issues on linux. As you point out this doesn't solve anything yet, and in the meantime #291 was solved separately.

brynpickering:
@timtroendle: as @adrienmellot mentioned, this is part of a wider aim to split out data pre-processing based on source rather than sector. If we want (relatively) small PRs, we cannot expect every one to fit neatly into a feature or bug fix - some are simply refactoring or a step in the right direction towards an end goal. In this case, @adrienmellot has linked to an issue that describes the purpose of this PR and I have updated the comment to link to the meta issue.

Issue #267, opened 2023-06-05
Error in rule `download_basins_database`


jnnr:
### What happened?

When executing the euro-calliope workflow in Linux, I get an error when trying to download `raw-hydro-basins.zip` from Dropbox.

```
Error in rule download_basins_database:
    jobid: 91
    output: data/automatic/raw-hydro-basins.zip
    conda-env: .snakemake/conda/9ce49abf6f4c5d02c089fd3ffcd28a1a
    shell:
        curl -sLo data/automatic/raw-hydro-basins.zip 'https://www.dropbox.com/sh/hmpwobbz9qixxpe/AADeU9iCgMd3ZO1KgrFmfWu6a/HydroBASINS/standard/eu/hybas_eu_lev07_v1c.zip?dl=1'
        (one of the commands exited with non-zero exit code; note that snakemake uses bash strict mode!)
```

The above terminal output is not very helpful - no specific error message. 

I executed the command `curl -sLo data/automatic/raw-hydro-basins.zip 'https://www.dropbox.com/sh/hmpwobbz9qixxpe/AADeU9iCgMd3ZO1KgrFmfWu6a/HydroBASINS/standard/eu/hybas_eu_lev07_v1c.zip?dl=1'` in my basic conda env and got no problem. When switching to the environment managed by snakemake, it failed.

Had a look at the logs (store it by calling curl -sLo .... >curl_log.log 2>&1"). In the log, I saw `HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)`

Calling curl with `--http1.1` solves the problem. Dropbox offers http2, though, so what is causing the problem?

### Version

1.2.0.dev

timtroendle:
Hi @jnnr, sorry for remaining silent for so long on this issue. I cannot reproduce this problem, neither with the curl version we use in this workflow nor with my local one, nor on the compute cluster.

I am happy to merge your PR but I'm a bit uncomfortable with not understanding it and falling back to such an old version of HTTP. Why only on your machine? Why only this rule? Shouldn't we touch all instances of curl in the workflow?

We manage curl versions in the workflow using conda, precisely to avoid situations like this. We currently use [version 7.76.0](https://github.com/calliope-project/euro-calliope/blob/ab44ce99e88fb06b42f290298bd5211f311c73dc/envs/shell.yaml#L5). Could you try updating to the [newest version on conda-forge](https://anaconda.org/conda-forge/curl), `8.6.0`?

[...]

PR #323, opened 2024-04-09
Fix error messages of curl

timtroendle:
The `s` option makes curl silent, including error messages. The newly added `S` option keeps curl silent but ensures error messages are not silent.

This is an improvement for #267 but not a fix.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


brynpickering:
I assume you're correct in it being `-sS` and not just `-S` (haven't checked the man page)?

timtroendle:
The `-sS` option ensures that curl is silent apart from error messages. In particular, there won't be the progress bar which is somewhat problematic in parallel execution.

brynpickering (approved):

PR #331, opened 2024-04-09
Improve creation of symlink to eurocalliopelib

sjpfenninger:
Fixes a newly-identified issue where if you move or rename the workflow directory after already having run once, `ensure_lib_folder_is_linked` does not work because the symlink exists but points at an invalid location.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,7 @@
 
 * **FIX** fixed optimisation tolerance of hydro power plants from xtol to xatol (#266).
 * **FIX** fixed rule `download_basins_database`, which previously failed on some linux and mac machines, by requiring a more recent curl in the environment `envs/shell.yaml` (#267).
+* **FIX** fixed creation of symlink to eurocalliopelib if workflow directory has been moved or renamed.
 
 ## 1.1.0 (2021-12-22)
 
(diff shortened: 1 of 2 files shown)


Snakefile
```diff
@@ -39,14 +39,18 @@ ALL_CF_TECHNOLOGIES = [
     "hydro-reservoir"
 ]
 
+
 def ensure_lib_folder_is_linked():
     if not workflow.conda_prefix:
         return
     link = Path(workflow.conda_prefix) / "lib"
-    if not link.exists():
+    if not link.exists():  # Returns False if link exists but is an invalid symlink
```
timtroendle, line 47: Took me a while to understand this comment given the amount of negations in this line.

Alternative suggestion for a comment in the next line: "Link either does not exist or is an invalid symlink."

CHANGELOG.md
```diff
@@ -57,6 +57,7 @@
 
 * **FIX** fixed optimisation tolerance of hydro power plants from xtol to xatol (#266).
 * **FIX** fixed rule `download_basins_database`, which previously failed on some linux and mac machines, by requiring a more recent curl in the environment `envs/shell.yaml` (#267).
+* **FIX** fixed creation of symlink to eurocalliopelib if workflow directory has been moved or renamed.
```
timtroendle, line 60: I would not call this a fix in the meaning of the changelog. The bug has never been released. Instead, I'd suggest to add this to line 46 in which this was introduced.

The line reads: "* **UPDATE** default conda prefix directory including consistent handling of the path to eurocalliopelib (#264)."

and I'd change it to: "* **UPDATE** default conda prefix directory including consistent handling of the path to eurocalliopelib (#264, #331)."

timtroendle (approved):
Looks good to me!

I'd change one comment and the location of the documentation of the change.

PR #333, opened 2024-04-09
Update the troubleshooting section in the docs

timtroendle:
Fixes # .

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/docs/workflow/troubleshooting.md
+++ b/docs/workflow/troubleshooting.md
@@ -41,7 +41,7 @@ You can now inspect this file to learn about the details of how hydro data is pr
 
 ### How can I debug a workflow step?
 
-Most of our workflow steps are Python scripts and you can debug them with default debugging methods. For example, you may include a `breakpoint()` statement in a line in which you want to drop into the debugger `pdb`.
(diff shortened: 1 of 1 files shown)


docs/workflow/troubleshooting.md
```diff
@@ -41,7 +41,7 @@ You can now inspect this file to learn about the details of how hydro data is pr
 
 ### How can I debug a workflow step?
 
-Most of our workflow steps are Python scripts and you can debug them with default debugging methods. For example, you may include a `breakpoint()` statement in a line in which you want to drop into the debugger `pdb`.
+Most of our workflow steps are Python scripts and you can debug them with default debugging methods. For example, you may include a `breakpoint()` statement in a line in which you want to drop into the debugger `ipdb`.
```
brynpickering, line 44:
```suggestion
Most of our workflow steps are Python scripts and you can debug them with default debugging methods. For example, you may include a `breakpoint()` statement in a line in which you want to drop into the debugger `ipdb` (the iPython-enabled [pdb](https://docs.python.org/3/library/)).
```

brynpickering (approved):

Issue #15, opened 2020-04-14
Don't forget about Iceland


brynpickering:
Iceland is always left out of European energy system models, poor thing.

Most European datasets already cover Iceland, and a DC interconnector is expected between it (her/him/them?) and the UK, so it seems worthwhile. Only thing is that we'll need to update some upstream dependencies, such as the datasets of technically eligible land area, currently grabbed from a [static Zenodo repository](https://zenodo.org/record/3533038).

timtroendle:
Good idea. If the input data sets of [the renewable potentials source](https://github.com/timtroendle/possibility-for-electricity-autarky) all cover Iceland, determining the technically eligible land area and updating the Zenodo repository will be easy.

PR #336, opened 2024-04-10
Feature ehighways resolution unts 318

### LINKED ISSUE #318: Build ehighways resolution units
### What can be improved?

Take regional level NUTS3 data using the correct NUTS year and aggregate up to eHighways resolution. See "data/statistical_units_to_ehighways_regions.csv" in sector-coupled repo for aggregation mapping

### Version

1.2.0.dev

jonathan-peel:
Fixes #318  .

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


jnnr:
A a note: ehighways uses NUTS3 2006 (checked by comparing the region names in anderski_2014 with `statistal-to-ehighways-units.csv)

scripts/shapes/units.py
timtroendle, line 1: Many functions in this file lack type hints.

config/default.yaml
```diff
@@ -39,6 +39,9 @@ data-pre-processing:
             MNE: [HRV]
             SRB: [HUN]
             CYP: [ROU]
+    statistical-to-custom-units:
+        # FYI: E-highways were initially defined using NUTS3 units from 2006, but this mapping was manually created to map NUTS units from any year up to 2016 to E-highways. This is picked up by rule `units` to build custom spatial units.
```
timtroendle, line 43: This comments seams out of context here. I don't know where this should go, but this looks like the wrong place to me.

config/default.yaml
```diff
@@ -314,6 +317,7 @@ scope:
         first-year: 2016
         final-year: 2016
 shapes: # This config must be consistent with data from https://doi.org/10.5281/zenodo.3244985.
+    # TODO: update this link to Bryn's new one, which includes the potentials for ehighways
```
timtroendle, line 320: Please update the link

config/default.yaml
```diff
@@ -423,3 +427,39 @@ shapes: # This config must be consistent with data from https://doi.org/10.5281/
...
+        Sweden: nuts3
+        United Kingdom: nuts3
+        Albania: gadm0
+        Bosnia and Herzegovina: gadm0
+        Macedonia, Republic of: gadm0
+        Montenegro: gadm0
+        Norway: nuts3
+        Serbia: gadm0
+        Switzerland: nuts3
+        Iceland: nuts0
```
timtroendle, line 465: Iceland is not currently supported by Euro-Calliope, see #15.

config/schema.yaml
```diff
@@ -141,6 +141,23 @@ properties:
                                     type: string
                                     pattern: ^[A-Z][A-Z][A-Z]$
                                     description: Country to fill missing values from.
+            statistical-to-custom-units:
+                type: object
+                description: >-
+                    Path to a csv file which defines how lower-level statistical units
+                    (i.e., NUTS and GADM) are combined to form units of a custom
```
timtroendle, line 148: GADM are not statistical units.

config/schema.yaml
```diff
@@ -141,6 +141,23 @@ properties:
...
+                    resolutions (e.g., ehighways). FYI: E-highways were initially
```
timtroendle, line 149: The FYI also seems ill-placed here, I am sorry to say.  This is a description of the configuration parameters, but your comment is a comment to the parameter _value_.

config/schema.yaml
```diff
@@ -377,6 +394,10 @@ properties:
                 type: string
                 description: Nuclear technology capacity scenario. "current" will use today's capacities based on the JRC power plant database (PPDB). All other scenario names should point to a configuration CSV file specifying minimum and maximum capacities per country in MW with non-zero capacities (columns=[country, min, max]).
                 pattern: ^(current)|(\w+.csv)$
+            nuts-year:
+                description: Version of the Nomenclature of territorial units for statistics to use in creating spatial units (see https://ec.europa.eu/eurostat/web/nuts/history for more info)
```
timtroendle, line 398: Nomenclature of Territorial Units for Statistics

config/schema.yaml
```diff
@@ -514,6 +535,7 @@ properties:
                                 - "Norway"
                                 - "Serbia"
                                 - "Switzerland"
+                                - "Iceland" # added for ehighways
```
timtroendle, line 538: Please remove. Adding Iceland requires other changes, too.

config/schema.yaml
```diff
@@ -680,10 +702,13 @@ properties:
                         type: string
                         description: Region code
     shapes:
-        description: Source for geospatial shapes of all locations. The config must be consistent with data from https://doi.org/10.5281/zenodo.3244985.
+        description: >-
+            Source for the lowest-level spatial units of all locations.
+            The config must be consistent with data from https://doi.org/10.5281/zenodo.3244985.
+        # TODO: replace zenodo link with Bryn's new one, which includes the potentials for ehighways
```
timtroendle, line 708: Please update link

rules/shapes.smk
```diff
@@ -78,16 +78,16 @@ rule administrative_borders_nuts:
     script: "../scripts/shapes/nuts.py"
 
 
-rule units:
+rule units: # for original resolutions of regional, national, continental
```
timtroendle, line 81: The comment is incorrect, right? This is for all resolutions.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
 
 def remix_units(
-    path_to_nuts, path_to_gadm, path_to_output, layer_name, layer_config, all_countries
+    path_to_nuts: str,
+    path_to_gadm: str,
+    path_to_output: str,
+    resolution,
+    resolution_config,
+    all_countries,
+    path_to_statistical_to_custom_units,
```
timtroendle, line 20: Last four parameters and output of the function miss type hints.

scripts/shapes/units.py
```diff
@@ -86,12 +182,13 @@ def _write_layer(gdf, path_to_file):
...
-    layer_name = snakemake.wildcards[0]
+    resolution = snakemake.wildcards[0]
     remix_units(
         path_to_nuts=snakemake.input.nuts,
         path_to_gadm=snakemake.input.gadm,
         path_to_output=snakemake.output[0],
-        layer_name=layer_name,
         all_countries=snakemake.params.all_countries,
-        layer_config=snakemake.params.layer_configs[layer_name],
+        resolution_config=snakemake.params.resolution_configs[resolution],
```
timtroendle, line 191: I understand this construct existed before, but it's suboptimal. Please hand over only the configuration of the right resolution to the rule, not the configs of all resolutions.

scripts/shapes/units.py
```diff
@@ -41,24 +149,12 @@ def _validate_source_layers(source_layers):
     ), "Source layers have different crs. They must match."
 
 
-def _validate_layer_config(all_countries, layer_config, layer_name):
-    assert all(country in layer_config for country in all_countries), (
+def _validate_resolution_config(all_countries, resolution_config, layer_name):
+    assert all(country in resolution_config for country in all_countries), (
         f"Layer {layer_name} is not correctly " "defined."
```
timtroendle, line 154: There is a superfluous pair of quotes.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+    units.loc[units["type"].isnull(), "name"] = "custom_unit"
+    units["type"].fillna("custom_unit", inplace=True)
+
+    return units
+
+
+def _build_layer(resolution_config, source_layers):
+    """
+    Inputs:
+    - resolution_config: a mapping of which statistical unit resolution each country should use as its base
```
timtroendle, line 110: The units are not necessarily statistically. NUTS is, GADM isn't.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+    - source_layers: a dict with keys as each statistical unit type (nuts[0-3] or gadmn[0-3])
```
timtroendle, line 111: Same here.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+        ]
+        for country, source_layer in resolution_config.items()
+    ])
+    assert isinstance(layer, pd.DataFrame)
+    return gpd.GeoDataFrame(layer, crs=crs)
 
 
 def _read_source_layers(path_to_nuts, path_to_gadm):
+    """Returns:
+    - source_layers: a dict with keys as each statistical unit type (nuts[0-3] or gadmn[0-3])
```
timtroendle, line 131: GADM isn't statistical.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
-        layer = _continental_layer(layer)
-    _write_layer(layer, path_to_output)
+    _validate_resolution_config(all_countries, resolution_config, resolution)
+    units = _build_layer(resolution_config, source_layers)
+    _validate_layer(units, resolution, all_countries)
+    if resolution == "continental":
+        # sum all of the units together into one block
+        units = _continental_layer(units)
+    elif path_to_statistical_to_custom_units != []:
+        # this code is only executed if there is a [statistical-to-custom-units file corresponding to the chosen resolution] in the config
```
timtroendle, line 32: What are the brackets for?

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+        # sum units together according to custom mapping
+        units = _custom_layer(units, path_to_statistical_to_custom_units)
+    _write_layer(units, path_to_output)
+
+
+def _custom_layer(base_units, path_to_statistical_to_custom_units):
+    """Combine the lower-level statistical units (aspecified in resolution_config)
```
timtroendle, line 39: Typo in comment.

timtroendle: The units are not necessarily statistical.

timtroendle: Is it useful to mention `resolution_config` here? It's not part of the function anyway.

config/default.yaml
```diff
@@ -39,6 +39,9 @@ data-pre-processing:
             MNE: [HRV]
             SRB: [HUN]
             CYP: [ROU]
+    statistical-to-custom-units:
```
timtroendle, line 42: As mentioned in some other comments, the base units are not necessarily statistical. They are for eHighways, but if you want to be more flexible than eHighways (which you don't have to be, but which you are), then this does not make sense. Why not: `base-to-custom-units`?

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+    into the custom units, according to statistical_to_custom_units, using _merge_units
+    function.
+
+    Inputs:
+    - path_to_statistical_to_custom_units: csv file that specifies which NUTS and GADM
```
timtroendle, line 44: Here again, there should be no `statistical` in the name if you want to be more flexible than eHighways.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+        units contribute to each custom unit
+    - layer: the GeoDataFrame covering the whole scope, where each country is formed
+        from units corresponding to the resolution specified in resolution_config
+    Outputs:
+    -
```
timtroendle, line 49: Output is missing.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+
+    Assumes:
+    - the base-level statistical units in layer and are either NUTS3 or whole countries (NUTS0 or GADM0)
```
timtroendle, line 52: GADM is not... you know what. :D

config/schema.yaml
```diff
@@ -377,6 +394,10 @@ properties:
                 type: string
                 description: Nuclear technology capacity scenario. "current" will use today's capacities based on the JRC power plant database (PPDB). All other scenario names should point to a configuration CSV file specifying minimum and maximum capacities per country in MW with non-zero capacities (columns=[country, min, max]).
                 pattern: ^(current)|(\w+.csv)$
+            nuts-year:
```
timtroendle, line 397: This parameter does not exist in the config. Please remove.

scripts/shapes/units.py
```diff
@@ -4,25 +4,133 @@
...
+    - any units which are whole countries will also be whole countries at the custom resolution
+    """
+
+    stat_to_custom_units_df = pd.read_csv(path_to_statistical_to_custom_units, header=0)
+    nuts_year = 2013  # This was a variable in Bryn's SC EC, but hardcoded to 2013 in EC. Consider that the year needs to match for both the (downloaded nuts units) and (nuts to custom region mapping)
```
timtroendle, line 57: Make this a constant at the top of the file.

timtroendle: I'd remove the first sentence. Alternatively, point to the line in the repo. "Bryn's SC EC" will not make sense to everyone and forever.

timtroendle (changes requested):
Awesome! It looks good, but it requires a few changes, like type hints, documentation, and solving some inconsistencies.

The one thing I am not liking stems from the source: There are now two different ways in which resolutions are defined: the previously-existing `shape` config in the config AND the new `statistical-to-custom-units` file. Ideally, there would be only one place.

I don't think this should be handled within this PR or ticket. But I would like to keep this in mind and consider solving in the future. @brynpickering, do you agree?

brynpickering:
> I don't think this should be handled within this PR or ticket. But I would like to keep this in mind and consider solving in the future. @brynpickering, do you agree?

I don't think it's necessarily a problem. You have standardised boundaries coming from various data sources (nuts, gadm, ...) and they are your building blocks in `shapes`. Then you have a mechanism to merge those shapes together with a custom mapping. It means we know that all shapes in the model are built from standard boundaries and there's no need for a mechanism to injest custom shapes.

brynpickering (commented):
Not sure what the issue is with setting a nuts year at the config level and then downloading that nuts year when generating any units (as I did in SC-EC). Since we only use NUTS for `ehighways` at the moment it effectively allows you to choose the NUTS year that will be used to create those shapes.

We'll need the ability at a later point to download arbitrary NUTS years (for subnationalising using Eurostat datasets) so I guess we can leave it until then to introduce it as a feature.

Issue #274, opened 2024-03-08
Add non-electrified transport supply techs


timtroendle:
### What can be improved?

In accordance with #273, this includes providing different suppy technologies that can supply the non-electrified transport demand.

### Version

1.2.0.dev

Issue #273, opened 2024-03-08
Add non-electrified transport demand


timtroendle:
### What can be improved?

This should include the sectors passenger, freight, shipping, and aviation.

### Version

1.2.0.dev

timtroendle:
Blocked by #269 and #272 .

Issue #81, opened 2021-04-26
Add electric vehicles with timeseries demand


brynpickering:
As part of adding transport demand (#17), we would resolve passenger and freight road transport. Both of these would be given in a demand for distance travelled over the year. Electric vehicles are an interesting component of our future energy system, and we should model the timeseries profile of their plug-in schedules. More vehicles plugged in = more potential to charge vehicles. The number of EVs the system chooses to invest in then dictates the quantity of battery capacity in the system and therefore the maximum potential for the system to charge vehicles (if 50% of vehicles are EVs and there are 100 million vehicles in total, and any EV has a 30kWh battery with 7kW max charge capability, then we can we can pump 350GW into EVs when they are all plugged in, or ~260GW when 75% are plugged in). 

In addition to this plug-in charge potential, we need to handle the fact that EV charge demand should be met at least on a weekly/monthly level. That is, the system *could* meet all demand based on charging all plugged-in vehicles at full capacity March-November, but this would give 3 months with no vehicles being charged. Conversely, it does not makes sense to use an hourly EV charge demand, since the reality is that electricity isn't needed at exactly the same time as it's being used. To model smart charging, we should allow the system to charge EVs whenever is most convenient on a system level, but such that all demand for electricity is still met at a weekly/monthly level.

Much of this functionality relies on updates to constraints and allowable timeseries parameters in Calliope (see constraints [here](https://github.com/calliope-project/calliope/blob/euro-calliope/calliope/backend/pyomo/constraints/eurocalliope.py)).

This has been implemented in our sector-coupled sibling [here](https://github.com/calliope-project/euro-calliope-2.0/blob/main/src/construct/hourly_ev_profiles.py) and in the form of new tech templates in this repositories [sector-coupled branch](https://github.com/calliope-project/euro-calliope/blob/sector-coupled/src/template/transport-techs.yaml).

- [ ] Add EV plug-in potential (RAMP-mobility)
- [ ] Add EV max charge capability based on number of vehicles (JRC IDEES)
- [ ] Add EV weekly/monthly demand (RAMP-mobility)

timtroendle:
This duplicates or complements #271 .

PR #338, opened 2024-04-10
Feature uncontrolled vs controlled charging

adrienmellot:
This PR does not fix any issue but is a step towards fixing #274 , #273 , #81 .

This separates fully electrified road transport demand into a controlled part (for now, as an annual demand to be met without any constraints), and an uncontrolled part (as a fixed timeseries profile). This share is a parameter to be decided by in the config file.

To note:
- There is an inconsistency with the override "keep-historic-electricity-demand-from-road-transport" because we've assumed historical electrified transport to be uncontrolled. If you activate the override, you are still counting this electricity demand as uncontrolled.

Co-authored by: Francesco Sanvito [sanvitofrancesco@gmail.com](mailto:sanvitofrancesco@gmail.com)
 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated

### DIFF
--- a/Snakefile
+++ b/Snakefile
@@ -175,8 +175,8 @@ rule model_template:
         ),
         demand_timeseries_data = (
             "build/models/{resolution}/timeseries/demand/electricity.csv",
-            "build/models/{resolution}/timeseries/demand/electrified-road-transport.csv",
-            "build/models/{resolution}/timeseries/demand/road-transport-historic-electrification.csv",
+            "build/models/{resolution}/timeseries/demand/uncontrolled-electrified-road-transport.csv",
+            "build/models/{resolution}/timeseries/demand/uncontrolled-road-transport-historic-electrification.csv",
             "build/models/{resolution}/timeseries/demand/electrified-heat-demand.csv",
             "build/models/{resolution}/timeseries/demand/heat-demand-historic-electrification.csv",
         ),
--- a/Snakefile
+++ b/Snakefile
@@ -32,6 +32,7 @@ wildcard_constraints:
 
 ruleorder: area_to_capacity_limits > hydro_capacities > biofuels > nuclear_regional_capacity > dummy_tech_locations_template
 ruleorder: bio_techs_and_locations_template > techs_and_locations_template
+ruleorder: create_controlled_road_transport_annual_demand > dummy_tech_locations_template
 
 ALL_CF_TECHNOLOGIES = [
     "wind-onshore", "wind-offshore", "open-field-pv",
@@ -98,7 +99,7 @@ rule all_tests:
 
 
 rule dummy_tech_locations_template:  # needed to provide `techs_and_locations_template` with a locational CSV linked to each technology that has no location-specific data to define.
-    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."
+    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."  # Not used only if ruleorder is updated at top of file
     input: rules.locations_template.output.csv
     output: "build/data/{resolution}/{tech_group}/{tech}.csv"
     conda: "envs/shell.yaml"
--- a/Snakefile
+++ b/Snakefile
@@ -99,7 +99,7 @@ rule all_tests:
 
 
 rule dummy_tech_locations_template:  # needed to provide `techs_and_locations_template` with a locational CSV linked to each technology that has no location-specific data to define.
-    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."  # Not used only if ruleorder is updated at top of file
+    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."  # Update ruleorder at the top of the file if you instead want the techs_and_locations_template rule to be used to generate a file
     input: rules.locations_template.output.csv
     output: "build/data/{resolution}/{tech_group}/{tech}.csv"
     conda: "envs/shell.yaml"
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -162,6 +162,7 @@ parameters:
             coaches-and-buses: Motor coaches, buses and trolley buses
             passenger-cars: Passenger cars
             motorcycles: Powered 2-wheelers
+        uncontrolled-charging-share: 1
     entsoe-tyndp:
         scenario: National Trends
         grid: Reference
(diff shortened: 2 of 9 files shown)


Snakefile
```diff
@@ -98,7 +99,7 @@ rule all_tests:
 
 
 rule dummy_tech_locations_template:  # needed to provide `techs_and_locations_template` with a locational CSV linked to each technology that has no location-specific data to define.
-    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."
+    message: "Create empty {wildcards.resolution} location-specific data file for the {wildcards.tech_group} tech `{wildcards.tech}`."  # Not used only if ruleorder is updated at top of file
```
brynpickering, line 102: Maybe update to: `Update ruleorder at the top of the file if you instead want the techs_and_locations_template rule to be used to generate a file`

config/default.yaml
```diff
@@ -162,6 +162,7 @@ parameters:
             coaches-and-buses: Motor coaches, buses and trolley buses
             passenger-cars: Passenger cars
             motorcycles: Powered 2-wheelers
+        uncontrolled-charging-share: 0.5
```
brynpickering, line 165:
```suggestion
        uncontrolled-ev-charging-share: 0.5
```

rules/transport.smk
```diff
@@ -1,7 +1,7 @@
 """Rules to process transport sector data."""
 
 
-rule download_transport_timeseries:
+rule download_transport_timeseries: # TODO have correct timeseries data once RAMP has generated the new charging profile and it's been put on Zenodo
```
brynpickering, line 4: Add the TODO on the line below. Makes it easier to read

rules/transport.smk
```diff
@@ -60,14 +80,14 @@ rule create_road_transport_timeseries:
...
         vehicle_type = "light-duty-vehicles|heavy-duty-vehicles|coaches-and-buses|passenger-cars|motorcycles"
     output:
-        main = "build/data/transport/timeseries/timeseries-{vehicle_type}.csv",
+        main = "build/data/transport/timeseries/timeseries-uncontrolled-{vehicle_type}.csv",
     script: "../scripts/transport/road_transport_timeseries.py"
 
 
-use rule create_road_transport_timeseries as create_road_transport_timeseries_historic_electrification with:
-    message: "Create timeseries for historic electrified road transport demand"
+use rule create_uncontrolled_road_transport_timeseries as create_unctronolled_road_transport_timeseries_historic_electrification with:
```
brynpickering, line 87: spelling

rules/transport.smk
```diff
@@ -78,34 +98,34 @@ use rule create_road_transport_timeseries as create_road_transport_timeseries_hi
...
-    message: "Aggregates timeseries for {wildcards.resolution} electrified road transport transport"
+    message: "Aggregates uncontrolled charging timeseries for {wildcards.resolution} electrified road transport transport"
     input:
         time_series = (
-            "build/data/transport/timeseries/timeseries-light-duty-vehicles.csv",
-            "build/data/transport/timeseries/timeseries-heavy-duty-vehicles.csv",
-            "build/data/transport/timeseries/timeseries-coaches-and-buses.csv",
-            "build/data/transport/timeseries/timeseries-passenger-cars.csv",
-            "build/data/transport/timeseries/timeseries-motorcycles.csv"),
+            "build/data/transport/timeseries/timeseries-uncontrolled-light-duty-vehicles.csv",
```
brynpickering, line 108: Turn this into a for loop that uses the keys from the config to decide what needs to be built.

scripts/transport/annual_transport_demand.py
```diff
@@ -229,8 +229,22 @@ def fill_missing_countries_and_years(
...
+        .mul(1 - uncontrolled_share)
+        .to_csv(snakemake.output.road_distance_controlled)
+    )
+    road_distance_uncontrolled = (
+        total_road_distance.rename("value")
+        .mul(uncontrolled_share)
+        .sub(total_historically_electrified_distance.rename("value"), fill_value=0)
+        .to_csv(snakemake.output.road_distance_uncontrolled)
+    )
+    road_distance_historically_electrified = (  # ASSUME historically electrified road consumption is all uncontrolled
```
brynpickering, line 246: put comment in its own line (above)

scripts/transport/road_transport_controlled_charging.py
brynpickering, line 1: I haven't looked, but how close is this to a copy-paste of the `uncontrolled` python file? If it is very similar, we should consider using the two files with a controlled/uncontrolled flag that decides on the functions to use etc.

brynpickering (commented):
A few comments to address

scripts/transport/road_transport_controlled_charging.py
adrienmellot: It's fairly different, because the `uncontrolled` scripts generate timeseries per vehicle type which are then aggregated and rescaled per resolution. This last step is similar but still different because contains different data structures. The `controlled` script is actually closer to the `rescale.py` in the heat sector.

templates/models/techs/demand/electrified-transport.yaml
```diff
@@ -1,29 +1,50 @@
...
+                carrier_con_equals:
+                    electricity: {{ locations.loc[location, year] }} # {{ (1 / scaling_factors.power) | unit("MWh") }}
+    {% endfor %}
+    {% endfor %}
+
 locations:
     {% for id, location in locations.iterrows() %}
-    {{ id }}.techs.demand_road_transport_electrified:
-    {{ id }}.techs.demand_road_transport_historic_electrified:
+    {{ id }}.techs.demand_road_transport_electrified_uncontrolled:
```
brynpickering, line 48: Needs controlled tech here too. I'd maybe update to:

```yaml
{{ id }}.techs:
  demand_road_transport_electrified_uncontrolled:
  demand_road_transport_historic_electrified_uncontrolled:
  demand_road_transport_electrified_controlled:

File: templates/models/techs/demand/electrified-transport.yaml [hunk 8]
```diff
@@ -1,29 +1,50 @@
...
+    demand_road_transport_historic_electrified_uncontrolled:
         essentials:
-            name: 'Removes historic electrified road transport demand'
+            name: 'Removes historic electrified road transport demand -- assumed uncontrolled'
             parent: supply
             carrier: electricity
         constraints:
-            resource: file=demand/road-transport-historic-electrification.csv
+            resource: file=demand/uncontrolled-road-transport-historic-electrification.csv
             resource_min_use: 1
```
brynpickering, line 17: should this be "force_resource"?

brynpickering (changes requested):
Need to also update docs `customisation.md` note `"demand/electrified-transport.yaml"`

templates/models/techs/demand/electrified-transport.yaml
adrienmellot: I've updated it, indeed I think it makes more sense

brynpickering (approved):

PR #339, opened 2024-04-10
Add pre-commit to base dependencies

### LINKED ISSUE #337: Add pre-commit to environment.yaml
### What happened?

EC has a pre-commit hook, but the regular environment.yaml does not include it.

To be safe, let's add it by default.

### Version

1.0.0

### Relevant log output

_No response_

brynpickering:
Fixes #337.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 ### Added (workflow)
 
+* **ADD** Pre-commit to base `euro-calliope` environment dependencies (#339).
 * **ADD** Ruff as our default linter and formatter (#285).
 * **ADD** DAG rule that generates a visualisation of Snakemake's directed acyclic graph (#208).
 * **ADD** IPython debugger to all conda environments to ease debugging (#254).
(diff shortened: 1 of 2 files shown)


timtroendle (approved):
Honestly, I am not sure I would make the pre-commit compulsory and I am also not sure that this is the right way. Given that this isn't a crucial point and you are in favour of it, I am fine merging this.

I wouldn't make it compulsory because it's not strictly necessary and it can be a nuisance. I was considering to remove it again. Given a brief IDE setup, the IDE takes care of 95% of this anyway. Plus, the CI handles all other cases.

I don't think this is the exact right place because it's mixing the roles of the developer and the user of the workflow. The env is for the user, but the change is for the developer. It's also not fail-safe because this only works once in the right env, which developer won't always be in. I, for example, often have a tab in which I run Snakemake and another tab in which I run Git commands, and so this setup wouldn't work in these occasions.

brynpickering:
Closing in favour of updating one's IDE to format on save.

Issue #310, opened 2024-04-08
Update steel industry demand and process it


irm-codebase:
Create rules/workflow to output industry demand at a national level:

**TODO:**

- [x] Get demand for steel industry from JRC-IDEES
- [x]  Replace energy consumption in reducing iron and blast furnaces with pre-computed hydrogen and electricity requirements for direct reduction of iron (DRI)
- [x] update energy consumption for processing steel based on electric arc processes applied to all steel, not only recycled steel

This issue is linked to https://github.com/calliope-project/euro-calliope/issues/102

irm-codebase:
Closed by #340

Issue #311, opened 2024-04-08
Update chemical industry demand


irm-codebase:
Update chemical industry demand
a. Get demand from JRC-IDEES
b. calculate quantity of ethylene, propylene, BTX, ammonia, and methanol (i.e. base chemicals) needed by country.
c. calculate hydrogen, CO2 and electricity demands to produce these base chemicals, based on molar masses

This issue is linked to https://github.com/calliope-project/euro-calliope/issues/102

Issue #309, opened 2024-04-08
Get demand for "other" industry and process it


irm-codebase:
Create rules/workflow to output demands for all industries without individual processing at a national level:

**TODO:**

- [ ] Identify which industry sectors have no specific processing steps.
- [ ]  Get demands for key final energy carriers (methane, diesel, electricity) and end uses (space heat).
- [ ] Produce the final file by combining the demands above 

This issue is linked to https://github.com/calliope-project/euro-calliope/issues/102

PR #340, opened 2024-04-10
Add industry module

### LINKED ISSUE #308: Define and create interfaces for the integration of an industry sub-module
It was agreed that we'd use the industry as an example of how to modularize the code into manageable bits.

Within snakemake there are two possible approaches:

- [wrappers](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#wrappers): this method allows us to "port" code from other repositories directly, meaning that the module itself does not live in the same git repo as Euro-Calliope.
- [modules](https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html#modules): this method calls another snakemake workflow that lives within the same repository. It's already used in SCEC to call EC rules.

Below you can find an initial analysis of the industry functionality within SCEC:

Image

irm-codebase:
Fixes #308, #310, #347, #345 and #346.

This pull request includes the "skeleton" of the industry module, and includes the steel industry sub-workflow as an example.

## Checklist

- [x] CHANGELOG updated
- [x] Steel rule done


modules/industry/config.yaml
```diff
@@ -0,0 +1,10 @@
+industry:
+    inputs:
+        path-energy-balances: build/data/annual-energy-balances.csv
+        path-cat-names: config/energy-balances/energy-balance-category-names.csv
+        path-carrier-names: config/energy-balances/energy-balance-carrier-names.csv
+    outputs:
+    params:
+        year-range: [2000, 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010, 2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018]
```
brynpickering, line 8: Just do start and end point and then use `slice(*year_range)` in your script. This should possibly be in the config under `pre-processing` as it also affects other eurostat/ch/jrc data.

modules/industry/industry.smk
```diff
@@ -0,0 +1,57 @@
+SRC_PATH = "src"
+TMP_PATH = "modules/industry/tmp"
```
brynpickering, line 2: `MODULE_PATH = "modules/industry"` and then all the others are  `MODULE_PATH + "..."`

modules/industry/industry.smk
```diff
@@ -0,0 +1,57 @@
...
+    params:
+        year_range = config["params"]["year-range"]
+    input:
+        path_energy_balances = config["inputs"]["path-energy-balances"],
+        path_cat_names = config["inputs"]["path-cat-names"],
+        path_carrier_names = config["inputs"]["path-carrier-names"],
+        path_jrc_energy = f"{DATA_PATH}/jrc_idees_processed_energy.csv.gz",
+        path_jrc_production = f"{DATA_PATH}/jrc_idees_processed_production.csv.gz",
+    output:
+        path_output = f"{TMP_PATH}/annual_demand_steel.csv"
```
brynpickering, line 25: I think Tim has now changed the naming convention from "demand" to "final_energy".

modules/industry/raw_data/jrc_idees_processed_energy.csv.gz
brynpickering, line 1: This should no longer be needed. We can now grab JRC data directly from a file server. See transport and heat demand modules.

modules/industry/src/steel_industry.py
```diff
@@ -0,0 +1,225 @@
+import eurocalliopelib.utils as ec_utils
+import pandas as pd
+from utils import formatting
+from utils import jrc_idees_parser as jrc
+
+CAT_NAME_STEEL = "Iron and steel"
+
+H2_LHV_KTOE = 2.863  # 0.0333 TWh/kt LHV -> 2.863ktoe/kt
```
brynpickering, line 8: Move these to the config

modules/industry/src/steel_industry.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def get_steel_demand_df(
+    year_range: list,
+    path_energy_balances: str,
+    path_cat_names: str,
+    path_carrier_names: str,
+    path_jrc_energy: str,
+    path_jrc_production: str,
+    path_output: str = "",
```
brynpickering, line 21: `Optional[str] = None` and then use `if path_output is not None:`.

modules/industry/src/utils/formatting.py
```diff
@@ -0,0 +1,137 @@
+import eurocalliopelib.utils as ec_utils
+import numpy as np
+import pandas as pd
+import utils.jrc_idees_parser as jrc
+
+LEVEL_ORDER = ["subsector", "country_code", "unit", "carrier", "year"]
```
brynpickering, line 6: move to config

brynpickering (commented):
Some comments. Will need @timtroendle to comment on the utils files. We have `eurocalliopelib` at the moment to store utility functions. Should these go there? Or does the modularisation require a different approach?

modules/industry/config.yaml
irm-codebase: Agreed. I had to use `range(*year_range)`, but it works well.
One issue here is that, by design, we've restricted access in the module to generic configuration.

For now I've kept it at the module level. Correcting this for all the configuration should be a separate issue.

modules/industry/src/steel_industry.py
irm-codebase: Config is meant to hold things that are exclusively user-relevant.

In this case, `CAT_NAME_STEEL` is predetermined by JRC and should not really be changed by the user because it can only lead to errors within `steel_industry.py`. Other sectors will use a similar structure.

Maybe this does apply for `H2_LHV_KTOE`. I'll move it to the config.

modules/industry/src/utils/formatting.py
irm-codebase: See my comments above regarding non-user specific configuration. This is only relevant to formatting, any other configuration will result in broken functionality.

modules/industry/config.yaml
brynpickering: Careful here: range does not include the final value by default.

modules/industry/src/steel_industry.py
brynpickering: recycled steel percentage is also an assumption / based on some (now unknown) data source, as is HDRI consumption

modules/industry/config.yaml
brynpickering, line 1: You should have a schema for this config.

modules/industry/raw_data/jrc_idees_processed_energy.csv.gz
irm-codebase: We've created issue #345 to tackle this. Merging the basic structure of the industry module is a blocker for issues #311 and #309.

modules/industry/config.yaml
irm-codebase: We've created issue #347 to tackle this.

A schema is a great idea, but the module's configuration might change too much while we port features. We'd prefer to do it once the chemical and "other" sectors are ported, to ensure the configuration file is mature enough.

modules/industry/config.yaml
irm-codebase: Fixed!

modules/industry/src/steel_industry.py
irm-codebase: We've moved only the recycled share of scrap steel to the cnf, for now. The others can also be moved later if we find it to be useful :)

modules/industry/src/steel_industry.py
```diff
@@ -8,13 +8,17 @@
 CAT_NAME_STEEL = "Iron and steel"
 
 H2_LHV_KTOE = 2.863  # 0.0333 TWh/kt LHV -> 2.863ktoe/kt
-RECYCLED_STEEL = 0.5  # 50% H-DRI Iron, 50% scrap steel
-H2_TO_STEEL = (1 - RECYCLED_STEEL) * 0.05  # 0.05t_h2/t_steel in H-DRI
 HDRI_CONSUMPTION = 0.0116  # H-DRI: 135kWh_e/t = 0.0116ktoe/kt
 
 
+def _get_h2_to_steel(recycled_steel_share: float) -> float:
+    """Get t_h2/t_steel, usually for H-DRI."""
```
brynpickering, line 15:
```suggestion
    """Get t_h2/t_steel, usually for H-DRI. Conversion factor of 0.05 t_h2/t_steel is assumed"""
```

modules/industry/src/steel_industry.py
```diff
@@ -8,13 +8,17 @@
...
+    return (1 - recycled_steel_share) * 0.05
```
brynpickering, line 16:
```suggestion
    # ASSUME: conversion factor of 0.05 t_h2/t_steel.
    return (1 - recycled_steel_share) * 0.05
```

timtroendle (changes requested):
That's an excited pull request, thanks!

In my review, I mainly focussed on the structure assuming @brynpickering has checked the data processing logic. There are a few consistency related issues that need addressing. 

In addition, this change is not adding a feature and it's not show-casing the entire integration and no tests. Would it be possible to refactor the change so it's a feature? What I mean is: can we build this into the model? I understand you derive annual steal demand, so from there you could quite easily create flat steel demand profiles for all locations and put those into the model. Only then would we see the entire integration pipeline: model feature, documentation, and tests.

irm-codebase:
> @irm-codebase and @tud-mchen6 maybe we can discuss the design briefly after the group meeting today? @brynpickering may be useful to include in the discussion.

That would be good! No problem on my side :+1:

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.2.0 (unpublished)
 
+### Added (models)
+
+* **ADD** industry module and steel industry energy demand processing. NOT CONNECTED TO THE MAIN WORKFLOW. Industry sectors pending: chemical, "other".
```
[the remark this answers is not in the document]
irm-codebase: The industry data pipeline has a lot of steps. Briefly:

1. Process industry sub-sectors (steel/chemicals/other...)
2. Merge subsector demands and test against dramatic increase/decrease in overall demand.
3. Scale demands based on a configuration .csv.
4. Create demand profiles (flat)

Adding everything in one go is too much, so we are working top down to keep things manageable. The "connection" to the main workflow will happen at around step 5, but first we need all the previous steps.

modules/industry/config.yaml
```diff
@@ -0,0 +1,10 @@
+industry:
+    inputs:
+        path-energy-balances: build/data/annual-energy-balances.csv
+        path-cat-names: config/energy-balances/energy-balance-category-names.csv
+        path-carrier-names: config/energy-balances/energy-balance-carrier-names.csv
+    outputs:
```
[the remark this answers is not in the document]
irm-codebase: This are quite necessary, in our mind: the general goal of the modules is to de-couple them from the main workflow. If/when we convert the module to a wrapper (which lives in a separate repository), it'll be important to make sure that the main workflow/code is agnostic to locations in the Euro-Calliope workflow.

To ensure this, we decided to add inputs in the config file.

modules/industry/config.yaml
```diff
@@ -0,0 +1,10 @@
+industry:
+    inputs:
```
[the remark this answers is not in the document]
irm-codebase: Hmmm... the purpose of the input section in `config.yaml` is to specify files that come from the main workflow.

How about `module-inputs`, to ensure clarity?

modules/industry/README.md
```diff
@@ -0,0 +1,3 @@
+# About
+
+Placeholder text for the industry module documentation.
```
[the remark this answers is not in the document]
irm-codebase: We've added a brief description in the `README.md` file.

This will be completed in conjunction with the `schema` file as we add more steps in the processing pipeline.

modules/industry/config.yaml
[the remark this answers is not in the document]
irm-codebase: This is the setup we are following:

**Inputs**: files produced by rules outside of our module.
**Outputs**: files produced by the module that are relevant to the main workflow (intermediary steps go into the /tmp folder).
**Configuration**: a .yaml file that gives users the ability to modify module behaviour.
**Module**: code and steps for industry. Ideally, users shouldn't need to mess with it unless something needs to be added.

module_setup

modules/industry/env_industry.yaml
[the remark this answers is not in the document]
irm-codebase: The end-goal is the opposite, actually: to let users specify their input files (some predetermined, others optional), and separate the module entirely.

We'd like to keep this env file to ensure containerization in case we decide to go a step further and use wrappers.

modules/industry/config.yaml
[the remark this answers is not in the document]
irm-codebase: We'll keep the current approach (for now).
You can find explanations in the schema.yaml of the module :+1:

modules/industry/industry.smk
```diff
@@ -0,0 +1,60 @@
...
+CONDA_PATH = "./env_industry.yaml"
+
+# TODO:
+# jrc_idees_processed* files in the raw_data folder should ideally be produced by a rule instead
+
+# Ensure rules are defined in order.
+# Otherwise commands like "rules.rulename.output" won't work!
+rule steel_industry:
+    message: "Calculate energy demand for the 'Iron and steel' sector in JRC-IDEES."
+    conda: CONDA_PATH
```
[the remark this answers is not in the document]
irm-codebase: The module has to be run from the base environment due to Snakemake quirks, so we only need one env file.
It does include snakemake (see `env_industry.yaml`). You should not need another one...

modules/industry/src/utils/formatting.py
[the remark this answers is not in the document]
irm-codebase: There is both jrcidees parsing and regular formatting in this folder. We'll keep the name utils to avoid having 3 folder layers.

modules/industry/src/steel_industry.py
```diff
@@ -0,0 +1,227 @@
+from typing import Optional
+
+import eurocalliopelib.utils as ec_utils
+import pandas as pd
+from utils import formatting
```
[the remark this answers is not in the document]
irm-codebase: I think this is a strong argument to use wrappers instead...

Moving these utils to another location (outside the module) sort of defeats the purpose of modularisation in the first place...

modules/industry/industry.smk
```diff
@@ -0,0 +1,60 @@
+# Paths dependent on main Snakefile
+MODULE_PATH = "modules/industry"
+TMP_PATH = f"{MODULE_PATH}/tmp"
+OUT_PATH = f"{MODULE_PATH}/out"
+DATA_PATH = f"{MODULE_PATH}/raw_data"
```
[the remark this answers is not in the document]
irm-codebase: The JRC step that @brynpickering will effectively remove the raw_data folder. We have changed out to build, though!

modules/industry/config.yaml
irm-codebase: > This config must go into the metadata that we create for model builds. See the `build_metadata` rule.

The general `config` already includes our module's configuration. Does this resolve your issue?

```
# Include modules
configfile: "modules/industry/config.yaml"
```

irm-codebase:
Relatively big update to this PR to fix #346. 

This was necessary because #354 made the pandas code go significantly slower. Moving to `xarray` made this module significantly faster. All results have been tested using checksums, and I can confirm the results are the same.

brynpickering:
@irm-codebase if you can rebase this onto current `develop` then I'll review

irm-codebase:
@brynpickering updated to the newest JRC version.

I had to modify a couple of lines at the JRC level because the production dataset did not have names for the variables. Hope that's ok.

modules/industry/industry.smk
```diff
@@ -0,0 +1,59 @@
...
+MODULE_PATH = "modules/industry"
+BUILD_PATH = f"{MODULE_PATH}/build"
+DATA_PATH = f"{MODULE_PATH}/raw_data"
+
+# Paths relative to this snakefile (snakemake behaviour is inconsitent)
+SCRIPT_PATH = "scripts"  # scripts are called relative to this file
+CONDA_PATH = "./env_industry.yaml"
+
+# TODO:
+# jrc_idees_processed* files in the raw_data folder should ideally be produced by a rule instead
```
brynpickering, line 11: fixed?

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+        xr.Dataset: dataframe with steel demand per country.
+    """
+    # Load data
+    energy_balances_df = pd.read_csv(
+        path_energy_balances, index_col=[0, 1, 2, 3, 4]
+    ).squeeze("columns")
+    cat_names_df = pd.read_csv(path_cat_names, header=0, index_col=0)
+    carrier_names_df = pd.read_csv(path_carrier_names, header=0, index_col=0)
+    jrc_energy = xr.open_dataset(path_jrc_industry_energy)
+    jrc_prod = xr.open_dataset(path_jrc_industry_production)
```
brynpickering, line 50: `open_dataarray` and then you don't need production to have a data variable

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+
+
+def _get_h2_to_steel(recycled_steel_share: float) -> float:
+    """Get t_h2/t_steel, usually for H-DRI."""
+    # ASSUME: conversion factor of 0.05 t_h2/t_steel.
+    return (1 - recycled_steel_share) * 0.05
+
+
+def get_steel_demand_df(
+    cnf_steel: dict,
```
brynpickering, line 21: minor point, but could you call this `config_steel`? `cnf` is not an abbreviation we use anywhere else

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+    auxiliary_intensity = jrc.get_auxiliary_electric_final_intensity(
+        "Electric arc", "Electric arc", jrc_energy, jrc_prod, fill_empty=True
+    )
+
+    # Total electricity consumption:
+    #   If the country produces steel from Iron ore (smelting):
+    #   sintering/pelletizing * iron_ore_% + smelting * recycled_steel_% + H-DRI + EAF + refining/rolling + finishing + auxiliaries
+    #   If the country only recycles steel:
+    #   smelting + EAF + refining/rolling + finishing + auxiliaries
+    recycled_share = cnf_steel["recycled-steel-share"]
```
brynpickering, line 155: this is the only thing extracted from the steel config. I would pass it thorugh as a direct parameter rather than wrapped in a dict. It is generally a good idea to split up config items in snakemake as it then tracks those specific config values for changes rather than the whole dictionary. I know you don't expect to need steel config params outside this script, but you never know!

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+    # only for country/year that handle iron ore and don't recycle all their steel
+    h_dri_h2_intensity = H2_LHV_KTOE * _get_h2_to_steel(recycled_share)
+
+    h2_intensity = electric_intensity.where(sintering_intensity > 0).fillna(0)
+    h2_intensity = h2_intensity.where(h2_intensity == 0, h_dri_h2_intensity)
+    h2_intensity = h2_intensity.assign_coords(carrier_name="hydrogen")
+
+    # Low heat
+    low_heat_intensity = jrc.get_subsection_useful_intensity(
+        "Electric arc", "Low enthalpy heat", "Electric arc", jrc_energy, jrc_prod
```
brynpickering, line 195: Is it only electric arc that has low enthalpy heat?

brynpickering: Ah ok, it's because we're assuming future space heating will be based on the electric arc intensity. Might be worth mentioning that in a comment.

brynpickering: Ignore this, it's in the main docstring.

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # Prettify
+    useful_intensity = ensure_standard_coordinates(useful_intensity)
+    useful_intensity = useful_intensity.assign_attrs(units="twh/kt")
+    useful_intensity.name = "useful_intensity"
+
+    return useful_intensity.fillna(0)
+
+
+# TODO: fix me!
```
brynpickering, line 95: Should this still be here?

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+    jrc_prod = xr.open_dataset(path_jrc_industry_production)
+
+    # Ensure dataframes only have data specific to this industry
+    cat_names_df = cat_names_df[cat_names_df["jrc_idees"] == CAT_NAME_STEEL]
+    jrc_energy = jrc_energy.sel(cat_name=CAT_NAME_STEEL)
+    jrc_prod = jrc_prod.sel(cat_name=CAT_NAME_STEEL)
+
+    # Process data
+    new_steel_demand = transform_jrc_subsector_demand(jrc_energy, jrc_prod, cnf_steel)
+    new_steel_demand = formatting.fill_missing_countries_years(
```
brynpickering, line 59: Is this the approach I took? I thought I filled in the _intensities_ based on neighbours and ffill/bfill and _steel production_ based on annual energy production, then combined those into a new electricity demand. I wonder what the difference in result would be.

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -0,0 +1,109 @@
...
+) -> xr.DataArray:
+    """Get final energy intensity of a given JRC section/subsection/material."""
+    # Extract relevant section and subsection data.
+    final_demand = jrc_energy["final"].sel(section=section, subsection=subsection)
+    useful_demand = jrc_energy["useful"].sel(section=section, subsection=subsection)
+    production = jrc_prod["demand"].sel(produced_material=material)
+
+    total_eff = useful_demand / final_demand
+    carrier_eff = total_eff.where(total_eff > 0).sel(carrier_name=carrier_name)
+    if fill_empty:
```
brynpickering, line 55: Aha, this is what I probably meant by filling empty data at the intensity level.

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -0,0 +1,109 @@
...
+    final_intensity = ensure_standard_coordinates(final_intensity)
+    final_intensity = final_intensity.assign_attrs(units="twh/kt")
+    final_intensity.name = "final_intensity"
+
+    assert final_intensity.sum() >= useful_intensity.sum(), "Creating energy!"
+
+    return final_intensity.fillna(0)
+
+
+def get_subsection_useful_intensity(
```
brynpickering, line 75: Is there a useful assertion you can add here?

scripts/jrc-idees/industry.py
```diff
@@ -26,59 +34,104 @@
...
+            cat_name=df.index.name.split(": ")[1],
+        )
+        .rename_axis(index="produced_material")
+    )
+    df_processed.index = (
+        df_processed.index.str.replace("(kt)", "", regex=False)
+        .str.replace("(kt ", "(", regex=False)
+        .str.strip()
+    )
+    df_processed = df_processed.assign(variable="demand")
```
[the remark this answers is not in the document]
brynpickering: Now that I see your changes, I guess having a "variable" column that is transformed into the dataset variables is fine. I would not bother having as an input arg to `df_to_xr`, just take it as given that it is the col `"variable"`. I'd also call this one "production" rather than "demand".

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -0,0 +1,109 @@
...
+import xarray as xr
+
+STANDARD_COORDS = ["cat_name", "year", "country_code", "carrier_name"]
+
+
+def ensure_standard_coordinates(ds: Union[xr.Dataset, xr.DataArray]):
+    """Remove all coordinates that do not match a predefined standard."""
+    removed_coords = [i for i in ds.coords if i not in STANDARD_COORDS]
+    removed_dims = [i for i in ds.dims if i in removed_coords]
+    if any(removed_dims):
```
brynpickering, line 12:
```suggestion
    removed_coords = set(ds.coords).difference(STANDARD_COORDS)
    removed_dims = removed_coords.intersection(ds.dims)
    if removed_dims: 
```
an empty set will resolve to False, otherwise True.

Is this an expected problem, that there would be dimensions with coordinates? Otherwise I'd say that the first two lines could just be `set(ds.dims).difference(STANDARD_COORDS)`

modules/industry/scripts/utils/formatting.py
```diff
@@ -0,0 +1,122 @@
...
+) -> xr.DataArray:
+    """
+    For countries without relevant data in JRC_IDEES we use their
+    Eurostat energy balance data to estimate future energy consumption, relative to the
+    energy balance data of the 28 countries for which we do have JRC IDEES data.
+    JRC data is available until 2015, so for future years data for all countries is filled
+    based on Eurostat energy balances.
+    Any other missing years are filled in based on average consumption of a country.
+    """
+    # Ensure countries follow our standard alpha3 convention.
```
brynpickering, line 22: They do in the output of the JRC processing, so no need to do that again here.

modules/industry/scripts/utils/formatting.py
```diff
@@ -0,0 +1,122 @@
...
+    _to_fill = _to_fill.interpolate_na(
+        dim="year", use_coordinate="year", method="linear"
+    )
+    _to_fill = _to_fill.bfill(dim="year")
+    all_filled = _to_fill.ffill(dim="year")
+
+    all_filled = jrc.ensure_standard_coordinates(all_filled)
+    all_filled = all_filled.assign_attrs(units="twh")
+
+    return all_filled
```
brynpickering, line 103: Once you have done all the filling, you could check that there is no case of conumption in a year being outside a range of the average. E.g. you could catch cases of zero consumption in a year when the other years for that country have non-zero data. Maybe doesn't work because some countries really didn't have any steel production in e.g. 2000, but just a thought of how to confirm that gap-filling has succeeded.

modules/industry/scripts/steel_industry.py
```diff
@@ -0,0 +1,221 @@
...
+HDRI_CONSUMPTION = 135e-6  # H-DRI: 135kWh_e/t
+
+
+def _get_h2_to_steel(recycled_steel_share: float) -> float:
+    """Get t_h2/t_steel, usually for H-DRI."""
+    # ASSUME: conversion factor of 0.05 t_h2/t_steel.
+    return (1 - recycled_steel_share) * 0.05
+
+
+def get_steel_demand_df(
```
brynpickering, line 20: you're now getting an `nc`/`dataset`, not `df`

modules/industry/scripts/utils/formatting.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+def to_calliope_style(subsector_demand: xr.Dataset) -> pd.DataFrame:
```
brynpickering, line 106: Is this used any more?

brynpickering (changes requested):
Great, we're close to having this ready to merge in!

scripts/jrc-idees/industry.py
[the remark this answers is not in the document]
irm-codebase: Fixed!
And yeah, I think that specifying the variable names is better in case anyone starts processing these things in a different way.

modules/industry/README.md
```diff
@@ -0,0 +1,17 @@
+# About
+
+This module is dedicated to getting and processing energy demand for industry.
+
+Basic info of this module:
+- Main data sources: JRC IDEES and eurostat
+- Spacial resolution: national (because of JRC IDEES)
+- Temporal resolution: annual aggregated
+
+Some industries are specified as 'sub-module' in this module, such as iron and steel industry. This allows flexibly configure the decarbonisation level of each industry separately. All other industries are grouped into 'other industies'.
```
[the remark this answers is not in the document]
irm-codebase: Noted, I fixed the spacing. I copy pasted your fix and added a small extra correction (industies -> industries).

modules/industry/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+For non-developer users, `config.yaml` is the main file to look into. Here, one can specify parameters such as the years to be included in the data processing, assumptions such as the share of recycled steel, and what specific industry sector to be included (such as iron and steel industry). The structure of the `config.yaml` file is:
```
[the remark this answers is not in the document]
irm-codebase: Ditto. Added the spacing.

modules/industry/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+        path-energy-balances: build/data/annual-energy-balances.csv
+        path-cat-names: config/energy-balances/energy-balance-category-names.csv
+        path-carrier-names: config/energy-balances/energy-balance-carrier-names.csv
+        path-jrc-industry-energy: build/data/jrc-idees/industry/processed-energy.nc
+        path-jrc-industry-production: build/data/jrc-idees/industry/processed-production.nc
+    outputs:
```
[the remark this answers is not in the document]
irm-codebase: Unfortunately yes, for now. If removed, `outputs` becomes a None item, and breaks the schema. These will go away once the full module workflow is complete.

modules/industry/scripts/steel_industry.py
irm-codebase: This is to keep the configuration of sectors consistent: chemical, other and any extras are meant to have their own section in the configuration. Passing the whole dictionary keeps the snakemake leaner and is easier to work with on the python side too.

Besides, if another script needs a specific parameter from industry, they can just grab it directly instead of taking the whole dictionary.

irm-codebase: I'll keep this as-is, for now.

modules/industry/scripts/steel_industry.py
irm-codebase: Seems like it was! We checked the results against those in SCEC during development, and it's how it was done. Here's a brief excerpt:

```pthon
all_consumption = pd.concat([
        steel_energy_consumption.reorder_levels(all_other_industry_consumption.index.names),
        chem_energy_consumption.reorder_levels(all_other_industry_consumption.index.names),
        all_other_industry_consumption,
        all_other_industry_space_heat
    ])
    all_consumption.columns = all_consumption.columns.astype(int).rename('year')
    all_filled_consumption = fill_missing_data(energy_balances, cat_names, carrier_names, all_consumption)
    units = all_filled_consumption.index.get_level_values('unit')
    all_filled_consumption.loc[units == 'ktoe'] = (
        all_filled_consumption.loc[units == 'ktoe'].apply(util.ktoe_to_twh)
```

`all_filled_consumption` already has the steel stuff, and `fill_missing_data` is the `fill_missing_countries_years` bit here.

It would be good to introduce an issue with improvements for industry, to make things better once the whole pipeline is in. There are a couple of things in the "other" processing that could improve, too.

modules/industry/scripts/utils/jrc_idees_parser.py
irm-codebase: Do you mean the `assert` at line 70?

modules/industry/scripts/utils/jrc_idees_parser.py
irm-codebase: No problems, then?

modules/industry/scripts/utils/jrc_idees_parser.py
irm-codebase: The "other" industry merge corrects it, and I do not want to deal with the merge conflict... so yeah.

modules/industry/scripts/utils/jrc_idees_parser.py
irm-codebase: Oh this is much more efficient! Thanks.

modules/industry/scripts/utils/formatting.py
irm-codebase: removed

modules/industry/scripts/utils/jrc_idees_parser.py
brynpickering: I mean one in `get_subsection_useful_intensity`

irm-codebase: I added a check for divide-by-zero. It's the only generic test that made sense to me here.

modules/industry/scripts/utils/formatting.py
irm-codebase: I tried a couple of approaches, but I could not land on something generic enough that justifies the addition.
If I go for averages, countries that introduced production fail (as you mention).

I think a more robust approach would be to introduce a testing stage after these files are generated, but before calliope translation. Checking if the rate of increase/decrease of the carriers in each industry category are within certain ranges would be the most universal approach, since in theory it would sideline this problem.

Skipped this for now.

irm-codebase:
@brynpickering added most of your suggestions, with the exception of the gap filling test.

Let me know if you find anything else during review. Otherwise, I'll merge this in and update the "chemical" and "other" cases.

brynpickering:
One thing that is possibly missing is documentation in the main euro-calliope docs about these new modules. But for now I think we can ignore that as we may well move this to a different repo before the next release...

timtroendle:
I would add documentation now. It's _now_ in this repo and we may or may not move it somewhere else in the future. Like the model, the documentation is quite modular anyway, isn't it? So we could easily move the industry bit out of the documentation, too, should we decide to pull out this module into its own repo. We may as well decide to keep the documentation in the main repo and simply mention the optional module (one aspect we haven't discussed at all so far).

irm-codebase:
@timtroendle I'd say that adding full documentation now is only additional overhead: this is only part of the module (we are still missing other industry and chemical industry), plus disaggregation, testing and scaling steps. Also, we have already added pre-eliminary documentation explaining usage (see the README.md).

Besides, I agree with Bryn: moving module subworkflows into a separate repo makes a lot of sense. Still, I created issue [#381](https://github.com/calliope-project/euro-calliope/issues/381) to make sure we stay on point with this.

For now, let's finish this PR once and for all, please 🙏🙏🙏?

timtroendle:
Yeah fair point. Sorry, didn't want to be annoying, it's just easy to forget these things.

envs/default.yaml
```diff
@@ -13,7 +13,8 @@ dependencies:
     - pycountry=18.12.8
     - jinja2=2.11.3
     - pip=21.0.1
-    - xarray=0.17.0
+    - xarray=2022.9.0
```
brynpickering, line 16: @irm-codebase  This is causing a merge conflict. Do you actually need the default env to be update if you have modularised industry?

irm-codebase: @brynpickering nope. Apologies, that's a leftover from were the module needed to update the country names. Since that's handled on the JRC part, we no longer have version conflicts there.

Fixed!

PR #354, opened 2024-04-11
Add industry JRC data processing

### LINKED ISSUE #345: Industry: add JRC processing step to remove raw_data files
Currently, industry rules need jrc files to be present in the `/industry/raw_data` folder.

Other sectors (heat, transport) have steps to create their own JRC data files. Industry should do the same.

TODO:

- [ ] Port extraction steps from SCEC
- [ ] Create files for processed_energy and processed_production in `module/tmp`

brynpickering:
Fixes #345. 

Workflow checks haven't completed on my device.

This starts the process of modularising the JRC-IDEES data processing pipeline, which I will later apply to transport and heat (and replace the existing processing scripts).

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


config/default.yaml
```diff
@@ -40,6 +40,7 @@ data-pre-processing:
             SRB: [HUN]
             CYP: [ROU]
 root-directory: .
+max-threads: 2
```
irm-codebase, line 43: Why do we need to constrain the run to just 2 threads? Could you add a comment?

irm-codebase, line 43: Is there a reason to cap at 2 threads? Could you add a comment?

rules/jrc-idees.smk
```diff
@@ -0,0 +1,47 @@
+"Rules regarding JRC-IDEES Data"
+
+JRC_IDEES_SPATIAL_SCOPE = [
```
irm-codebase, line 3: **Nice-to-have:**
Should we consider moving this to the configuration?
As it is, data will always be downloaded for all JRC countries even if it is unused (such as in the minimal configuration).

irm-codebase (approved):
Couple of nice to haves. Nothing too important.

config/default.yaml
brynpickering: Best to be conservative as you never know how limited someone's devices will be. There's no real reason for 2 over 1 or 3 or 4. @timtroendle does 2 seem reasonable?

rules/jrc-idees.smk
brynpickering: No, you need data from some countries to fill in neighbours, so even if they're not in the list of model countries we need to pull in and process all data

timtroendle:
I started my review and I am trying to finish it today but It may take a while.

config/default.yaml
irm-codebase: In that case, could you mention this limit in the configuration? We mostly develop using clusters with 24 + cores available. This might slow down development if people are not aware of it.

Mentioning the possibility of overriding this might be useful to people.

brynpickering:
I have a few local changes to commit that address the file unzipping (and downloading) to limit to just the defined countries and their infill neighbours (if necessary). Will push that later this evening.

timtroendle:
Ok, I will finish the review tomorrow in that case.

rules/jrc-idees.smk
brynpickering: I've now combined main countries and infill countries into one list and only unzip and process that list of countries. It's a bit of a proof of concept and might be too verbose to be worth keeping

brynpickering:
@timtroendle ready for review now. 

I'm aware that `get_countries_to_unzip` is quite verbose. It would be less so if we could access eurocalliopelib from the main snakemake env (and so use the country code conversion utils). Anyway, I've done it as a bit of a proof-of-concept to show how we _could_ filter the JRC files to process to just those that are relevant to the configured countries (and their infill neighbours, where necessary). I would be fine to revert this to processing all countries and then having the filtering happening much further down the line.

lib/eurocalliopelib/utils.py
```diff
@@ -16,10 +18,10 @@ def eu_country_code_to_iso3(eu_country_code):
 
 def convert_country_code(input_country, output="alpha3"):
     """
-    Converts input country code or name into either a 2- or 3-letter code.
+    Converts input country code or name into either either a 2- or 3-letter code.
 
     ISO alpha2: alpha2
-    ISO alpha2 with Eurostat codes: alpha2_eu
```
brynpickering, line 22: Codes are used by eurostat _and_ JRC-IDEES

lib/eurocalliopelib/utils.py
```diff
@@ -49,12 +51,101 @@ def convert_country_code(input_country, output="alpha3"):
...
+            mapped_codes[country_code] = convert_country_code(
+                country_code, output=output
+            )
+        except LookupError:
+            print(f"Skipping country/region {country_code} in annual energy balances")
+            continue
+    return mapped_codes
+
+
+def rename_and_groupby(
```
brynpickering, line 81: Quite verbose but will be used by other scripts that I have to merge in at a later date.

rules/jrc-idees.smk
```diff
@@ -0,0 +1,76 @@
...
+        )
+    params:
+        sector_title_case = lambda wildcards: wildcards.sector.title(),
+        # These params are referenced so that the rule re-runs when they change value.
+        # They are used in the `get_countries_to_unzip` helper function.
+        infill_country_config = config["data-pre-processing"]["fill-missing-values"]["jrc-idees"],
+        countries = config["scope"]["spatial"]["countries"]
+    wildcard_constraints:
+        sector = "((industry)|(transport)|(tertiary))"
+    output: temp(directory("build/data/jrc-idees/{sector}/unprocessed"))
```
brynpickering, line 57: I'm outputting to a folder rather than wildcarded individual files because it makes other references to this data lighter (heat, transport, etc. only need to reference the directory) _and_ it seems to make sense to me to do all the unzipping in one go into a temporary directory as it is a very quick step and all the files then get deleted as soon as the downstream processing is complete. The rule is much simpler when not trying to filter countries (see earlier commits)

scripts/heat/jrc_idees.py
brynpickering, line 1: This file will move/change eventually. I have a script ready to process tertiary sector data that is comparable to the industry one in this PR

scripts/transport/jrc_idees.py
brynpickering, line 1: This file will move/change eventually. I have a script ready to process transport sector data that is comparable to the industry one in this PR

rules/jrc-idees.smk
irm-codebase: Thank you!
And agreed, I suppose it's fine to just do all countries unless the files are huge if this does not work well with the rest of the workflow.

rules/transport.smk
```diff
@@ -14,10 +14,7 @@ rule download_transport_timeseries:
 rule jrc_idees_transport_processed:
     message: "Process {wildcards.dataset} transport data from JRC-IDEES to be used in understanding current and future transport demand"
     input:
-        data = expand(
-            "build/data/jrc-idees/transport/unprocessed/{country_code}.xlsx",
-            country_code=JRC_IDEES_SCOPE
-        )
+        data = "build/data/jrc-idees/transport/unprocessed"
```
irm-codebase, line 17: Question: will this break snakemake operations?
Passing just a folder might break operations since it won't detect changes in the files when building the DAG.
If this is due to my suggestion to make the countries extracted flexible, then it was not a good one...

Another option is to add a function in the smk that build an array with the filenames in this folder, and specify rule order...

irm-codebase (commented):
Just a few comments in relation to SMK DAG generation.

rules/transport.smk
brynpickering: The directory is temporary - it is deleted once all rules that rely on it are completed (in the case of JRC-IDEES data, there is only one downstream rule per unzipped directory). It's not really related to keeping all countries extracted flexible. It's more about limiting the number of times you have to reference that array of filenames from different rules.

brynpickering: This means that whenever you build the DAG again, if it needs to re-run a rule that relies on the directory then unzipping files into the directory will need to be re-run (since those files have been deleted). Since the unzipping is a light operation, this keeps us from having lots of unnecessary files lying around (you can just get them again whenever you want by unzipping the downloaded data)

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,202 @@
+from itertools import product
+from multiprocessing import Pool
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+from eurocalliopelib import utils
```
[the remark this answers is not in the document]
irm-codebase: @sjpfenninger, @tud-mchen6 and I tried in linux, but unfortunately updating `xarray` in the default environment makes other parts of the code stop running.

For some reason, the fix in #357 stops working for the JRC steps... We think the solution is to introduce a new environment for the JRC processing section.

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+        )
+        for dim_item, da_data in expected.items():
+            np.testing.assert_array_equal(new_da.sel(foo=dim_item), da_data)
+
+    @pytest.mark.parametrize(
+        ("rename_dict", "expected"),
+        [
+            (
+                {"A": "A1", "B": "B1", "C": "C1", "D": "D1", "E": "E1"},
+                {"A1": [1, 5], "B1": [np.nan, 4], "D1": [4, np.nan], "E1": [5, 1]},
```
timtroendle, line 95: Hm this expected behaviour surprises me. So if `dropna=True` you do not sum?

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+        for dim_item, da_data in expected.items():
+            np.testing.assert_array_equal(new_da.sel(foo=dim_item), da_data)
+
+    @pytest.mark.parametrize(
+        ("rename_dict", "expected"),
+        [
+            (
+                {"A": "A1", "B": "B1", "C": "C1", "D": "D1", "E": "E1"},
+                {
+                    "A1": [1, 5],
```
timtroendle, line 123: This expected behaviour also surprises me. Why is this not summed? And where does the `5` come from?

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+        ],
+    )
+    @pytest.mark.parametrize("da", ["single_index", "multi_index"], indirect=True)
+    def test_rename_groupby_keep_renamed(self, rename_dict, expected, da):
+        new_da = rename_and_groupby(
+            da, rename_dict, keep_non_renamed=True, dim_name="foo"
+        )
+
+        assert (
+            new_da.coords["foo"].to_index().symmetric_difference(expected.keys()).empty
```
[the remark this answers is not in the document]
timtroendle: Actually the next for loop should also go into that function. You are repeating everything in each test anyway.

timtroendle (changes requested):
Looks good, but I had a range of minor comments.

Snakefile
```diff
@@ -16,6 +16,7 @@ techs_template_dir = f"{model_template_dir}techs/"
 
 include: "./rules/shapes.smk"
 include: "./rules/data.smk"
+include: "./rules/jrc-idees.smk"
```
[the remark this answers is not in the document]
brynpickering: I prefer one per source as they do become large enough rule files to be worth splitting off. It also is in line with the concept of modularising different major data sources.

config/schema.yaml
```diff
@@ -144,6 +144,9 @@ properties:
     root-directory:
         type: string
         description: Path to the root directory of euro-calliope containing scripts and template folders.
+    max-threads:
```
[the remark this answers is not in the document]
brynpickering: Would you be OK with increasing max threads at the profile level?

lib/eurocalliopelib/utils.py
```diff
@@ -16,10 +18,10 @@ def eu_country_code_to_iso3(eu_country_code):
 
 def convert_country_code(input_country, output="alpha3"):
     """
-    Converts input country code or name into either a 2- or 3-letter code.
+    Converts input country code or name into either either a 2- or 3-letter code.
 
     ISO alpha2: alpha2
-    ISO alpha2 with Eurostat codes: alpha2_eu
+    ISO alpha2 with Eurostat codes: alpha2_eurostat
```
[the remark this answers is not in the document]
brynpickering: Sorry, I reverted this to `_eu` as JRC also uses it (you're right, it's EU-specific)

rules/jrc-idees.smk
```diff
@@ -0,0 +1,47 @@
...
+            f"data/automatic/jrc-idees/{country_code}.zip"
+            for country_code in [
+                pycountry.countries.lookup(country).alpha_2 for country in config["scope"]["spatial"]["countries"]
+            ]
+            if country_code in JRC_IDEES_SPATIAL_SCOPE
+        ]
+    params: sector_title_case = lambda wildcards: wildcards.sector.title()
+    wildcard_constraints:
+        sector = "((industry)|(transport)|(tertiary))"
+    output: temp(directory("build/data/jrc-idees/{sector}/unprocessed"))
```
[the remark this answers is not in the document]
brynpickering: hmm, ok.

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+    convert_country_code,
+    convert_valid_countries,
+    rename_and_groupby,
+)
+
+
+class TestRenameAndGroupby:
+    @pytest.fixture
+    def da(self, request):
+        data = [1, 2, 3, 4, 5]
```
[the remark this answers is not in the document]
brynpickering: It's good to have different values so you know that you've got the correct index items selected / grouped / summed later.

rules/jrc-idees.smk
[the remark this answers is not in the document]
brynpickering: Problem is that there's the `get_countries_to_unzip(config)` helper that now needs referencing in three different snakefiles. It also requires requires referencing the config options `config["data-pre-processing"]["fill-missing-values"]["jrc-idees"]`, `config["scope"]["spatial"]["countries"]` in each rule so that changes in those config items triggers a re-run. We could _not_ filter countries to process, so avoiding this helper function and associated params, or we filter but have a directory output so only need to reference everything once.

scripts/jrc-idees/industry.py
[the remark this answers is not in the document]
irm-codebase: Fix has been merged to main. I confirm it solves the issue.

brynpickering:
@timtroendle the only remaining question here is on the use of a directory output for unzipped JRC data. It's really a question around whether to filter countries or not. If we don't filter countries then we can return to the rule unzipping individual files with ease. If we filter, then it becomes reasonably cumbersome to reference the filtering helper function in multiple rule files. I could move the heat and transport JRC processing rules into `jrc-idees.smk` which would make referencing the helper function clearer.

tests/lib/test_utils.py
brynpickering: I have added a bit more explanation in the creation of the fixture

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+            (
+                {"A": "A1", "B": "A1", "C": "B1", "D": "B1", "E": "E1"},
+                {"A1": 3, "B1": 7, "E1": 5},
+            ),
+            ({"A": "A1", "B": "B1"}, {"A1": 1, "B1": 2, "C": 3, "D": 4, "E": 5}),
+            ({"A": "A1", "B": "B1", "C": "B1"}, {"A1": 1, "B1": 5, "D": 4, "E": 5}),
+        ],
+    )
+    @pytest.mark.parametrize("da", ["single_index", "multi_index"], indirect=True)
+    def test_rename_groupby_keep_renamed(self, rename_dict, expected, da):
```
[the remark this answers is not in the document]
brynpickering: Yes. I've also changed this to `drop_other_dim_items` to try and make it easier to understand the argument.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,202 @@
...
+
+    processed_data = pd.concat(demand_consumption).sort_index()
+    processed_data = processed_data.apply(utils.ktoe_to_twh)
+
+    return processed_data
+
+
+def process_production(data_filepaths):
+    all_dfs = []
+    for file in data_filepaths:
```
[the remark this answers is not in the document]
brynpickering: Updated for clarity. Different because it accesses just one sheet and is _much_ quicker. So you don't need to multi-process.

tests/lib/test_utils.py
```diff
@@ -1,39 +1,222 @@
...
+    rename_and_groupby,
+)
+
+
+class TestRenameAndGroupby:
+    @pytest.fixture
+    def da(self, request):
+        data = [1, 2, 3, 4, 5]
+        if request.param == "single_index":
+            index = pd.Index(["A", "B", "C", "D", "E"], name="foo")
```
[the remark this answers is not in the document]
brynpickering: Updated

tests/lib/test_utils.py
brynpickering: No, it's just a convenience to be able to compare to a single value in the other tests as it parameterises over a  single and multi index dataarray. Comparing either [1, 1] or [1] to 1 succeeds with numpy array comparisons. Can't do this with NaNs as the data is different in each array being compared (e.g. [1, np.nan]).

Snakefile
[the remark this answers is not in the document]
brynpickering: One could imagine a future where we split off JRC processing completely and just store the pre-built files on zenodo for convenience.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,214 @@
...
+    data_filepaths = list(Path(data_dir).glob("*.xlsx"))
+    if dataset == "energy":
+        processed_data = process_sheets(data_filepaths, threads, get_jrc_idees_energy)
+        unit = "twh"
+    elif dataset == "production":
+        processed_data = process_sheets(data_filepaths, 1, get_jrc_idees_production)
+        unit = "kt"
+
+    processed_data.columns = processed_data.columns.rename("year").astype(int)
+    processed_da = processed_data.stack().rename("jrc-idees-industry-twh").to_xarray()
```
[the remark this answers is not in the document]
irm-codebase: This proooooobably affects other outputs of this JRC module. But it would make further processing easier on our side.

irm-codebase (commented):
Suggested some small improvements to the output xarray datasets.

scripts/jrc-idees/industry.py
```diff
@@ -0,0 +1,214 @@
...
+
+    return (
+        df.loc[start.index[0] : end.index[0]]
+        .iloc[1:-1]
+        .dropna(how="all")
+        .assign(
+            country_code=df.index.name.split(":")[0],
+            cat_name=df.index.name.split(": ")[1],
+        )
+        .rename_axis(index="produced_material")
```
irm-codebase, line 90: This one is also a tad confusing. It seems that it's the _production method_ of a given material in some cases (Electric arc, Integrated steelworks both give kt of steel?).

In some cases it has units, in others it does not. Is there a way to make this tidier?

irm-codebase:
One last thing I noticed while processing data for #355: it seems this new JRC changes the final amounts for some energy carriers. Is this expected?

While implementing the processing for "other" industries, I ran into checksum differences between the JRC data in SCEC and EC. At first I thought it was the different energy unit ( `ktoe` to `twh`), but after more testing I found it really is not the case.

Should I open a separate issue for this, or would you like to debug it within this PR?

*****

Here are some checksum results before / after this JRC update:

**Before (in SCEC), already converted to `twh`**:
```python
ipdb> jrc_energy_df.loc[jrc_energy_df.index.get_level_values("carrier_name") == "Diesel oil (incl. biofuels)"].sum().sum()
2848.481153523807
ipdb> 
```

**After (using this PR in EC)**:
```python
ipdb> jrc_energy.sel(carrier_name="Diesel oil (incl. biofuels)").sum()
<xarray.Dataset>
Dimensions:       ()
Coordinates:
    carrier_name  <U27 'Diesel oil (incl. biofuels)'
Data variables:
    value         float64 2.309e+03
```

scripts/jrc-idees/industry.py
brynpickering: Not without lots of hardcoding. I would leave this to downstream applications and not this processing script, I think.

brynpickering:
@timtroendle I've moved back to non-directory output. See what you think about the country filtering - do you think it's overkill?

@irm-codebase I've cleaned up the output data files. Units are now only referenced in the xarray attributes, "..-twh" doesn't get attached to either dataset name. Energy is a dataset with two variables, production is a single array. 

I haven't managed to check the data inconsistency yet. I'll do that tomorrow.

brynpickering:
I've cleaned up the code so that it now matches the output of the old SC-EC notebook ([here](https://github.com/calliope-project/sector-coupled-euro-calliope/blob/main/notebooks/auxiliary_data_processing/process_jrc_idees.ipynb)). There were some issues with one of the cell colours not being captured for chemicals industry and with assigning carrier names to subsections. Both issues are now fixed and there are logging / data checking points added to keep an eye on it in future.

I've removed country filtering as it actually kills missing data filling for the heat sector. It could be cleaned up there to handle only a subset of countries, but I'd say that's a separate PR.

timtroendle:
Separate comment here re: country filtering. I am totally fine with not filtering countries at this stage of the workflow. I agree that it adds unnecessary complexity, especially considering that these files and their processing is very light weight.

config/schema.yaml
timtroendle, line 147: This parameter doesn't exist anymore and it should therefore not be in schema.yaml.

scripts/jrc-idees/heat.py
```diff
@@ -31,9 +31,8 @@
 
 
 def process_jrc_heat_tertiary_sector_data(
-    paths_to_national_data: list[str], out_path: str
+    paths_to_national_data: list[Path], out_path: str
```
timtroendle, line 34: Wait! These cannot be paths, they are strings coming directly from Snakemake.

scripts/jrc-idees/transport.py
```diff
@@ -39,12 +39,11 @@
 
 
 def process_jrc_transport_data(
-    paths_to_data: list[str],
-    dataset: object,
+    paths_to_data: list[Path],
```
timtroendle, line 42: Hm... is it me, or is this wrong too? Or is this a new Snakemake feature I am not aware of?

timtroendle (approved):
Looks all good to me. I find the file output clean and I agree with the non-filtering of countries.

Honestly, I went rather quickly through this as I had no major comments in the first iteration anyways. If there is anything specific I should have a close look at other than the file output, the country filtering, and the minor comments above, please let me know.

scripts/jrc-idees/transport.py
irm-codebase: I believe that is just a python annotation for better hinting.

irm-codebase:
> Separate comment here re: country filtering. I am totally fine with not filtering countries at this stage of the workflow. I agree that it adds unnecessary complexity, especially considering that these files and their processing is very light weight.

Honestly... I agree. It was not a good suggestion on my part. I did not follow KISS rules on this one...
Besides, if we do end up moving to a more modular setup, it makes sense to just filter stuff out at the end.

lib/eurocalliopelib/utils.py
```diff
@@ -49,12 +53,99 @@ def convert_country_code(input_country, output="alpha3"):
         return pycountry.countries.lookup(input_country).alpha_3
 
 
-# conversion utils
+def convert_valid_countries(country_codes: list, output: str = "alpha3") -> dict:
+    """
+    Convert a list of country codes / names to a list of uniform ISO coded country
+    codes. If an input item isn't a valid country (e.g. "EU27") then print the code and
+    continue, instead of raising an exception
+
```
irm-codebase, line 61: Quick question here: why not raise an exception, or introduce an optional flag for exception deactivation?
If possible, I'd raise exceptions by default, and use deactivation when needed (cautiously).

Specially because these are generic utils.

lib/eurocalliopelib/utils.py
```diff
@@ -49,12 +53,99 @@ def convert_country_code(input_country, output="alpha3"):
         return pycountry.countries.lookup(input_country).alpha_3
 
 
-# conversion utils
+def convert_valid_countries(country_codes: list, output: str = "alpha3") -> dict:
+    """
+    Convert a list of country codes / names to a list of uniform ISO coded country
+    codes. If an input item isn't a valid country (e.g. "EU27") then print the code and
+    continue, instead of raising an exception
```
irm-codebase, line 60: I'd be careful around skipping exceptions by default, since this is an utility function.

I'd introduce a flag to ignore exceptions, with the default being raising them. This helps avoid issues down the line since exceptions skipping is made explicit in the code calling this.

lib/eurocalliopelib/utils.py
```diff
@@ -49,12 +53,99 @@ def convert_country_code(input_country, output="alpha3"):
         return pycountry.countries.lookup(input_country).alpha_3
 
 
-# conversion utils
+def convert_valid_countries(country_codes: list, output: str = "alpha3") -> dict:
+    """
+    Convert a list of country codes / names to a list of uniform ISO coded country
+    codes. If an input item isn't a valid country (e.g. "EU27") then print the code and
```
irm-codebase, line 59: I'd be careful around skipping exceptions by default, since this is an utility function.

I'd introduce a flag to ignore exceptions, with the default being raising them. This helps avoid issues down the line since exceptions skipping is made explicit in the code calling this.

irm-codebase (approved):
Small comment regarding default exception ignoring in one file. Rest looks good!
I'll approve this. Change that line at your discretion :)

irm-codebase:
@brynpickering just to note that this fix might have introduced some extra hurdles... see  #383

PR #356, opened 2024-04-12
Feature EV controlled charging custom constraints

### LINKED ISSUE #81: Add electric vehicles with timeseries demand

### LINKED ISSUE #350: Constrain controlled charging of EVs in custom constraints
### What can be improved?

The share of controllable electrified annual road transport demand (i.e., whose charging pattern is optimised by the solver) should be constrained. This requires custom constraints as in the sector-coupled EC.

### Version

1.2.0.dev

adrienmellot:
Fixes #350.
Fixes #81.

Within the realm of full electrification of road transport, this PR allows the controlled charging share of this electricity demand to be constrained.

Possibly, we want to have the run.py file to be integrated in a better way, and also in workflow tests.

## Checklist

- [x] CHANGELOG updated
- [ ] Documentation updated


rules/transport.smk
```diff
@@ -45,16 +48,20 @@ rule annual_transport_demand:
         road_distance_historically_electrified = "build/data/transport/annual-road-transport-distance-demand-historic-electrification.csv",
     script: "../scripts/transport/annual_transport_demand.py"
 
-rule create_controlled_road_transport_annual_demand:
-    message: "Create annual demand for controlled charging at {wildcards.resolution} resolution"
+rule create_controlled_road_transport_annual_demand_and_installed_capacities:
```
brynpickering, line 51: Could this be split into two rules or not? I.e. annual demand and installed capacities are two rules

rules/transport.smk
```diff
@@ -63,6 +70,21 @@ rule create_controlled_road_transport_annual_demand:
         main = "build/data/{resolution}/demand/electrified-transport.csv",
     script: "../scripts/transport/road_transport_controlled_charging.py"
 
+rule create_controlled_ev_charging_parameters:
```
brynpickering, line 73: set a wildcard constraint on the allowed `dataset_name` values? It would help understand later all the files that can be generated by this rule

scripts/transport/road_transport_controlled_charging.py
```diff
@@ -65,6 +65,57 @@ def convert_annual_distance_to_electricity_demand(
...
+        df_ev_numbers.align(battery_size, level="vehicle_type")[1]
+        .squeeze()
+        .mul(df_ev_numbers)
+        .groupby(level=["country_code", "year"])
+        .sum()
+        .loc[country_codes]
+        .unstack("year")
+        .mul(transport_scaling_factor)
+    )
+    df_ev_chargeable_distance = df_ev_chargeable_distance.assign(**{
```
brynpickering, line 105: Add an ASSUME comment that >2015 uses 2015 data. Also use `final_year` to set the upper bound? `range(2016, final_year + 1)` (and don't even do this extrapolation if `final_year` <=2015)

scripts/transport/road_transport_controlled_charging.py
```diff
@@ -65,6 +65,57 @@ def convert_annual_distance_to_electricity_demand(
...
+        str(year): df_ev_chargeable_distance[2015] for year in range(2016, 2019)
+    })
+
+    df_ev_chargeable_distance.columns = df_ev_chargeable_distance.columns.astype(int)
+    # df_ev_chargeable_distance.index.name = None
```
brynpickering, line 110: delete

scripts/transport/road_transport_controlled_constraints.py
```diff
@@ -0,0 +1,103 @@
...
+
+def get_national_ev_profiles(
+    ev_profiles_path: str,
+    dataset_name: str,
+    demand_range: dict[str, int],
+    first_year: int,
+    final_year: int,
+    country_neighbour_dict: dict[str, list[str]],
+    country_codes: list[str],
+):
```
brynpickering, line 50: Missing return type

config/run.py
brynpickering, line 1: I'd suggest having a `run.py` in `scripts` and then a `add_constraints.py` in the `scripts/transport/` directory. The `run.py` has most of the content of this file _except_ the constraints. It then lods constraints as modules and loads them into the model:

```python
import importlib.util


model = calliope.read_netcdf(path_to_temp_model)
model.run(build_only=True)

for file in constraint_files:
    spec=importlib.util.spec_from_file_location("constraints", file)

    # creates a new module based on spec
    constraint_adder = importlib.util.module_from_spec(spec)

    # executes the module in its own namespace
    # when a module is imported or reloaded.
    spec.loader.exec_module(constraint_adder)

    constraint_adder.add_eurocalliope_constraints(model)

new_model = model.backend.rerun()
new_model.to_netcdf(path_to_output)
```

Then you can have the run step be a snakemake rule with the reference to the constraint scripts in the inputs.

brynpickering: This is untested by me, loading files as modules might need some trial and error work

brynpickering (changes requested):
Thanks for this @adrienmellot, shows how complex our EV modelling is. Too complex...?

Anyway, some reasonably major comments about reorganising file structures:

1. `run.py` could be restructured into a snakemake rule script that calls constraint adding scripts on-the-fly
2. the controlled EV annual demand and installed capacities rule should probably be split in two since they have some mutually exclusive inputs/params. They can use the same script (since they share the regionalisation step) but some parameter / wildcard is used in the script to decide whether `df_demand` or `df_charging_potentials` is generated.

adrienmellot:
@brynpickering I've addressed the first part of the comments regarding style, not yet the part concerning the `run.py` and the file structure for the custom constraints.

Regarding your comment about splitting the controlled EV annual demand and installed capacities rule into two rules, I've discussed it with Tim and it makes life harder to split it. The reason for that is that the rule outputs one `.csv` file only which is later read to generate the `.yaml` for the model. 

Splitting the rule into two would either require generating two `.csv` files, which makes it harder for the `.yaml` file to be generated afterwards (or two `.yaml` files, which we don't really want), or to have a rule order where one rule first generates a `.csv` and the second rule appends to this. All in all, a little more complex than having just one rule?

brynpickering:
OK, it's because the template filler can only handle one location-specific CSV. This probably needs fixing as it leads to quite messy implementations like this. Not something to solve in this PR!

adrienmellot:
For now custom constraints are not implemented (#385). I've thus removed the `run.py` file that there was to avoid any confusion. We discussed that the data processing could still be let through even though for now it is not immediately useful. I haven't updated the changelog and documentation for this PR as I don't think that it would not make much sense for now, it should probably be done once #385 is solved!

scripts/transport/road_transport_controlled_constraints.py
```diff
@@ -0,0 +1,103 @@
...
+    df = get_national_ev_profiles(
+        ev_profiles_path=snakemake.input.ev_profiles,
+        dataset_name=snakemake.wildcards.dataset_name,
+        demand_range=snakemake.params.demand_range,
+        first_year=snakemake.params.first_year,
+        final_year=snakemake.params.final_year,
+        country_neighbour_dict=snakemake.params.country_neighbour_dict,
+        country_codes=([
+            pycountry.countries.lookup(c).alpha_3 for c in snakemake.params.countries
+        ]),
```
brynpickering, line 94:
```suggestion
        country_codes=utils.convert_valid_countries(snakemake.params.countries).values()
```

scripts/transport/road_transport_controlled_constraints.py
```diff
@@ -0,0 +1,103 @@
+import pandas as pd
+import pycountry
```
brynpickering, line 2:
```suggestion
from eurocalliopelib import utils
```

config/schema.yaml
```diff
@@ -368,9 +371,43 @@ properties:
                             motorcycles:
                                 type: string
                                 description: JRC-IDEES name of motorcycles.
+                    ev-battery-sizes:
+                        type: object
+                        description: EV battery size per vehicle type.
```
brynpickering, line 376: Give unit here rather than as a comment in the config

brynpickering (approved):
Final set of suggestions, otherwise OK to go ahead!

brynpickering (changes requested):
Actually, can't approve - it needs a changelog entry ;)

brynpickering (approved):

PR #357, opened 2024-04-12
Pin hdf5 and h5py in all xarray environments

sjpfenninger:
This is pulling out the fix from #317 (commit 7b6bcf3) to resolve errors importing xarray on linux-x86, given that it seems #317 might not be merged any time soon.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


sjpfenninger:
In the meantime it seems that the fix perhaps does not work consistently, but we can still use this PR to track the required change, if further fixes are needed.

sjpfenninger:
Upon further testing, `libnetcdf=4.7` seems to make things work in #354. @irm-codebase Can you check and confirm?

timtroendle (approved):
Apart from a tiny documentation issue looks good to me!

PR #358, opened 2024-04-12
Fix uncontrolled charging RAMP profile

### LINKED ISSUE #287: Ensure existing RAMP profiles are appropriate for full electrification
### What can be improved?

We currently use [pre-compiled RAMP profiles](https://zenodo.org/record/6579421) for full electrification of small and light electric vehicles (implemented in #271).

This makes only sense, if these profiles represent charging power, not driving power. We nee to ensure that that's the case.

If that's not the case, we need to make necessary changes. Either post process the profiles, create other profiles using RAMP, or document this assumpution (if we think it's not too bad).

### Version

1.2.0.dev

FraSanvit:
Fixes #287.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -20,7 +20,7 @@ data-sources:
     swiss-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/gesamtenergiestatistik.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvNzUxOQ==.html
     swiss-industry-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/teilstatistiken.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvODc4OA==.html
     ev-data: https://zenodo.org/record/6579421/files/ramp-ev-consumption-profiles.csv.gz?download=1
-    charging-ev-data: https://sandbox.zenodo.org/records/45530/files/uncontrolled-charging-profiles.csv.gz?download=1
+    uncontrolled-ev-data: https://sandbox.zenodo.org/records/45530/files/uncontrolled-charging-profiles.csv.gz?download=1 # TODO: convert into Zenodo repository
 data-pre-processing:
     fill-missing-values:
         jrc-idees:
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -20,6 +20,7 @@ data-sources:
     swiss-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/gesamtenergiestatistik.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvNzUxOQ==.html
     swiss-industry-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/teilstatistiken.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvODc4OA==.html
     ev-data: https://zenodo.org/record/6579421/files/ramp-ev-consumption-profiles.csv.gz?download=1
+    charging-ev-data: https://sandbox.zenodo.org/records/45530/files/uncontrolled-charging-profiles.csv.gz?download=1
 data-pre-processing:
     fill-missing-values:
         jrc-idees:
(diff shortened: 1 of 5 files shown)


config/schema.yaml
```diff
@@ -107,6 +107,10 @@ properties:
                 type: string
                 pattern: ^(https?|http?):\/\/.+
                 description: Web address of electric vehicle data.
+            uncontrolled-ev-data:
+                type: string
+                pattern: ^(https?|http?):\/\/.+
+                description: Web address of electric vehicle data.
```
brynpickering, line 113: Improve the description? Say what should be in the columns/rows if it is a CSV

brynpickering: Maybe it should be called `uncontrolled_ev_profiles`?

scripts/transport/road_transport_timeseries.py
```diff
@@ -16,13 +16,13 @@ def create_road_transport_demand_timeseries(
...
 ) -> None:
     # Read annual road transport distance into panda dataframe
+
     df_annual = (
         pd.read_csv(path_to_annual_data, index_col=[0, 1, 2], parse_dates=[2])
         .squeeze()
         .xs(vehicle_type)
         .xs(slice(str(first_year), str(final_year + 1)), level="year", drop_level=False)
         .unstack("country_code")
-        .tz_localize("UTC")
```
brynpickering, line 25: Why is this no longer needed?

scripts/transport/road_transport_timeseries.py
```diff
@@ -69,7 +68,8 @@ def create_road_transport_demand_timeseries(
 
 def fill_empty_country(df, country_neighbour_dict):
     for country, neighbours in country_neighbour_dict.items():
-        assert country not in df.columns
+        if country in df.columns:
```
brynpickering, line 71: Is this because some countries now exist that didn't before?

brynpickering (commented):
Nice and small, very good. If we have improved input data from RAMP now (e.g. more countries), it would be good to have new demand, plugin, and battery consumption profiles all in the same zenodo record. Do you have the other two profiles from the run you did?

scripts/transport/road_transport_timeseries.py
adrienmellot: In this RAMP data it's already localised by UTC. `df_annual` is just constant in any case, so since the `df_timeseries` is already localised I think we can safely remove it.

config/schema.yaml
adrienmellot: We called it `uncontrolled-ev-data` just to be consistent with the `ev-data` for the controlled charging constraints (in the other PR, we've renamed this to `controlled-ev-data` so both are consistent). I guess they could both be called `(un)controlled-ev-profiles` if it's clearer

FraSanvit:
> Nice and small, very good. If we have improved input data from RAMP now (e.g. more countries), it would be good to have new demand, plugin, and battery consumption profiles all in the same zenodo record. Do you have the other two profiles from the run you did?

Not really. The plug-in profiles are custom outputs, but I can easily do that.

config/schema.yaml
brynpickering: `(un)controlled-ev-profiles` sounds good to me.

scripts/transport/road_transport_timeseries.py
brynpickering: OK, I would add a log message here that tells the user when a country is being skipped.

scripts/transport/road_transport_timeseries.py
brynpickering: My concern would be that the old RAMP data required this localization for some reason and that if you were to change the data source in the config to use old data then this script would break. If that wouldn't happen, then all's good.

config/schema.yaml
```diff
@@ -107,6 +107,10 @@ properties:
                 type: string
                 pattern: ^(https?|http?):\/\/.+
                 description: Web address of electric vehicle data.
+            uncontrolled-ev-data:
```
brynpickering, line 110: `data` -> `profiles`?

config/default.yaml
```diff
@@ -20,6 +20,7 @@ data-sources:
     swiss-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/gesamtenergiestatistik.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvNzUxOQ==.html
     swiss-industry-energy-balance: https://www.bfe.admin.ch/bfe/en/home/versorgung/statistik-und-geodaten/energiestatistiken/teilstatistiken.exturl.html/aHR0cHM6Ly9wdWJkYi5iZmUuYWRtaW4uY2gvZGUvcHVibGljYX/Rpb24vZG93bmxvYWQvODc4OA==.html
     ev-data: https://zenodo.org/record/6579421/files/ramp-ev-consumption-profiles.csv.gz?download=1
+    uncontrolled-ev-data: https://sandbox.zenodo.org/records/45530/files/uncontrolled-charging-profiles.csv.gz?download=1 # TODO: convert into Zenodo repository
```
brynpickering, line 23: `data` -> `profiles`?

rules/transport.smk
```diff
@@ -11,6 +10,16 @@ rule download_transport_timeseries:
     localrule: True
     shell: "curl -sSLo {output} {params.url}"
 
+rule download_uncontrolled_timeseries:
+    # TODO: move into rule download_transport_timeseries once PR 356 is merged
```
brynpickering, line 14: Am I right in saying that you want PR 356 merged in first? I'll go over there and leave a review, then!

scripts/transport/road_transport_timeseries.py
adrienmellot: I've added the localisation back to be sure

scripts/transport/road_transport_timeseries.py
adrienmellot: Done now

rules/transport.smk
adrienmellot: Actually no we'll fix the TODO once this PR 358 is merged in, sorry for the confusion

brynpickering (approved):
Great, just one last thing and then you can merge this in: update the changelog!

brynpickering:
@FraSanvit if you could re-generate all the profiles we need for euro-calliope with the latest RAMP (incl. the new countries) that would be great. Not for this PR, but at some point.

adrienmellot:
> Great, just one last thing and then you can merge this in: update the changelog!

I've looked at the changelog and I am not sure we can really update it with this PR because we're fixing something that is 'added' to v1.2.0. What do you think?

timtroendle:
You could simply add this PR number to that existing entry.

PR #361, opened 2024-04-15
Update default profile to be a workflow-profile

### LINKED ISSUE #268: Add Snakemake workflow profile as override of local profile
### What can be improved?

Snakemake 7.29.0 added the option to have a workflow profile that overrides local profiles. By default, this is `profiles/default`. We should use this mechanism to allow for local profiles.

As the workflow profile default works for us out of the box, we only need to update to 7.29.0. We can remove the env variable in the default environment.

### Version

1.2.0.dev

timtroendle:
Default profile (in `profiles/default`) is now the default workflow profile. Snakemake activates it automatically using the name. This allows the user to inject other profile parameters using the `--profile` cli option. This also allows us to provide the conda configuration as an addition to, rather than a replacement of, the default profile.

Fixes #268

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


profiles/conda/config.yaml
sjpfenninger, line 1: As of v23.10.0, conda uses libmamba as its solver by default. So the two should be equally fast. Do we still need two separate profiles? Furthermore, I think as long as conda is installed as well as mamba, conda is used anyway, irrespective of whether you used mamba to install the `euro-calliope` environment -- unless you run snakemake with `--conda-frontend mamba`.

sjpfenninger (changes requested):
The docs actually don't mention that there is a default workflow profile. This should probably be added, e.g. under the "Build" heading in `build.md`. Particularly mentioning that it sets `cores` to `2`.

brynpickering:
Am I right in saying that it would now not be necessary to include `--profile profiles/conda` when calling snakemake?

docs/workflow/build-remote.md
```diff
@@ -1,7 +1,7 @@
 # Build on a cluster
 
 You may want to build the model on a cluster.
-While you can build Euro-Calliope on [any cluster that is supported by Snakemake](https://snakemake.readthedocs.io/en/v6.1.1/executing/cluster.html), our default configuration is targeted at, and tested on, ETH's Euler cluster.
+While you can build Euro-Calliope on [any cluster that is supported by Snakemake](https://snakemake.readthedocs.io/en/v8.10.7/tutorial/additional_features.html#cluster-or-cloud-execution), our default configuration is targeted at, and tested on, ETH's Euler cluster.
```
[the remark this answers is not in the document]
sjpfenninger: Oh, we're using mkdocs, right. That makes it easier and we should definitely do it...

profiles/conda/config.yaml
timtroendle: Can we make the question whether we want to drop `mamba` a separate question? Right now, `mamba` is the default and whenever you want to run the workflow without, you have to use `--profile="profiles/conda"` or `--conda-frontend conda` which is essentially the same.

timtroendle:
No, it isn't. This pull request isn't changing any behaviour, but it adds the possibility to inject your own profile through `--profile="your-profile"`.

We could get rid of this by dropping mamba support and running conda by default.

profiles/conda/config.yaml
sjpfenninger: Yes, agreed, that question is beyond the scope of this PR

brynpickering (commented):
Should `euler/config.yaml` also be updated? It has `conda-prefix` and `software-deployment-method` which are now duplicated in the default profile.

timtroendle:
Yes, that makes sense. Done.

PR #366, opened 2024-04-23
Added pyproject.toml with ruff and pytest config.

### LINKED ISSUE #365: Centralize python configurations in pyproject.toml whenever possible
### What can be improved?

Euro-Calliope should follow [PEP 621](https://peps.python.org/pep-0621/) and move  python-relevant project configuration to a `pyproject.toml` file.

**Advantages**
- This standard makes our project easier to evaluate for other python developers.
- Less files = less maintenance.

For now I've identified the following tools that can be moved: 
- `.ruff.toml` ([see here for specification](https://docs.astral.sh/ruff/configuration/))
- `pytest.ini` ([see here for specification](https://docs.pytest.org/en/latest/reference/customize.html))

### Version

1.0.0

irm-codebase:
Fixes #365  .

## Checklist

- [x] CHANGELOG updated
- [x] Tests added to cover contribution (not applicable)
- [x] Documentation updated (not applicable)
- [x] Configuration schema updated (not applicable)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## 1.2.0 (unpublished)
 
-### Added (files)
-
-* **ADD** pyproject.toml file to centralize python/project details (#365).
-
 ### Added (models)
 * **ADD** fully-electrified heat demand (#284).
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## 1.2.0 (unpublished)
 
+### Added (files)
+
+* **ADD** pyproject.toml file to centralize python/project details (#365).
+
 ### Added (models)
 * **ADD** fully-electrified heat demand (#284).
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## 1.2.0 (unpublished)
 
+### Added (files)
```
brynpickering, line 5: Maybe move this to `workflow` since we've been trying to keep this to just `Added (models)` and `Added (workflow)`. Or don't have it in the changelog, it's not necessarily something anyone _needs_ to know about.

brynpickering (approved):

CHANGELOG.md
irm-codebase: Ah, I thought that was always necessary! I reverted the commit, to make things simpler.

Thanks!

timtroendle (approved):

PR #368, opened 2024-05-01
Fix localrules

timtroendle:
The localrules keyword breaks when a workflow is used as a module. This does not seem to get fixed, see https://github.com/snakemake/snakemake/issues/945

Instead, the recommended solution seems to be to use the localrule directive per rule.

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


irm-codebase (approved):
I like this approach. Quite explicit.

However, localrule issues have been fixed in https://github.com/snakemake/snakemake/pull/2180, but we have not updated our snakemake version.

Still, this approach is way more maintainable since the directive is in each rule instead of a separate line.

timtroendle:
Thanks for the review @irm-codebase. Just to add: we did update Snakemake recently and are on almost the newest version (3 micro updates have been published since, :D).

And yes, the fix is the `localrule` thing, but the `localrules` is still broken. I think `localrules` should be deprecated, but that's for the Snakemake devs to decide.

brynpickering (approved):
Add changelog entry?

timtroendle:
Yes, that makes sense.

PR #369, opened 2024-05-01
Fix Linux compatibility without breaking ARM macOS compatibility

timtroendle:
Recent changes in #357 broke ARM compatibility as libnetcdf 4.7 is not available for macOS ARM. The first available version if 4.8.0 and this is also the only version that is compatible with the other dependencies.

@sjpfenninger could you please check whether this still works on linux?

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


timtroendle:
@irm-codebase, could you test this on Linux? If not, can you suggest someone on Linux that could review and test?

timtroendle:
@sjpfenninger I am currently working on this again. I am introducing a CI test that does these checks so we don't have to do them manually.

sjpfenninger:
Great. Also running this on linux right now, it seems to be fine so far. All the envs built.

sjpfenninger (approved):
Works in local testing

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,58 @@
...
+  environments-are-installable:
+    name: Environments are installable
+    runs-on: ["ubuntu-latest", "macos-latest"]
+    steps:
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+          miniconda-version: "latest"
+          activate-environment: environment.yaml
+      - name: DAG env
+        shell: bash -el {0}
```
brynpickering, line 20: Set this is a top-level default:

```
defaults:
  run:
    shell: bash -l {0}
```

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,58 @@
...
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+
+jobs:
+  environments-are-installable:
+    name: Environments are installable
+    runs-on: ["ubuntu-latest", "macos-latest"]
```
brynpickering, line 13: These need to be in a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) not in a list on `runs-on` (see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idruns-on) for why not a list):

```
jobs:
  environments-are-installable:
    strategy:
      matrix:
        os: [ubuntu-latest, macos-latest]
    runs-on: ${{ matrix.os }}
```

brynpickering (changes requested):

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,44 @@
...
+
+on:
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+
+jobs:
+  environments-are-installable:
```
brynpickering, line 11: You need to checkout the project to have access to the files: `- uses: actions/checkout@v3`

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,44 @@
...
+            os: ["ubuntu-latest", "macos-latest"]
+    defaults:
+        run:
+            shell: bash -l {0}
+    steps:
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+            miniconda-version: latest
+      - name: DAG env
+        run: conda env create -f ../envs/geo.yaml --dry-run
```
brynpickering, line 26: These should be `conda create`?

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,45 @@
...
+    defaults:
+        run:
+            shell: bash -l {0}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: conda-incubator/setup-miniconda@v3
+        with:
+            miniconda-version: latest
+      - name: DAG env
+        run: conda env create -f ./envs/geo.yaml --dry-run
```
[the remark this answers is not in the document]
timtroendle: I think "env create" is just fine. This is similar to how we define it in the docs.

timtroendle:
In the process, I've decided to drop support for Intel macOS. Maintaining compat of ARM macOS and Linux is hard enough and Intel macOS will only become less important in the future.

timtroendle:
@brynpickering, the new tests took 4 minutes to run. Do we really want that?

Possibly, I could split the job into several jobs so they run in parallel.

sjpfenninger (changes requested):
Wait... I spoke too soon..., this fails again in the same old way on linux:

```
Job 101: Extract national potentials and cost from raw biofuel data.

[...]

from ._netCDF4 import *
ImportError: f6f640ac8f16d5d71ef1d7e200f26108_/lib/python3.9/site-packages/netCDF4/../../../libnetcdf.so.19: undefined symbol: H5Pset_fapl_ros3
```

sjpfenninger:
I think it's good to test that the environments are installable, but this example shows that it's not enough - the failure here is some arcane library import issue that only occurs when the library is actually imported. I wonder if at least for cases like this, an additional test that simply imports all libraries, would be useful.

timtroendle:
Maybe. I wonder whether what we are seeing here is a bug. This _should_ not happen, right? So I wonder if it's worth to run complex tests for such a case.

Much worse is the fact that this is breaking linux compat. That means we either break linux or mac. A real fix would require updating many more dependencies.

timtroendle:
Well they are all kinda slow and because they run in sequence in just adds up. Interestingly, the ARM build was a lot faster.

brynpickering:
You could add another matrix level which is the environment file name, then they would run in parallel jobs. It would just be a longer list to look through in the checks.

timtroendle:
That's exactly what I am currently preparing :D

timtroendle:
Ok this is much faster. It clutters the list of checks slightly, but it's in fact cleaner too.

I will now try to find a solution to the dependency problem.

timtroendle:
This is a mess.

What I found out so far:
* Pinning h5py was a mistake. It was never a dependency and has been  introduced as a dependency.
* libnetcdf 4.8 seems the main problem. Previous versions of Euro-Calliope always used <4.8 because of some third packages (maybe numpy, scipy, or gdal) that required 4.7 instead of 4.8.
* libnetcdf <4.8 isn't available for ARM macOS
* libnetcdf 4.8.1 doesn't fix the problem.
* libnetcdf 4.9.2 fixes the problem. It requires many updates in other dependencies.
* calliope 0.6.10 requires libnetcdf < 4.9

Seems like a deadlock to me.

timtroendle:
We could provide different versions of libnetcdf for different platforms: 4.8 for Mac, 4.7 for Linux. Snakemake is able to do that, but it would be in a hacky way.
https://snakemake.readthedocs.io/en/stable/snakefiles/deployment.html#freezing-environments-to-exactly-pinned-packages

timtroendle:
@brynpickering, @sjpfenninger, doesn't this mean that current Calliope is broken on Linux? Or am I missing something here?

irm-codebase:
> @brynpickering, @sjpfenninger, doesn't this mean that current Calliope is broken on Linux? Or am I missing something here?

If by that you mean v0.7, no, it works quite well (I've tested the latest dev version on my machine).

timtroendle:
Thanks, good to know. In fact, I thought of 0.6.10. Even here, I am pretty sure it's not broken, but then some of my bullet points above must be wrong:

> * libnetcdf <4.8 isn't available for ARM macOS
> * libnetcdf 4.8.1 doesn't fix the problem of Linux compat.
> * libnetcdf 4.9.2 fixes the problem of Linux compat. It requires many updates in other dependencies.
> * calliope 0.6.10 requires libnetcdf < 4.9

timtroendle:
For reference, the above bullet points were incomplete. Obviously, the problem is not in a single library but in the wrong combination of libraries. The following list works on both linux and ARM mac and should solve this issue here:

```
    - netCDF4=1.6.2
    - hdf5=1.12.2
    - libnetcdf=4.8.1
```

timtroendle:
This (^) seems to fix the Linux compat issue but it requires massive updates in all envs that include GDAL. This is because libgdal depends on hdf5, too, and requires updates. That then requires updates in GDAL, which requires updates in fiona, rasterio, rasterstats, which then require updates in numpy, scipy, and even Python.

It would make sense to fiddle out the minimal change that works, but I won't have time before next week to work on that.

brynpickering:
A nightmare... In the non-calliope envs, you could do libnetcdf>4.8 to avoid the need to update the GDAL-related things?

brynpickering:
Necessary updates don't seem to have been quite so drastic as you suggested @timtroendle. I think this is now ready for review!

CHANGELOG.md
```diff
@@ -37,6 +38,19 @@
 
 ### Updated (workflow)
 
+* **UPDATE** environments to fix issues on Linux and Macos-arm64:
```
timtroendle, line 41: Should include reference to this pull request #369 and #357

CHANGELOG.md
```diff
@@ -65,7 +79,7 @@
 * **FIX** fixed optimisation tolerance of hydro power plants from xtol to xatol (#266).
 * **FIX** source of Exclusive Economic Zones (EEZ) to use a cache on [zenodo](https://sandbox.zenodo.org/records/45135) so that we can keep using v11 (#332).  FIXME: update to actual zenodo record before next Euro-Calliope release.
 * **FIX** fixed rule `download_basins_database`, which previously failed on some linux and mac machines, by requiring a more recent curl in the environment `envs/shell.yaml` (#267).
-* **FIX** pin `h5py`, `hdf5` and `libnetcdf` in all environments which rely on `xarray`, to prevent issues on linux-x86 (#357).
+* **FIX** pin `hdf5` and `libnetcdf` in all environments which rely on `xarray`, to prevent issues on linux-x86 (#357, #369).
```
timtroendle, line 82: Seems duplicate of the UPDATE statement above. Considering that this was never broken in releases, this sounds like an update and not a fix to me.

timtroendle:
> Necessary updates don't seem to have been quite so drastic as you suggested @timtroendle. I think this is now ready for review!

Awesome, thanks for the work! I just love my drama. I'll run the minimal workflow on both systems to verify nothing obvious is broken.

.github/workflows/environments.yaml
```diff
@@ -0,0 +1,38 @@
+name: Tests of conda environments on both Linux and Mac
+
+on:
+    push:
+        branches:
+            - main
+            - develop
+    pull_request:
+
```
brynpickering, line 9: Would be worth limiting this to only run when there have been changes in the relevant files.
Even better would be to skip matrix items that don't have file changes, but that's more complex (and relies on another third-party plugin).

```suggestion
on:
    push:
        branches:
            - main
            - develop
        paths:
           - envs/*.yaml
           - templates/environment.yaml
           - requirements-*.yaml
    pull_request:
        paths:
           - envs/*.yaml
           - templates/environment.yaml
           - requirements-*.yaml
```

brynpickering (approved):
One suggestion for an update to the workflow file, otherwise good to go.

.github/workflows/environments.yaml
timtroendle: Looks nice, great suggestion! The doc says the checks will stay in 'pending' in case path filters kick in, which is maybe not so nice. Anyways, let's give this a try and we can always remove it if we don't like it.

PR #370, opened 2024-05-03
Update ehighways resolution shape generation

### LINKED ISSUE #318: Build ehighways resolution units

brynpickering:
Supercedes #336
Fixes #318 

This is in response to a question you had @timtroendle around defining the shape source in `shapes` in the config. I went to clean up the unresolved comments from #336 and realised that it is a lot cleaner to take that approach. This PR leads to a less flexible solution - E-Highways is baked in more than in #336 - but it is cleaner.

As with #336, it only works with NUTS2013, but I will open an issue about that and have a fix ready to go that I'll add via another PR.
 
## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


scripts/shapes/units.py
```diff
@@ -3,95 +3,145 @@
...
-    ), f"Countries are missing in layer {layer_name}."
+def _iso3(country_name: str) -> str:
+    return utils.convert_country_code(country_name, output="alpha3")
 
 
-def _iso3(country_name):
-    return pycountry.countries.lookup(country_name).alpha_3
+def _continental_layer(units: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
+    "For the continental resolution, dissolve all"
+    #
```
timtroendle, line 107: Remove

scripts/shapes/units.py
```diff
@@ -3,95 +3,145 @@
...
-    assert all(
-        _iso3(country) in layer.country_code.unique() for country in countries
-    ), f"Countries are missing in layer {layer_name}."
+def _iso3(country_name: str) -> str:
+    return utils.convert_country_code(country_name, output="alpha3")
 
 
-def _iso3(country_name):
-    return pycountry.countries.lookup(country_name).alpha_3
+def _continental_layer(units: gpd.GeoDataFrame) -> gpd.GeoDataFrame:
```
timtroendle, line 105: Consider renaming to `merge_to_continental_layer` or similar.

timtroendle (approved):
I am liking this a lot. Just a few minor comments.

timtroendle (changes requested):
CHANGELOG and documentation updates are missing.

config/schema.yaml
```diff
@@ -517,6 +521,7 @@ properties:
                                 - "Norway"
                                 - "Serbia"
                                 - "Switzerland"
+                                - "Iceland" # added for ehighways
```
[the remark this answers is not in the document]
brynpickering: sorry, slipped through the net

envs/default.yaml
```diff
@@ -16,7 +16,7 @@ dependencies:
     - netCDF4=1.5.6
     - hdf5=1.10
     - h5py=3.1.0
-    - libnetcdf=4.7
+    - libnetcdf=4.8.0
```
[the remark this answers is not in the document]
brynpickering: I don't have a linux device to try it on, so we'll have to wait for one of the others

brynpickering:
@timtroendle I've addressed your issues but...

On running the minimal workflow with updated zenodo pointer (to new potentials data that includes `ehighways`), a previous issue rears its head (#99). This would be solved by merging in #239.

timtroendle:
Will have a look at #239 soon.

brynpickering:
Minimal workflow now runs successfully, including `all_tests` (which covers the `ehighways` resolution as well as the `regional` one).

config/schema.yaml
```diff
@@ -396,6 +396,10 @@ properties:
                 type: string
                 description: Nuclear technology capacity scenario. "current" will use today's capacities based on the JRC power plant database (PPDB). All other scenario names should point to a configuration CSV file specifying minimum and maximum capacities per country in MW with non-zero capacities (columns=[country, min, max]).
                 pattern: ^(current)|(\w+.csv)$
+            nuts-year:
+                description: Version of the Nomenclature of Territorial Units for Statistics to use in creating spatial units (see https://ec.europa.eu/eurostat/web/nuts/history for more info).
+                type: number
+                enum: [2013]  # FIXME: enable other years
```
timtroendle, line 402: Doesn't sound like a fixme to me.

timtroendle (approved):

PR #382, opened 2024-05-14
Update NUTS processing

### LINKED ISSUE #380: Update NUTS link to enable download of different NUTS years
### What can be improved?

The current NUTS shapefile data source is 2013 and although that year is given in the data source URL ( https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_2013_01M_SH.zip), the link to it doesn't work if changing the year. 

If we want to link tabular eurostat data to NUTS shapefiles, we'll need access to different years (different tables are based on different years). The correct link to the NUTS webserver with all years is: https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/

### Version

1.2.0.dev

brynpickering:
Fixes #380.

In the process I have cleaned up the NUTS processing script. It now has less going on for the same result and has type hinting and more docstrings.

## Checklist

- [x] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -9,7 +9,7 @@ data-sources:
     hydro-basins: https://www.dropbox.com/sh/hmpwobbz9qixxpe/AADeU9iCgMd3ZO1KgrFmfWu6a/HydroBASINS/standard/eu/hybas_eu_lev07_v1c.zip?dl=1
     hydro-stations: https://zenodo.org/record/5215920/files/energy-modelling-toolkit/hydro-power-database-v10.zip?download=1
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
-    nuts:  https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/shp/NUTS_RG_01M_{nuts_year}_4326.shp.zip
+    nuts: https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/shp/NUTS_RG_01M_{nuts_year}_4326.shp.zip
     potentials: https://zenodo.org/records/6600619/files/raw-potentials.zip?download=1
     entsoe-tyndp: https://2020.entsos-tyndp-scenarios.eu/wp-content/uploads/2020/06/TYNDP-2020-Scenario-Datafile.xlsx.zip
     jrc-ppdb: https://zenodo.org/record/3574566/files/JRC-PPDB-OPEN.ver1.0.zip
--- a/config/default.yaml
+++ b/config/default.yaml
@@ -9,7 +9,7 @@ data-sources:
     hydro-basins: https://www.dropbox.com/sh/hmpwobbz9qixxpe/AADeU9iCgMd3ZO1KgrFmfWu6a/HydroBASINS/standard/eu/hybas_eu_lev07_v1c.zip?dl=1
     hydro-stations: https://zenodo.org/record/5215920/files/energy-modelling-toolkit/hydro-power-database-v10.zip?download=1
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
-    nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_{nuts_year}_01M_SH.zip
+    nuts:  https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/shp/NUTS_RG_01M_{nuts_year}_4326.shp.zip
     potentials: https://zenodo.org/records/6600619/files/raw-potentials.zip?download=1
     entsoe-tyndp: https://2020.entsos-tyndp-scenarios.eu/wp-content/uploads/2020/06/TYNDP-2020-Scenario-Datafile.xlsx.zip
     jrc-ppdb: https://zenodo.org/record/3574566/files/JRC-PPDB-OPEN.ver1.0.zip
--- a/lib/eurocalliopelib/utils.py
+++ b/lib/eurocalliopelib/utils.py
@@ -21,15 +21,26 @@ def eu_country_code_to_iso3(eu_country_code):
     return convert_country_code(eu_country_code, output="alpha3")
 
 
-def convert_country_code(input_country, output="alpha3"):
+def convert_country_code(
+    input_country: str,
+    output: Literal["alpha2", "alpha2_eu", "alpha3", "name"] = "alpha3",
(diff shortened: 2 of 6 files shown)


config/default.yaml
```diff
@@ -9,7 +9,7 @@ data-sources:
     hydro-basins: https://www.dropbox.com/sh/hmpwobbz9qixxpe/AADeU9iCgMd3ZO1KgrFmfWu6a/HydroBASINS/standard/eu/hybas_eu_lev07_v1c.zip?dl=1
     hydro-stations: https://zenodo.org/record/5215920/files/energy-modelling-toolkit/hydro-power-database-v10.zip?download=1
     load: https://data.open-power-system-data.org/time_series/2019-06-05/time_series_60min_stacked.csv
-    nuts: https://ec.europa.eu/eurostat/cache/GISCO/geodatafiles/NUTS_{nuts_year}_01M_SH.zip
+    nuts:  https://ec.europa.eu/eurostat/cache/GISCO/distribution/v2/nuts/shp/NUTS_RG_01M_{nuts_year}_4326.shp.zip
```
timtroendle, line 12: Do these have the same resolution? If they were more or less precise compared with the previous file, this could lead to different results.

lib/eurocalliopelib/utils.py
```diff
@@ -23,11 +23,12 @@ def eu_country_code_to_iso3(eu_country_code):
 
 def convert_country_code(input_country, output="alpha3"):
     """
-    Converts input country code or name into either a 2- or 3-letter code.
+    Converts input country code or name into a 2- or 3-letter code or a normalised name.
 
     ISO alpha2: alpha2
     ISO alpha2 with EU codes: alpha2_eu
     ISO alpha3: alpha3
+    name: country name
```
timtroendle, line 31: This must be inverted: `country name: name`

lib/eurocalliopelib/utils.py
```diff
@@ -23,11 +23,12 @@ def eu_country_code_to_iso3(eu_country_code):
 
 def convert_country_code(input_country, output="alpha3"):
     """
-    Converts input country code or name into either a 2- or 3-letter code.
+    Converts input country code or name into a 2- or 3-letter code or a normalised name.
```
timtroendle, line 26: maybe: `normalised pycountry name`

scripts/shapes/nuts.py
```diff
@@ -1,178 +1,138 @@
...
-    attributes = pd.DataFrame(attributes)  # to be able to remove geo information
-    del attributes["geometry"]
-    shapes.merge(attributes, on="NUTS_ID", how="left").to_file(
-        path_to_output, driver=OUTPUT_DRIVER
-    )
-
 
-def normalise(path_to_nuts, path_to_output, crs, study_area, all_countries, schema):
-    """Normalises raw NUTS data.
+LOGGER = logging.getLogger(__name__)
```
timtroendle, line 14: How does the logger work? It clearly doesn't log to a file, right? Is console logging the default?

scripts/shapes/nuts.py
```diff
@@ -1,178 +1,138 @@
...
-def _fix_country_feature(feature):
-    # * IDs should have three letters instead of two
-    # * many country names are broken or missing
-    feature["properties"]["id"] = utils.eu_country_code_to_iso3(
-        feature["properties"]["id"]
+    """Preprocess downloaded NUTS data into a format that aligns with other shapefile data sources (e.g. GADM).
+
+    Args:
+        path_to_nuts (str): Path to downloaded NUTS data.
+        path_to_output (str): Path to save output GeoPackage.
```
timtroendle, line 29: "path to save geopackage TO", right?

scripts/shapes/nuts.py
```diff
@@ -1,178 +1,138 @@
...
+            "id": gdf_nuts.NUTS_ID,
+            "name": gdf_nuts.NAME_LATN,
+            "type": np.where(gdf_nuts.LEVL_CODE == 0, "country", np.nan),
+            "layer": "nuts" + gdf_nuts.LEVL_CODE.astype(str),
+            "proper": True,
+            "country_code": gdf_nuts.CNTR_CODE.apply(utils.eu_country_code_to_iso3),
+        },
+        crs=gdf_nuts.crs,
+    ).to_crs(crs)
+    gdf_nuts_clean = _fix_country_names(gdf_nuts_clean)
```
timtroendle, line 49: to avoid storing and restoring these data in different variables, you could "pd.pipe" them.

scripts/shapes/nuts.py
```diff
@@ -1,178 +1,138 @@
...
+    gdf_nuts: gpd.GeoDataFrame,
+    study_area: shapely.geometry.Polygon,
+    all_countries: list[str],
+) -> gpd.GeoDataFrame:
+    """Keep only geometries that are in the study area.
 
+    Remove:
+    * countries not in study country codes
+    * entire shapes is not in study area bounding box
+    * parts of multipolygons that are not in the study area (keeping the rest)
```
timtroendle, line 69: maybe change for understandability: 

```parts of multipolygons that are not in the study area (e.g. remove islands and keep the rest)```

timtroendle (approved):
Looks good to me. There are a few comprehension questions and minor improvement suggestions.

config/default.yaml
timtroendle: Two more comments here:

* Only now do I see the devilish superfluous whitespace. Had I seen this before, of course I would have rejected the pull request right away.
* Expecting the `{nuts_year}` wildcard in the schema could be a good idea. We already do that with other data sources.

brynpickering: 1. Yes, they are the same resolution (01M).
2. SHAME
3. will do.

scripts/shapes/nuts.py
brynpickering: It doesn't really work at the moment. If you set the log level to `warning` I think it will log to stdout, otherwise it's effectively silent. It's a placeholder for sorting out logging properly :P

scripts/shapes/nuts.py
brynpickering: "path to which geopackage will be saved" I suppose is the most grammatically correct...

scripts/shapes/nuts.py
brynpickering: I've updated it to actually log to the console. Could be an option to set the logging level using a config param or to log to a file given by the rule. For now it simply emulates what the former print statements were doing.

PR #390, opened 2024-06-06
Separate hot water and space heating demand profiles

### LINKED ISSUE #389: Separate space heating and hot water demand
### What can be improved?

#343 introduced heat demand timeseries, but the way that When2Heat works is that the base profile is a combination of space heating and hot water demand.

Since we have separated data from Eurostat and JRC-IDEES on annual hot water and space heating demands, the profiles should also be separated.

### Version

1.2.0.dev

brynpickering:
Fixes #389 

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/rules/heat.smk
+++ b/rules/heat.smk
@@ -1,16 +1,7 @@
-rule download_gridded_temperature_data:
-    message: "Download gridded temperature data"
-    params: url = config["data-sources"]["gridded-temperature-data"]
-    output: protected("data/automatic/gridded-weather/temperature.nc")
-    conda: "../envs/shell.yaml"
-    localrule: True
-    shell: "curl -sSLo {output} '{params.url}'"
-
-
-rule download_gridded_10m_windspeed_data:
-    message: "Download gridded 10m wind speed data"
-    params: url = config["data-sources"]["gridded-10m-windspeed-data"]
-    output: protected("data/automatic/gridded-weather/wind10m.nc")
+rule download_gridded_weather_data:
+    message: "Download gridded {wildcards.data_var} data"
+    params: url = lambda wildcards: config["data-sources"]["gridded-weather-data"].format(data_var=wildcards.data_var)
+    output: protected("data/automatic/gridded-weather/{data_var}.nc")
     conda: "../envs/shell.yaml"
     localrule: True
     shell: "curl -sSLo {output} '{params.url}'"
@@ -87,10 +78,11 @@ use rule create_heat_demand_timeseries as create_heat_demand_timeseries_historic
     output:
         "build/models/{resolution}/timeseries/demand/heat-demand-historic-electrification.csv",
 
+
 rule population_per_weather_gridbox:
     message: "Get {wildcards.resolution} population information per weather data gridbox"
     input:
-        wind_speed = rules.download_gridded_10m_windspeed_data.output[0],
+        weather_grid = "data/automatic/gridded-weather/grid.nc",
         population = rules.raw_population_unzipped.output[0],
         locations = rules.units.output[0]
     params:
@@ -105,8 +97,8 @@ rule unscaled_heat_profiles:
     message: "Generate {wildcards.resolution} heat demand profile shapes per for {wildcards.year} from weather and population data"
     input:
         population = rules.population_per_weather_gridbox.output[0],
-        wind_speed = rules.download_gridded_10m_windspeed_data.output[0],
-        temperature = rules.download_gridded_temperature_data.output[0],
+        wind_speed = "data/automatic/gridded-weather/wind10m.nc",
+        temperature = "data/automatic/gridded-weather/temperature.nc",
         when2heat = rules.download_when2heat_params.output[0]
     params:
         lat_name = "lat",
--- a/rules/heat.smk
+++ b/rules/heat.smk
@@ -101,8 +101,8 @@ rule population_per_weather_gridbox:
     script: "../scripts/heat/population_per_gridbox.py"
 
 
-rule gridded_unscaled_heat_profiles:
-    message: "Generate gridded heat demand profile shapes for {wildcards.year} from weather and population data"
+rule unscaled_heat_profiles:
+    message: "Generate {wildcards.resolution} heat demand profile shapes per for {wildcards.year} from weather and population data"
     input:
         population = rules.population_per_weather_gridbox.output[0],
         wind_speed = rules.download_gridded_10m_windspeed_data.output[0],
@@ -112,5 +112,5 @@ rule gridded_unscaled_heat_profiles:
         lat_name = "lat",
         lon_name = "lon",
     conda: "../envs/default.yaml"
-    output: "build/data/{resolution}/gridded_hourly_unscaled_heat_demand_{year}.nc"
-    script: "../scripts/heat/gridded_unscaled_heat_profiles.py"
+    output: "build/data/{resolution}/hourly_unscaled_heat_demand_{year}.nc"
+    script: "../scripts/heat/unscaled_heat_profiles.py"
--- a/rules/heat.smk
+++ b/rules/heat.smk
@@ -94,7 +94,8 @@ rule population_per_weather_gridbox:
 
 
 rule unscaled_heat_profiles:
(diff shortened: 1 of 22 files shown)


scripts/heat/unscaled_heat_profiles.py
```diff
@@ -188,7 +218,8 @@ def get_daily_heat_demand(
 
     def heat_function(t, parameters):
```
irm-codebase, line 219: I prefer to avoid using inner functions unless protection from outside callers is absolutely necessary. The main reason is that you cannot write tests for these.

Consider bringing it outside and renaming to `_heat_function` to specify that it's an inner function.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -206,6 +237,28 @@ def heat_function(t, parameters):
     return daily(temperature, average_wind, all_parameters, heat_function)
 
 
+def get_daily_hot_water_demand(
+    temperature: xr.DataArray, average_wind: xr.DataArray, all_parameters: pd.DataFrame
+) -> xr.DataArray:
+    """
+    A function for the daily water heating demand is derived from BDEW et al. 2015
```
irm-codebase, line 244: Please provide doi's for citations. I've found that it's sometimes quite difficult to find which paper the code is referring to.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -64,29 +64,53 @@ def get_unscaled_heat_profiles(
...
+    )
 
     # Map profiles to daily demand
     hourly_heat = get_hourly_heat_profiles(
         reference_temperature, daily_heat, hourly_params
     )
 
+    # As per When2Heat, we assume the reference temperature is always 30C for hot water demand calcs
+    hourly_hot_water = get_hourly_heat_profiles(
+        reference_temperature.clip(min=30), daily_hot_water, hourly_params
```
irm-codebase, line 78: Please move the magic number to the top of the file as constant `REF_TEMPERATURE` (or similar), explaining what it is.

Makes future modularisation easier, and helps people identify assumptions faster even if we do not modularise.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -188,7 +218,8 @@ def get_daily_heat_demand(
 
     def heat_function(t, parameters):
         # BDEW et al. 2015 describes this function combining parameters
-        celsius = t - 273.15  # The temperature input is in Kelvin
+        # Handle situation where the temperature input is in Kelvin
+        celsius = t - 273.15 if (t > 100).all() else t
```
irm-codebase, line 222: This feels a tad error prone... is there a reason for it?
I think either making users specify the unit, or documenting that they must convert to °C beforehand is more solid than changing it without notice.

irm-codebase (commented):
Quick glance at the code. Overall, it looks good!

Just a couple of minor comments.

scripts/heat/unscaled_heat_profiles.py
brynpickering: This is reasonable, though there are many assumptions in here that come directly from When2Heat and I wouldn't want to give the impression that they are _our_ assumptions...

scripts/heat/unscaled_heat_profiles.py
brynpickering: Fixed in #391

scripts/heat/unscaled_heat_profiles.py
brynpickering: OK, I've updated it to have them at the top anyway.

brynpickering:
Sorry for the messy operations - fixing a github diff issue was a pain.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -224,24 +257,107 @@ def daily(
...
+        temperature (xr.DataArray): Reference daily temperature in degrees C.
+        parameters (pd.DataFrame): Relevant parameters from When2Heat.
+
+    Returns:
+        xr.DataArray: Daily relative hot water demand.
+
+    """
+    celsius_clipped = temperature.clip(min=HOT_WATER_LOWER_BOUND_TEMP)
+
+    return parameters["m_w"] * celsius_clipped + parameters["b_w"] + parameters["D"]
```
irm-codebase, line 351: Small improvement here would be to create variables with explicit names for these parameters.
But it seems like when2heat also has this style, so it's a nice to have.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -224,24 +257,107 @@ def daily(
...
+                    all_parameters[(building, "windy")],
+                )
             )
             for building in buildings
         ],
         dim=pd.Index(buildings, name="building"),
     )
 
 
+def group_gridcells(gridded_data: xr.DataArray, weight: xr.DataArray) -> xr.DataArray:
```
irm-codebase, line 274: Looks like an internal "helper" function. Consider `_group_gridcells`.

irm-codebase (approved):
Minor comments. Reply/fix is optional.
Approved!

rules/heat.smk
```diff
@@ -1,16 +1,7 @@
...
-    shell: "curl -sSLo {output} '{params.url}'"
-
-
-rule download_gridded_10m_windspeed_data:
-    message: "Download gridded 10m wind speed data"
-    params: url = config["data-sources"]["gridded-10m-windspeed-data"]
-    output: protected("data/automatic/gridded-weather/wind10m.nc")
+rule download_gridded_weather_data:
+    message: "Download gridded {wildcards.data_var} data"
+    params: url = lambda wildcards: config["data-sources"]["gridded-weather-data"].format(data_var=wildcards.data_var)
```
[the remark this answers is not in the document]
timtroendle: Given that this downloads at least three different files, it seems useful to me. Otherwise we'd have to repeat the rule and config entries three times.

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -1,49 +1,69 @@
...
-    within themselves, but without meaningful units.
-    The profiles need to be scaled so their magnitude matches annual heat demand data
-    in a subsequent step.
-
+    """Produces time series of heat demand profiles with the correct shape, and consistent within themselves, but without meaningful units.
+
+    The profiles need to be scaled so their magnitude matches annual heat demand data in a subsequent step.
+
+    Args:
+        path_to_population (str): Gridded population data, which will act as the weighting to got from grid-level profiles to profiles per model unit.
```
timtroendle, line 39: FYI: In other workflows, I stopped handing over paths as strings but rather handing data in and out. After doing this for a while, I'd recommend it. Maybe we could discuss this technical detail at one of the dev calls.

The function signature is then:

```Python
def business_logic(data: pd.DataFrame) -> pd.DataFrame:
    # code here
    return df

if __name__ == "__main__":
    data = business_logic(pd.read_csv(snakemake.input[0]))
    data.to_csv(snakemake.outptu[0])
```

timtroendle (approved):

rules/heat.smk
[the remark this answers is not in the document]
brynpickering: four different ones including changes in #393!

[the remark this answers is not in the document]
irm-codebase: Guess it's just preference then! can stay as is :+1:

scripts/heat/unscaled_heat_profiles.py
brynpickering: I can see merits of both. Especially if reading the file is simple (no need to define header/row levels or parse dates, etc.), it would probably be easier to debug with data being handed in, rather than strings. Once you have lots going on with the data loading, I'm less certain about having so much going on in `if __name__ == "__main__"` because then debugging becomes hard (e.g. you can't load those parts of the script as methods in an interactive session).

Let's discuss at the next dev call.

scripts/heat/unscaled_heat_profiles.py
brynpickering: Yeah, pretty sure they don't even have full names in the base source (the german govt. agency documentation).

PR #391, opened 2024-06-07
Update zenodo source of gridded weather data

brynpickering:
Fixes some review comments from @irm-codebase in #390.

Made a specific PR as they are changes that are worth considering in isolation, I believe. Deals with data units, mostly.

I'll deal with the rest of the review comments in #390 directly.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/config/schema.yaml
+++ b/config/schema.yaml
@@ -114,7 +114,7 @@ properties:
             gridded-weather-data:
                 type: string
                 pattern: ^(https?|http?):\/\/.+{data_var}.*\.(nc)$
-                description: "Web address of gridded weather data. Expecting a netCDF file with the coordinates [time, site], the variables [{data_type}, lat, lon], and the attribute 'unit'."
+                description: "Web address of gridded weather data. Expecting a netCDF file with the coordinates [time, site], the variables [{data_var}, lat, lon], and the attribute 'unit'."
             when2heat-params:
                 type: string
                 pattern: ^(https?|http?):\/\/.+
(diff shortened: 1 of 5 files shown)


config/schema.yaml
```diff
@@ -111,14 +111,10 @@ properties:
...
+            gridded-weather-data:
                 type: string
-                pattern: ^(https?|http?):\/\/.+
-                description: "Web address of gridded temperature data. Expecting a netCDF file with the coordinates [time, site], and the variables [temperature, lat, lon]."
-            gridded-10m-windspeed-data:
-                type: string
-                pattern: ^(https?|http?):\/\/.+
-                description: "Web address of gridded 10m wind speed data. Expecting a netCDF file with the coordinates [time, site], and the variables [temperature, lat, lon]."
+                pattern: ^(https?|http?):\/\/.+{data_var}.*\.(nc)$
+                description: "Web address of gridded weather data. Expecting a netCDF file with the coordinates [time, site], the variables [{data_type}, lat, lon], and the attribute 'unit'."
```
timtroendle, line 117: Should be `[{data_var}, lat, lon]` I guess?

scripts/heat/unscaled_heat_profiles.py
```diff
@@ -218,17 +222,14 @@ def get_daily_heat_demand(
 
     def heat_function(t, parameters):
```
timtroendle, line 223: Type hints are missing everywhere but I understand they will be added in #390.

timtroendle (approved):
Looks good to me.

scripts/heat/unscaled_heat_profiles.py
brynpickering: Yes, and better variable naming

PR #393, opened 2024-06-07
Feature: heat pump cop profiles

### LINKED ISSUE #80: Model heat pumps as timeseries-influenced supply technology
Heat pumps have a temperature-dependent COP and capacity. This needs to be reflected in the addition of heat pumps as a supply technology (#39). The heat pump manufacturer WAMAK provides data sheets for all its heat pumps, which I've used to create input datasets for the temperature dependence of COP and capacity (see [this notebook](https://github.com/calliope-project/euro-calliope-2.0/blob/main/notebooks/cop.ipynb) in the sector-coupled sibling).

This requires a temperature timeseries, ideally on the same resolution and from the same source as the capacity factors (#31).

handling this temperature dependence to produce input data is implement [here](https://github.com/calliope-project/euro-calliope-2.0/blob/main/src/construct/heat_pump_characteristics.py) in our sector-coupled sibling.

brynpickering:
Fixes #80 by adding COP profiles.

Builds onchanges made to input weather data in #390.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
 
 * **ADD** Spatial resolution that aligns with the regions defined by the [e-Highway 2050 project](https://cordis.europa.eu/project/id/308908/reporting) (`ehighways`) (#370).
 
-* **ADD** fully-electrified heat demand (#284, #343, #390, #391).
+* **ADD** fully-electrified heat demand (#284, #343, #390, #391) and heat pumps with variable COP to meet that demand (#393).
```
irm-codebase, line 11: Swap for #393 for #80. The approach here generally tends to focus on issues, not PRs (although I might be wrong here).

config/default.yaml
```diff
@@ -201,6 +202,20 @@ parameters:
...
+            radiator-conventional: 65
+            hot-water: 60
+        space-heat-sink-ratio:  # All values are assumed.
+            underfloor: 0.1
+            radiator-large: 0.15
+            radiator-conventional: 0.75
+        heat-pump-ratio:  # see https://stats.ehpa.org, 2018 market data assuming current ashp = air-to-air AND air-to-water
+            ashp: 0.9
+            gshp: 0.1
+        correction-factor: 0.85  # [@Ruhnau:2019]
```
irm-codebase, line 218: Change for `[@Ruhnau:2019:10.1038/s41597-019-0199-y]`?

rules/heat.smk
```diff
@@ -106,3 +114,22 @@ rule unscaled_heat_profiles:
     conda: "../envs/default.yaml"
     output: "build/data/{resolution}/hourly_unscaled_heat_demand.nc"
     script: "../scripts/heat/unscaled_heat_profiles.py"
+
+
+rule heat_pump_cop_subset:
```
irm-codebase, line 119: Consider renaming to `heat_pump_cop`. I think the pypsa people usually name rules and scripts the same, and it's a good rule to follow. Makes searching a bit less ambiguous.

scripts/heat/heat_pump_cop.py
```diff
@@ -0,0 +1,164 @@
...
+
+    temperature_ds = xr.merge(
+        _load_temperature_data(filepath, first_year, final_year)
+        for filepath in [path_to_temperature_air, path_to_temperature_ground]
+    )
+
+    # 1. Get characteristics per sink method
+    pre_grouped_heat_pump_characteristics = (
+        xr.open_dataarray(path_to_heat_pump_characteristics)
+        .mean("product")  # ASSUME: take the average of all heat pump products
```
irm-codebase, line 53: Is `ASSUME` a general style we follow for assumptions?
It might be good to specify this in the docstring of the function.

scripts/heat/heat_pump_cop.py
```diff
@@ -0,0 +1,164 @@
...
+        .sel(data_type="COP")
+        .interp(sink_temp=list(sink_temperature.values()))
+        .assign_coords(sink_temp=list(sink_temperature.keys()))
+    )
+    # 2. Combine sink methods into space heating and hot water end uses,
+    # using weightings for space heating (hot water is a distinct sink method already)
+    assert (
```
irm-codebase, line 60: Small recommendation: if you are validating inputs, it's generally better to put these at the very start of the function. It's more explicit to other developers, and generally safer (the function will break before anything else is run if an input is set up improperly).

Nice to have.

envs/default.yaml
```diff
@@ -17,6 +17,7 @@ dependencies:
     - netCDF4=1.6.2
     - hdf5=1.12.2
     - libnetcdf=4.8.1
+    - scipy=1.9.1
```
irm-codebase, line 20: Is this needed? This is not used in this PR.

irm-codebase (commented):
Minor comments.
Things look generally good!

timtroendle:
I am confused. Why are you intending to merge into the `feature-hot-water` branch? Which PR to check first? I will now have  a look at the feature-hot-water PR.

CHANGELOG.md
brynpickering: We've been doing a bit of both so far... More PRs than issues in recent changelog entries actually. I agree that issues should be prioritised, but we'll need to go in and clean up the changelog to align everything 😄

config/default.yaml
brynpickering: No, we have these short refs that one can look up in the reference list according to that ref name.

rules/heat.smk
brynpickering: Yeah, I'm not sure why `subset` creeped in there (it's a rule moved over from another repo). I would usually follow the same approach, even without knowing the PyPSA people do it ;)

scripts/heat/heat_pump_cop.py
brynpickering: Yeah, we've been doing it this way for a while. It's a bit like "TODO" and "FIXME". I prefer to keep assumptions close to where the action is, especially since docstrings don't get used in the standard way in snakemake scripts (rendered in docs, referred to in interactive sessions).

envs/default.yaml
brynpickering: Yes, it's a dependency of xarray when doing linear interpolation. Xarray has a relatively slim set of core dependencies. You only realise you need something for one of their modules when your code breaks 😄

brynpickering:
> I am confused. Why are you intending to merge into the feature-hot-water branch? Which PR to check first? I will now have a look at the feature-hot-water PR.

To reduce the diff that you would be reviewing. Sorry for pushing this before the hot water one was ready! I thought it would be nice to have it online so that you _could_ review it while I was fixing up review comments on the hot water PR.

envs/default.yaml
irm-codebase: Should've known. I've seen this on in my own code...

CHANGELOG.md
irm-codebase: I think including PRs is fine. I mostly commented because I thought it was enforced, so no biggie.

irm-codebase (approved):
Looks good!

scripts/heat/heat_pump_cop.py
```diff
@@ -0,0 +1,163 @@
...
+    cop_winter = cop_monthly.sel(month=[12, 1, 2]).mean("month")
+    cop_summer = cop_monthly.sel(month=[6, 7, 8]).mean("month")
+    assert (
+        cop_summer > cop_winter
+    ).all(), "Found higher heating COP values in winter than in summer."
+    assert (cop >= 1).all(), "Found improbably low heat pump COP values (< 1)."
+
+    # population weighted COP.
+    weight = population / population.sum(["site"])
+    # `ds_cop` has dims [site, time], `weight` has dims [site, id], we want a final array with dims [id, time]
```
timtroendle, line 111: ds_cop = cop

scripts/heat/heat_pump_cop.py
```diff
@@ -0,0 +1,163 @@
...
+    cop = cop.fillna(cop_ashp)
+
+    # Sanity check that there is a. higher COP in summer than winter, b. no COP < 1 (worse than direct electrical heating)
+    cop_monthly = cop.groupby("time.month").mean()
+    cop_winter = cop_monthly.sel(month=[12, 1, 2]).mean("month")
+    cop_summer = cop_monthly.sel(month=[6, 7, 8]).mean("month")
+    assert (
+        cop_summer > cop_winter
+    ).all(), "Found higher heating COP values in winter than in summer."
+    assert (cop >= 1).all(), "Found improbably low heat pump COP values (< 1)."
```
timtroendle, line 107: These would make great unit tests. I would move them there.

timtroendle (changes requested):
Looks great.

In the sake of implementing features that can be tested right away, would it possibly to extend this PR to actually building in the heat pump? It seems to me that only a few minor things are missing so that we'd actually have heat pumps, right?

scripts/heat/heat_pump_cop.py
```diff
@@ -0,0 +1,163 @@
...
+    path_to_output: str,
+):
+    """Calculate heat pump Coefficient of Performance (COP) based on manufacturer data.
+
+    COP is calculated for air-source and ground-source heat pumps according to different source temperature data.
+
+    COP is aggregated from grid-cells to model resolution using a population-weighted sum.
+
+    Args:
+        path_to_temperature (str): Gridded temperature timeseries data.
```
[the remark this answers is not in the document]
brynpickering: We access them from zenodo individually and, because they are large files, I'd keep them separate. Not sure it's worth the intermediate snakemake step to merge them

scripts/heat/heat_pump_cop.py
brynpickering: pros of moving to unit tests: 

* more clear grouping of our tests 
* availability of pytest infrastructure
cons:  
* you don't find out something is a problem until the entire workflow has been completed (should we instead be aiming for fast failure...?)
* the test is far removed from the code that would cause test failure
* (in this example) the test would come after the data aggregation. I'd fix this by moving the aggregation of gridcells to IDs into the next rule in the process, so this rule's outputs can be fed into a test, but then the output files are quite large.

brynpickering: I guess this is a more high level question of when to test in a snakemake workflow. I'm have implemented the change locally but do wonder whether the cons outweigh the pros...

config/default.yaml
```diff
@@ -212,6 +213,20 @@ parameters:
             solid-fossil-eff: 0.15 # [Ramanathan_1994] scaled down 60%, based on values calculated by [Karunanithy_2016]
             biofuel-eff: 0.1 #  [Ramanathan_1994] scaled down 60%, based on values calculated by [Karunanithy_2016]
             electricity-eff: 0.5  # [Karunanithy_2016] based on 2/3 40% efficient direct electric, 1/3 70% efficient induction
+    heat-pump:
+        sink-temperature:
```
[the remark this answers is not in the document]
timtroendle: I think that's fine but it should be mentioned.

scripts/heat/heat_pump_cop.py
timtroendle: I see your point and I agree this may require a high-level discussion. I just want to add another important pro: you can test many more things and, because it's pytest, quite easily. For example, you can execute each test by country.

timtroendle: In terms of code modularity, it would be good to have tests (especially unit tests) closer to the actual scripts. In theory we could have a test runner per "module" (== ".smk" file), but it would add overhead.

brynpickering:
> would it possibly to extend this PR to actually building in the heat pump? It seems to me that only a few minor things are missing so that we'd actually have heat pumps, right?

Do you mean extending to having the electricity demand from heating be based on the when2heat profile & heat pump COP? Or do you mean having heat pumps as a templated technology option?

brynpickering:
@timtroendle I have:

1. added a heat pump supply template.
2. processed the heat pump outputs into a COP timeseries file.
3. moved data checks to tests.

It was a bit more work than expected ;)

I also found it a pain to debug tests as you can't drop into the debugger with `breakpoint()` in the same way as with the other scripts. It requires adding `--pdb` to the pytest call. I've added functionality to allow that in the `test` rule/runner.

brynpickering:
Failing tests are due to too many requests to access either the miniconda `.sh` file or a cached version of it. Not sure how we resolve that.

scripts/heat/heat_pump_cop.py
brynpickering: Moved although I maintain that for intermediate datasets it is better to do the checks as part of the script. For final model files, we can then have tests. This stops us having an overcomplicated DAG where loads of intermediate data files need to be kept around until the penultimate step.

templates/models/techs/supply/heat.yaml
timtroendle, line 1: In terms of modularity, should this file include _only_ heat pumps? I tend to think yes. Also, I'd prefer to include other supply technologies in other PRs, together with their carrier supplies which seem missing in this PR.

timtroendle (changes requested):
Awesome!

A few more minor changes. More importantly: updates to the docs are changing. There should be updates here:
https://euro-calliope.readthedocs.io/en/latest/model/customisation/#importing-modules

In the sake of modularity, shouldn't we split the heat yaml template into several files? I tend to think yes. Also, I would add additional techs in additional PRs, together with the respective carriers they require.

timtroendle:
I wouldn't worry too much about the failing tests. @setup-miniconda's tests are failing, too, so I am assuming they will have this high up on their agenda.

tests/model/test_heat.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+def test_cop_all_at_least_1(cop_timeseries):
+    """Expecting no COP < 1 (worse than direct electrical heating)"""
+    assert (
+        cop_timeseries.stack() >= 1
+    ).all(), "Found improbably low heat pump COP values (< 1)."
+
+
+def test_unscaled_heat(unscaled_space_heat_timeseries):
```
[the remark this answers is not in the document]
timtroendle: No, here it doesn't make sense because there are no locations.

timtroendle:
👆Oops, I wanted to suggest edits and now I've committed them. Must have missed the "suggest" option, sorry.

tests/model/test_heat.py
```diff
@@ -0,0 +1,25 @@
+def test_cop_monthly(cop_timeseries):
```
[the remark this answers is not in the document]
brynpickering: It's effectively being done per location anyway, right? If the test fials, you'd be able to pinpoint the offending location pretty quickly when debugging...

templates/models/techs/supply/heat.yaml
brynpickering: oops, sorry. A staging error. I'll revert the non-electricity ones. I'd maybe do it per type of carrier in?

templates/models/techs/supply/heat.yaml
```diff
@@ -0,0 +1,104 @@
...
+            parent: conversion
+            carrier_in: biofuel
+            carrier_out: biofuel_heat
+        constraints:
+            energy_eff: 0.8
+            lifetime: 20
+        costs:
+            monetary:
+                # TODO: weight averaging per region based on dwelling ratio
+                energy_cap: {{ mean([5.9 / 10, 5.9 / 8, 76 / 400, 45 / 160]) * 1e6 * scaling_factors.specific_costs }}  # {{ (1 / scaling_factors.specific_costs) | unit("EUR2015/MW_heat") }} | average of new/existing and single-/multi-family homes.
```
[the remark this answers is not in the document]
brynpickering: They're all from the DEA. It's definitely better to have them explicitly stated in here as they're easier to compare to the DEA tables. Still, I should probably state it over multiple lines (is that possible with the jinja2 template?) with a comment for each

scripts/heat/group_gridded_timeseries.py
```diff
@@ -0,0 +1,88 @@
...
+        .to_xarray()
+        .sel(end_use=["space_heat", "hot_water"])
+        .sum("cat_name")
+    )
+
+
+def _site_weighted_ave(
+    id: str, gridded_data: xr.Dataset, grid_weight: xr.DataArray
+) -> xr.Dataset:
+    """Multi-processing helper function to get the weighted average of all gridcells for a given spatial unit (id)."""
```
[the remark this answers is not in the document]
brynpickering: More that it only exists because it is needs to be its own function so that it can be called by the multiprocessor.

scripts/heat/group_gridded_timeseries.py
```diff
@@ -0,0 +1,88 @@
...
+    threads: int,
+) -> pd.DataFrame:
+    """Group gridded heat data into resolution-specific units, taking a weighted average of grid values.
+
+    Also take a weighted average of hot water and space heat demand (using per-unit annual demands) to return a single "heat" timeseries.
+
+    Args:
+        gridded_data (xr.Dataset): Gridded timeseries space heat and hot water data.
+        grid_weight (xr.DataArray): Weighted mapping from grid (a.k.a. "site") to units.
+        annual_demand (xr.DataArray): Per-unit annual space heating and hot water demands.
```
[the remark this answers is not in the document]
brynpickering: It's doing two sets of weighted averages: weighted averaging "sites" into "IDs" (using population) and weighted averaging "space heat" and "hot water" demand into a general profile for "heat" (using annual demands).

brynpickering:
@timtroendle I think that's all done now. I've opened various issues for things that need to be resolved off the back of this PR.

templates/models/techs/supply/heat-from-electricity.yaml
```diff
@@ -0,0 +1,74 @@
...
+
+locations:
+    {% for id, location in locations.iterrows() %}
+    {{ id }}:
+        techs:
+            heat_pump:
+            electric_heater:
+    {% endfor %}
+
+overrides:
```
timtroendle, line 44: These overrides should be in the `templates/models/techs/storage/heat.yaml`, right? Otherwise we are creating a dependency between these two files.

I mean, this will break if users don't import the file with the storage techs.

timtroendle: Also: consider adding these overrides to `tests/resources/test.yaml`.

timtroendle (approved):
Nice!

Just one, I think important, comment: there is a dependency between the "storage/heat" and "supply/heat" files now which should not be the case. By moving the overrides to the "storage/heat" you should be able to get rid of the dependency. The only remaining dependency are the carrier definitions and we will not and don't have to get rid of those.

templates/models/techs/supply/heat-from-electricity.yaml
```diff
@@ -0,0 +1,74 @@
+{# TODO: weight single-/multi-family technology costs based on regional dwelling ratio #}
```
timtroendle, line 1: Would be useful to link to the Github issue here: #406.

templates/models/techs/supply/heat-from-electricity.yaml
brynpickering: We're creating a dependency either way as they also influence the output carrier of the supply technology. I think they are better off in `supply/heat-from-electricity.yaml` as the storage techs are actually abstract tech groups so there's nothing wrong with including that YAML file as an input by default.

As an aside: this won't be an issue in v0.7 as we can trigger a storage buffer for these kinds of techs when defining them, so I wouldn't put too much weight on whether we've got it perfect right now.

timtroendle: Ok up to you. I guess you have the better overview.

I just think we should try to minimise dependencies between files as much as possible. As far as I can tell, this would be the first dependency between two files. It's also slightly surprising that storage gets added through a file in "supply" instead of "storage".

brynpickering: OK, I've decoupled them now, although this does add a bit of unnecessary model complexity when _not_ using the storage buffers. Since this will all be swept away with v0.7 that complexity is probably fine for now.

PR #399, opened 2024-06-11
Fix warning spam caused by geopandas update

irm-codebase:
Fixes  spam caused by geopandas (`op` was replaced with  `predicate`). Found while debugging something else.

See: https://geopandas.org/en/stable/docs/reference/api/geopandas.sjoin.html

## Checklist

- [ ] ~~CHANGELOG updated~~
- [ ] ~~Tests added to cover contribution~~
- [ ] ~~Documentation updated~~
- [ ] ~~Configuration schema updated~~

### DIFF
--- a/rules/hydro.smk
+++ b/rules/hydro.smk
@@ -155,7 +155,7 @@ rule hydro_capacities:
 rule capacity_factors_hydro:
     message: "Generate capacityfactor time series for hydro electricity on {wildcards.resolution} resolution."
     input:
-        capacities = rules.hydro_capacities.output[0],
+        capacities = rules.hydro_capacities.output.supply,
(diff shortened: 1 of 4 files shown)


rules/hydro.smk
```diff
@@ -155,7 +155,7 @@ rule hydro_capacities:
 rule capacity_factors_hydro:
     message: "Generate capacityfactor time series for hydro electricity on {wildcards.resolution} resolution."
     input:
-        capacities = rules.hydro_capacities.output[0],
+        capacities = rules.hydro_capacities.output.supply,
```
irm-codebase, line 158: This is the only case were a rule with more than one output had `output[0]` in `hydro.smk`.
Small improvement.

timtroendle (approved):

PR #394, opened 2024-06-10
Feature non-electric vehicles in transport

adrienmellot:
Fixes #273, #274 and #351 .

This PR introduces diesel vehicles on top of EVs, to reproduce the sector-coupled euro-calliope workflow. 

## Checklist

Any checks which are not relevant to the PR can be pre-checked by the PR creator. All others should be checked by the reviewer. You can add extra checklist items here if required by the PR.

- [ ] CHANGELOG updated
- [ ] Minimal workflow tests pass
- [ ] Tests added to cover contribution
- [ ] Documentation updated
- [ ] Configuration schema updated

PR #409, opened 2024-07-02
 Add `.jinja` suffix; reorganise template rules

### LINKED ISSUE #404: Add `.jinja` to all template files
### What can be improved?

with a `.jinja` suffix, IDEs can apply correct syntax highlighting to our template files.

### Version

1.2.0.dev

brynpickering:
Fixes #404 
Also fixes `ruleorder` issue highlighted in #394. I achieved this by merging `{tech_group}/{tech}` into `{tech_and_group}` which allows us to add wildcard constraints that are combinations of tech group and tech (currently, only `supply/biofuel`).

## Checklist

- [x] CHANGELOG updated
- [ ] Documentation updated


Snakefile
```diff
@@ -96,30 +94,28 @@ rule all_tests:
...
     wildcard_constraints:
-        tech_group = "(?!transmission).*"  # i.e. all but transmission
+        # Exclude all outputs that have their own `techs_and_locations_template` implementation
+        tech_and_group = "(?!transmission\/|supply\/biofuel).*"
     conda: "envs/default.yaml"
-    output: "build/models/{resolution}/techs/{tech_group}/{tech}.yaml"
+    output: "build/models/{resolution}/techs/{tech_and_group}.yaml"
     script: "scripts/template_techs.py"
 
+use rule techs_and_locations_template as dummy_tech_locations_template with:
```
timtroendle, line 114: consistent naming would be "dummy_techs_and_locations_template".

timtroendle: Anyways, maybe different names would be more helpful in communicating the difference. I don't really understand why this is called a "dummy". Is it maybe: "tech_and_locations_template_with(out)_location_specific_data"? That seems the be the differentiating factor to me.

Or even more radical and better in my opinion: "model_component_with(out)_location_specific_data". We don't have a clear rule, but typically rules are named by (1) what they do or (2) what they produce, not by their inputs (here: template).

Or:  "module_with(out)_location_specific_data". That's how we call these files in the documentation. My favourite right now.

timtroendle (changes requested):
The jinja part of this PR makes a lot of sense.

The part related to the "tech_locations_template" rules is nice in the sense that it removes all these rule dependencies. It comes at the price that "tech_group" and "tech" are merged into a single wildcard. That's less clear I find, but I didn't find a better solution either.

I definitely think the rules and wildcards should have different naming. I always takes my a while to understand what they are doing.

* "tech_and_group" -> "group_and_tech", because that's the order.
* I suggest to rename all "template" rules to "module" rules, because rules are typically named by their outputs (a model module in these cases) instead of their inputs (a template).
* "locations_template" -> "locations", or "locations_module"
* "techs_and_locations_template" -> "tech_module" (potentially plus "with(out)_location_data", see my comment)
* "bio_techs_and_locations_template" -> "biofuel_tech_module"

Snakefile
```diff
@@ -96,30 +94,28 @@ rule all_tests:
...
+    # For all cases where we don't have any location-specific data that we want to supply to the template
+    input:
+        template = techs_template_dir + "{tech_and_group}.yaml.jinja",
+        locations = rules.locations_template.output.csv
```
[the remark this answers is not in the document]
brynpickering: `locations_from_template`? I still like the link to templates in the rule name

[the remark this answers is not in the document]
brynpickering: or `locations_calliope_input`?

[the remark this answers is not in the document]
timtroendle: Personally, I like `locations_module` best. We already call these files modules and this would be most consistent. I am fine with renaming these files to something else, but I'd be consistent -- including in the docs.

Snakefile
```diff
@@ -96,32 +94,30 @@ rule all_tests:
...
     script: "scripts/template_techs.py"
 
+use rule module_with_location_specific_data as module_without_location_specific_data with:
+    # For all cases where we don't have any location-specific data that we want to supply to the template
+    input:
+        template = techs_template_dir + "{group_and_tech}.yaml.jinja",
+        locations = rules.locations.output.csv
 
-rule no_params_model_template:
+rule model_file_without_specific_data:
```
timtroendle, line 120: `module_without_specific_data`?

Snakefile
```diff
@@ -132,7 +128,7 @@ rule no_params_model_template:
     shell: "cp {input.template} {output}"
 
 
-rule no_params_template:
+rule non_model_file_without_specific_data:
```
timtroendle, line 131: `auxiliary_file(_without_specific_data)` maybe?

Snakefile
```diff
@@ -143,10 +139,10 @@ rule no_params_template:
...
 
-rule model_template:
+rule model:
     message: "Generate top-level {wildcards.resolution} model configuration file from template"
     input:
-        template = model_template_dir + "example-model.yaml",
+        template = model_template_dir + "example-model.yaml.jinja",
         non_model_files = expand(
             "build/models/{template}", template=["environment.yaml", "README.md"]
         ),
```
timtroendle, line 148: `input_files` -> `modules`

rules/shapes.smk
```diff
@@ -140,10 +140,10 @@ rule eez:
         """
 
 
-rule locations_template:
+rule locations:
```
timtroendle, line 143: `locations_module`?

Snakefile
```diff
@@ -143,10 +139,10 @@ rule no_params_template:
     shell: "cp {input.template} {output}"
 
 
-rule model_template:
+rule model:
     message: "Generate top-level {wildcards.resolution} model configuration file from template"
     input:
-        template = model_template_dir + "example-model.yaml",
+        template = model_template_dir + "example-model.yaml.jinja",
         non_model_files = expand(
```
timtroendle, line 146: `auxiliary_files`?

timtroendle (approved):
I like this. There are a few more things we could change, so that we have a consistent name for this type of files.

PR #410, opened 2024-07-02
Adding_Trial_Logger_Mayank_7/2/24

MayankKonduri:
Fixes # .

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated


timtroendle:
I merged the current develop branch into this PR to get latest updates.

envs/test.yaml
```diff
@@ -1,14 +1,13 @@
 name: test
 channels:
     - conda-forge
-    - gurobi
```
timtroendle, line 4: This is a change that you applied locally to be able to run without access to gurobi. This change should not be merged and should therefore be reversed.

It may be a good idea to better support use without gurobi, but this would require its own PR.

templates/environment.yaml
```diff
@@ -1,7 +1,6 @@
 name: euro-calliope
 channels:
     - conda-forge
-    - gurobi
```
timtroendle, line 4: This is a change that you applied locally to be able to run without access to gurobi. This change should not be merged and should therefore be reversed.

It may be a good idea to better support use without gurobi, but this would require its own PR.

templates/models/example-model.yaml
```diff
@@ -8,7 +8,7 @@ model:
     timeseries_data_path: ./timeseries
     subset_time: ['{{ year }}-01-01', '{{ year }}-01-01']
 run:
-    solver: gurobi
+    solver: cbc
```
timtroendle, line 11: This is a change that you applied locally to be able to run without access to gurobi. This change should not be merged and should therefore be reversed.

It may be a good idea to better support use without gurobi, but this would require its own PR.

scripts/demand/national_load.py
```diff
@@ -16,10 +16,14 @@
 """
 
 import calendar
+import logging
 
 import pandas as pd
 import pycountry
 
+logger = logging.getLogger(_name_)
+logging.basicConfig(filename="snakemake.log", encoding="utf-8", level=logging.DEBUG)
```
timtroendle, line 25: This will not work, as you are handing over `snakemake.log` as a string rather than an object. It's an object, so you need to remove the quotes.

scripts/demand/national_load.py
```diff
@@ -16,10 +16,14 @@
 """
 
 import calendar
+import logging
 
 import pandas as pd
 import pycountry
 
+logger = logging.getLogger(_name_)
```
timtroendle, line 24: This line is failing for me.

rules/demand.smk
```diff
@@ -19,7 +19,9 @@ rule electricity_load_national:
         final_year = config["scope"]["temporal"]["final-year"],
         data_quality_config = config["quality-control"]["load"],
         countries = config["scope"]["spatial"]["countries"]
-    output: "build/data/electricity-demand-national.csv"
+    output:
+	csv: "build/data/electricity-demand-national.csv"
```
timtroendle, line 23: This is a Snakemake syntax error. "csv" is an output, so it (1) should be indented, and (2) requires `=` instead of `:`.

timtroendle (changes requested):
Thanks for this PR, @MayankKanduri. There are two issues with your PR. First, you have committed changes that were necessary to run the workflow locally on your machine but that should not be merged to the main codebase. These are the changes related to replacing gurobi with coincbc. Please revert these changes, otherwise we cannot merge.

Second, even more important, your changes do not work. They fail on my machine and they fail on the GitHub actions (see their output). Please test your changes locally first to make sure you have no syntax error and that you receive the expected result. You can run the Snakemake rule that you have changed with this command:

`snakemake --configfile="config/minimal.yaml" "build/data/electricity-demand-national.csv" `

20191216-C3S-L2_GHG-GHG_PRODUCTS-TANSO2-GOSAT2-SRFP-DAILY-v2.0.0.nc
timtroendle, line 1: This file should not be part of the repo.

cdsapirc
timtroendle, line 1: This file should not be part of the repo.

download.zip
timtroendle, line 1: This file should not be part of the repo.

download_data.py
timtroendle, line 1: This file looks like a test you were running. It should certainly not be part of the repository.

envs/test.yaml
timtroendle: These gurobi-based changes still need to be reverted.

envs/test.yaml
```diff
@@ -1,14 +1,13 @@
...
 dependencies:
     - python=3.9
     - ipdb=0.13.13
     - numpy=1.23
     - pandas=1.5
     - xarray=2022.3
-    - gurobi=9.5.1
+    - coincbc
```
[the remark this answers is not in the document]
timtroendle: These gurobi-based changes still need to be reverted.

logs/electricity_load_national.log
timtroendle, line 1: This logging file is almost exactly what we are looking for. However, the file itself should not be part of the repository. It will be created whenever users are running the workflow.

templates/environment.yaml
timtroendle: These gurobi-based changes still need to be reverted.

templates/models/example-model.yaml
timtroendle: These gurobi-based changes still need to be reverted.

timtroendle (changes requested):

PR #348, opened 2024-04-11
Add chemicals industry (old)

tud-mchen6:
Fixes #311 .

Adding chemicals industry in the industry module.

## Checklist

Any checks which are not relevant to the PR can be pre-checked by the PR creator. All others should be checked by the reviewer. You can add extra checklist items here if required by the PR.

- [x] CHANGELOG updated
- [x] Minimal workflow tests pass
- [ ] Tests added to cover contribution (not relevant)
- [ ] Documentation updated (not relevant)
- [ ] Configuration schema updated (not relevant)

PR #418, opened 2024-07-25
Add chemicals industry (new)

### LINKED ISSUE #311: Update chemical industry demand

tud-mchen6:
Fixes #311. This is the newer version; discarding #348. Everything works, but still in progress of prettifying.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/modules/industry/scripts/chemicals_industry.py
+++ b/modules/industry/scripts/chemicals_industry.py
@@ -56,7 +56,6 @@ def get_chemicals_demand_df(
     new_chemicals_demand = filling.fill_missing_countries_years(
         energy_balances_df, cat_names_df, carrier_names_df, new_chemicals_demand
     )
-    breakpoint()
     new_chemicals_demand.to_netcdf(output_path)
 
 
--- a/modules/industry/scripts/chemicals_industry.py
+++ b/modules/industry/scripts/chemicals_industry.py
@@ -3,11 +3,17 @@ import xarray as xr
 from utils import filling
 from utils import jrc_idees_parser as jrc
 
+H2_LHV_KTOE = 2.863  # 0.0333 TWh/kt LHV -> 2.863ktoe/kt
+MWH_PER_T_TO_KTOE_PER_KT = 0.08598  # 1 MWh/t -> 1 GWh/kt -> 0.08598 ktoe/kt
+METHANOL_LHV_KTOE = 0.476  # 19.915 MJ/kg LHV -> 19915000MJ/kt -> 0.476ktoe/kt
+KTOE_TO_TWH = 0.01163
+KT_TO_T = 1000
+
 CAT_NAME_CHEMICALS = "Chemicals Industry"
 
 
 def get_chemicals_demand_df(
-    config: dict,
+    # config: dict,
     energy_balances: str,
     cat_names: str,
     carrier_names: str,
@@ -43,16 +49,21 @@ def get_chemicals_demand_df(
 
     # Process data
     new_chemicals_demand = transform_jrc_chemicals_subsector_demand(
-        jrc_energy, jrc_prod, config
+        jrc_energy,
+        jrc_prod,
+        # config
     )
     new_chemicals_demand = filling.fill_missing_countries_years(
         energy_balances_df, cat_names_df, carrier_names_df, new_chemicals_demand
     )
+    breakpoint()
     new_chemicals_demand.to_netcdf(output_path)
 
 
 def transform_jrc_chemicals_subsector_demand(
-    jrc_energy: xr.Dataset, jrc_prod: xr.Dataset, config_steel: dict
+    jrc_energy: xr.Dataset,
+    jrc_prod: xr.Dataset,
+    # config_chemicals: dict
 ) -> xr.Dataset:
     """Processing of chemicals industry energy demand for different carriers.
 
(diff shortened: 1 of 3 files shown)


tud-mchen6:
My high-level questions, just to confirm:

1. Do we only consider **energy** demand? During the process of extracting electricity demand (and assuming all demands that could possibly be met by electricity are met by electricity), some demands that cannot be met by electricity will be filtered out.
2. Do we assume that all low enthalpy heat demand cannot be met by electricity?
3. Regarding the "energy demand" of Basic chemicals: what carriers are they?
4. On the one hand, we estimate electricity demand from `jrc_energy`, while on the other hand, we estimate Basic chemicals related demand from `jrc_prod`. Will there be overlap or double-counting?

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        jrc_energy_df (xr.Dataset): jrc country-specific steel energy demand.
+        jrc_prod_df (xr.Dataset): jrc country-specific steel production.
+        config_chemicals (dict): configuration for the chemicals industry.
+
+    Returns:
+        xr.Dataset: processed dataframe with the expected steel energy consumption.
+    """
+    # Get the electricity consumption (assuming every demand that could be met by electricity is met by electricity)
+    electrical_consumption = (
+        (jrc.replace_carrier_final_demand("Electricity", jrc_energy))
```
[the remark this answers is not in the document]
irm-codebase: The best way to answer this is probably looking at what the old code does for both chemicals and steel here, not "other" industry, which is unique.

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -144,26 +144,36 @@ def replace_carrier_final_demand(
     Returns:
         xr.DataArray: final demand data met using the specified carrier.
     """
-    useful_dem_tot = jrc_energy["useful"].sum("carrier_name")
+    useful_dem_tot = jrc_energy["useful"].sum(
+        "carrier_name"
+    )  # MC note: including all carriers
```
irm-codebase, line 149: Please remove these comments.
If the function is not descriptive enough, improve the docstring.

Easier to maintain down the line (and avoids `black` creating extra lines because of the comment).

modules/industry/scripts/utils/jrc_idees_parser.py
```diff
@@ -144,26 +144,36 @@ def replace_carrier_final_demand(
...
 
     # Prettify
     carrier_final_demand = carrier_final_demand.drop(["carrier_name"])
     carrier_final_demand = carrier_final_demand.assign_attrs(units="twh")
     carrier_final_demand.name = "final"
-
-    assert useful_dem_tot.sum() < carrier_final_demand.sum(), "Creating energy!"
-
+    # assert useful_dem_tot.sum() < carrier_final_demand.sum(), "Creating energy!" # MC note: but useful_dem_tot includes all carrier demands while carrier_final_demand only contains the demand related to one specific carrier. Surely this is fair comparison?
+    # MC note: commented out because this is not correct for chemicals industry. but need to coordinate with Ivan regarding how this function should be written.
```
irm-codebase, line 176: See my comment in the "other industry" PR.

irm-codebase (changes requested):
Nice job! However, I have several  suggestions.

Biggest one is that I feel like you are not using the correct JRC function for the electricity consumption stuff. See the specific comment. Feel free to comment if you think I got something wrong!

irm-codebase:
> My high-level questions, just to confirm:
> 
>     1. Do we only consider **energy** demand? During the process of extracting electricity demand (and assuming all demands that could possibly be met by electricity are met by electricity), some demands that cannot be met by electricity will be filtered out.
> 
>     2. Do we assume that all low enthalpy heat demand cannot be met by electricity?
> 
>     3. Regarding the "energy demand" of Basic chemicals: what carriers are they?
> 
>     4. On the one hand, we estimate electricity demand from `jrc_energy`, while on the other hand, we estimate Basic chemicals related demand from `jrc_prod`. Will there be overlap or double-counting?

I'll try to help here, but I am unfamiliar with the chemicals section of the code. I can only help with the jrc helper functions and so on. If there are more questions, I can help when I'm back from my vacation.

1. For chemicals, no. Any "energy relevant" carrier matters here. I do not remember the SCEC specifics, but these could be: ammonia, H2, CO2, etc. Please check the SCEC code and include all the ones used there.
2. Low enthalpy heat can be met by whatever technology the model chooses (heatpumps, boilers, wood, etc). Just use the same method as the iron and steel sector.
3. Check the SCEC code please. That is probably there, or in the configuration files.
4. Please see the SCEC code, or ask @brynpickering if you suspect double counting is occurring. My guess is "no", if you use the correct subsections.

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
+import pandas as pd
+import xarray as xr
+from utils import filling
+from utils import jrc_idees_parser as jrc
+
+H2_LHV_KTOE = 2.863  # 0.0333 TWh/kt LHV -> 2.863ktoe/kt
```
brynpickering, line 6: Some of these values are also in the steel script. To avoid the maintenance burden, perhaps we move these to a single file (e.g. `utils/constants.py`?)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+) -> xr.Dataset:
+    """Processing of chemicals industry energy demand for different carriers.
+
+    Calculates energy consumption in the iron and steel industry based on expected
+    change in processes to avoid fossil feedstocks. All process specific energy consumption
+    (energy/t_steel) is based on the Electric Arc process (EAF), except sintering, which
+    will be required for iron ore processed using H-DRI, but is not required by EAF.
+
+    This function does the following:
+    1.
```
brynpickering, line 75: Fill?

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+    }
+    co2_demand = {  # tCO2/t
+        "Ethylene": 0,
+        "Propylene": 0,
+        "BTX": 0,
+        "Ammonia": 0.112,  # inc. 0.35t_urea/t_ammonia
+        "Methanol": 1.373,
+    }
+
+    energy_demand = {  # MWh/t
```
brynpickering, line 117: Move to config (there are user assumptions here)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+    # Get the electricity consumption (assuming every demand that could be met by electricity is met by electricity)
+    electrical_consumption = (
+        (jrc.replace_carrier_final_demand("Electricity", jrc_energy))
+        .drop_sel(subsection="Low enthalpy heat")
+        .sum(dim=["section", "subsection"])
+        .drop_vars("cat_name")
+    )
+
+    # Gather relevant params for Basic chemicals in the Chemicals Industry
+    h2_demand = {  # t/t
```
brynpickering, line 95: move to top of file (they are constants)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        "Ethylene": 0,
+        "Propylene": 0,
+        "BTX": 0,
+        "Ammonia": 0.178,
+        "Methanol": 0.189,
+    }
+    methanol_demand = {
```
brynpickering, line 102: move to top of file (they are constants)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        "Ethylene": 2.83,
+        "Propylene": 2.83,
+        "BTX": 4.3,
+        "Ammonia": 0,
+        "Methanol": 1,
+    }
+    co2_demand = {  # tCO2/t
```
brynpickering, line 109: move to top of file (they are constants)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    mass = {  # Bazzanella and Ausfelder, 2017
+        "Ethylene": 21.7,
+        "Propylene": 17,
+        "BTX": 15.7,
+        "Ammonia": 17,
+        "Methanol": 2,
+    }
+    molar_mass = {
```
brynpickering, line 132: move to top of file (they are constants)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+
+    energy_demand = {  # MWh/t
+        "Ethylene": 1.4,  # MTO process
+        "Propylene": 1.4,  # MTO process
+        "BTX": 1.4,  # MTA process
+        "Ammonia": 2.05,  # inc. 0.35t_urea/t_ammonia
+        "Methanol": 1.5,  # auxiliary demand, could just be assumed as already included in existing electricity demand
+    }
+
+    mass = {  # Bazzanella and Ausfelder, 2017
```
brynpickering, line 125: move to top of file (they are constants)

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        "Methanol": 2,
+    }
+    molar_mass = {
+        "Ethylene": 28.05,
+        "Propylene": 42.08,
+        "BTX": 93,
+        "Ammonia": 17.01,
+        "Methanol": 32.04,
+    }
+    moles = {k: mass[k] / molar_mass[k] for k in mass}
```
brynpickering, line 139: I would separate the mass-to-molar mass calc into its own function

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        jrc_prod,
+        # config
+    )
+    new_chemicals_demand = filling.fill_missing_countries_years(
+        energy_balances_df, cat_names_df, carrier_names_df, new_chemicals_demand
+    )
+    new_chemicals_demand.to_netcdf(output_path)
+
+
+def transform_jrc_chemicals_subsector_demand(
```
brynpickering, line 62: I don't think we're using "subsector" any more. See the steel script

modules/industry/scripts/chemicals_industry.py
brynpickering: Actually, move to config: these contain user assumptions

modules/industry/scripts/chemicals_industry.py
```diff
@@ -0,0 +1,215 @@
...
+        * KTOE_TO_TWH
+    )
+    chem_methanol_consumption = chem_methanol_consumption.assign_attrs(
+        units="twh"
+    ).assign_coords(carrier_name="methanol")
+    chem_co2_consumption = sum(masses[k] * co2_demand[k] for k in co2_demand) * KT_TO_T
+    chem_co2_consumption = chem_co2_consumption.assign_attrs(units="t").assign_coords(
+        carrier_name="co2"
+    )
+    # TODO: find out how co2 is called in jrc
```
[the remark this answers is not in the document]
brynpickering: Yeah, this is why we provide CO2 demand as constants in a dictionary earlier in the script.

brynpickering (changes requested):
THanks for this @tud-mchen6! I would restructure the chemicals industry script a bit to move constants around and to distribute it into smaller functions that you can then give nice docstrings for (for future maintainability, e.g. so others don't wonder the same things you have wondered when making this contribution!)

brynpickering:
In answer to your questions:

> Do we only consider energy demand? During the process of extracting electricity demand (and assuming all demands that could possibly be met by electricity are met by electricity), some demands that cannot be met by electricity will be filtered out.

We completely overhaul the chemical industry stuff and as a result only keep direct electricity consumption, which refers to all energy requirements to process chemicals. I believe that the non-electricity consumption in JRC-IDEES refers to _creating_ relevant chemicals, which we completely replace with our own demands for CO2, hydrogen, and auxiliary electricity. **This probably requires some review though.**

> Do we assume that all low enthalpy heat demand cannot be met by electricity?

It gets turned into "space heat demand" later, which can be met by any available technologies for that.

> Regarding the "energy demand" of Basic chemicals: what carriers are they?

Electricity.

> On the one hand, we estimate electricity demand from jrc_energy, while on the other hand, we estimate Basic chemicals related demand from jrc_prod. Will there be overlap or double-counting?

The electricity demand from jrc_energy refers to demand to process basic chemicals. The electricity demand we get from processing jrc_prod is for creating those basic chemicals. They _should_ be mutually exclusive sets of demands that can be added together without double counting.

Issue #39, opened 2020-07-21
Add heat supply technologies


brynpickering:
With heat demand (#16) comes the need for heating technologies. These include:

[ ] Heat pumps with temperature dependent performance
[ ] Biofuel boilers
[ ] Conventional boilers for use with synthetic fuels and/or Hydrogen
[ ] Combined heat and power plants (inc. waste incinerators?)
[ ] Rooftop solar thermal panels

brynpickering:
This has been implemented as a Calliope model YAML template in this repositories [sector-coupled branch](https://github.com/calliope-project/euro-calliope/blob/sector-coupled/src/template/heat-techs.yaml)

brynpickering:
Heat pump timeseries data would be dealt with via #80

PR #419, opened 2024-07-25
Separate biofuel supply; add biofuel boiler

### LINKED ISSUE #417: Separate biofuel as a unique energy carrier
### What can be improved?

Euro-Calliope has a dispatchable technology that takes in biofuel and produces electricity, but it is a `supply` technology. To allow for a decision on _how_ to use limited biofuel resources we need a `biofuel` energy carrier. This would allow us to have biofuel boilers as a heat supply technology (#39)

### Version

1.0.0

brynpickering:
Fixes #417 
Adds partial fix to #39 

## Checklist

- [x] CHANGELOG updated
- [x] Documentation updated


docs/model/customisation.md
```diff
@@ -102,9 +104,27 @@ Here, we describe each module in terms of the technologies they contain (`callio
...
+
+        **biofuel_flow_limits**: Distribute annual total production limit evenly across all hours of the year and allow biofuel storage up to 50% of total annual production.
+
+??? note "conversion/heat-from-biofuel.yaml"
+
+    === "Technologies"
+
+        **biofuel_boiler**: Biofuel-consuming boiler.
+
+        **biofuel_tech_heat_to_demand**: Dummy technology to convert biofuel boiler output to a carrier that can be used to meet heat demand.
```
timtroendle, line 119: This is a unclear to me. Why is this a dummy technology?

templates/models/techs/conversion/electricity-from-biofuel.yaml.jinja
```diff
@@ -0,0 +1,19 @@
+techs:
+    electricity_from_biofuel:  # from [@JRC:2014] Table 48 Anaerobic digestion
+        essentials:
+            name: Electricity from anaerobically digested biofuel
+            parent: conversion
+            carrier_in: biofuel
```
timtroendle, line 6: This seems to remove the old approach, meaning that it won't be possible anymore to avoid modelling biofuel explicitly. I am fine with as long as it doesn't impact performance much. Do you know?

timtroendle (approved):
Looks good to me!

My only concern is the removal of the existing mechanism. This leads to the addition of at least one more carrier in all models that use biofuel. Do we know about the runtime and numerical scaling implications?

templates/models/techs/supply/biofuel.yaml.jinja
```diff
@@ -1,23 +1,34 @@
...
+    {{ id }}.techs.biofuel_supply:
     {% endfor %}
+
+group_constraints:
+    {% for id, location in locations.iterrows() %}
+    biofuel_max_prod_{{ id }}:
+        locs: [{{ id }}]
+        techs: [biofuel_supply]
+        carrier_prod_max:
+            biofuel: {{ location.biofuel_potential_mwh_per_year * scaling_factors.power }} # {{ (1 / scaling_factors.power) | unit("MWh") }}
```
timtroendle, line 23: This line may be negative for numerical scaling as it introduces a fairly large number. Previously this was solved by applying the efficiency and assuming a 6 month use period, which reduced the number by ~factor 4.

docs/model/customisation.md
```diff
@@ -15,13 +15,13 @@ You have the following three options:
 With the Calliope model in your hands, you will be able to change any model parameter, any technology specifics, and the model definition to your liking.
 This kind of customisation can be useful to get to know the model and its parameters.
 To create reliable results, we advise making manual changes only to the model definition (`example-model.yaml`) as this makes it possible to trace those changes later.
-A typical customisation here would be to change the solver from `gurobi` to an open-source solver, e.g. `cbc` (see [Calliope's documentation](https://calliope.readthedocs.io/en/v{{ calliope_version }}/user/config_defaults.html#run-configuration)).
+A typical customisation here would be to change the solver from `gurobi` to an open-source solver, e.g. `cbc` (see [Calliope's documentation](<https://calliope.readthedocs.io/en/v{{> calliope_version }}/user/config_defaults.html#run-configuration)).
```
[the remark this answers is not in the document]
brynpickering: hmm, that looks like it snuck in for some reason. It certainly doesn't look like it makes any sense to me.

docs/model/customisation.md
brynpickering: Because it doesn't actually represent a "real" technology. It's just a modelling artefact to be able to have storage buffers per heat tech (this will be a non-issue when using calliope v0.7)

templates/models/techs/supply/biofuel.yaml.jinja
brynpickering: Differences in ranges:
## Before
```
[2024-07-30 17:51:46] DEBUG    Matrix range     [4e-05, 1e+02]
[2024-07-30 17:51:46] DEBUG    Objective range  [1e+00, 1e+00]
[2024-07-30 17:51:46] DEBUG    Bounds range     [1e-05, 2e+01]
[2024-07-30 17:51:46] DEBUG    RHS range        [8e-10, 2e+02]
```

## After
```
[2024-07-30 18:04:13] DEBUG    Matrix range     [4e-05, 1e+02]
[2024-07-30 18:04:13] DEBUG    Objective range  [1e+00, 1e+00]
[2024-07-30 18:04:13] DEBUG    Bounds range     [1e-05, 2e+01]
[2024-07-30 18:04:13] DEBUG    RHS range        [8e-10, 1e+03]
```

So, yes, the RHS range increases by an order of magnitude.

templates/models/techs/conversion/electricity-from-biofuel.yaml.jinja
brynpickering: I don't know about memory use, but runtime for building and solving the optimisation problem increases by ~3% in a test I made (UK+IRL, 1yr @ 2hr resolution, using `%timeit` magic).

brynpickering:
@timtroendle if we keep the existing approach as an option, how would you see it being included as a model file? It could be kept as its own module alongside `biofuel.yaml`, but then it needs a name that clearly distinguishes it (`supply/electricity-from-biofuel.yaml`, as a distinction from `conversion/electricity-from-biofuel.yaml`? Or `supply/bio-electricity.yaml` or similar?)

templates/models/techs/supply/biofuel.yaml.jinja
brynpickering: Not sure there is an equivalent way to handle that with biofuel as a separate carrier. I guess you just don't _need_ `carrier_prod_max` if you have an hourly `resource` since that resource already effectively sets an annual value once it's all been added up.

brynpickering: Yeah, so without the annual limit, it becomes:

```
[2024-07-30 18:25:14] DEBUG    Matrix range     [4e-05, 1e+02]
[2024-07-30 18:25:14] DEBUG    Objective range  [1e+00, 1e+00]
[2024-07-30 18:25:14] DEBUG    Bounds range     [1e-05, 2e+01]
[2024-07-30 18:25:14] DEBUG    RHS range        [8e-10, 5e+02]
```

templates/models/techs/conversion/electricity-from-biofuel.yaml.jinja
brynpickering: actually, ran it again and it was the other way around. So, within margins of error, there is no change 😅

templates/models/techs/supply/biofuel.yaml.jinja
timtroendle: But that would mean that biofuel must be burned whenever it's becoming available as a `resource`, right? That wouldn't be ideal either.

timtroendle:
I am not sure whether we _should_ keep it, but if we do a name like `supply/electricity-only-biofuel.smk` may work?

timtroendle:
I mean the numerical scaling is an issue, but then again, it didn't seem to affect your runs, right? So maybe we can get around it without keeping the original approach.

templates/models/techs/supply/biofuel.yaml.jinja
brynpickering: No, the resource per timestep just gives a quantity _available_. So long as `force_resource` isn't set to `True` it won't be forced to make use of it. It can also choose to make use of it but to store it instead of having it converted to electricity etc.

brynpickering:
@timtroendle I've reorganised the techs slightly so that we have an `electrified-biofuel.yaml` module that can be used in place of `biofuel.yaml` and `electricity-from-biofuel.yaml`. They are drop in replacements for each other, so the objective function value is the same using either, but the latter would be used when introducing other uses for biofuel (e.g. heating).

timtroendle (approved):

PR #434, opened 2024-09-21
Update zenodo link to v1.1 of pre-built model

brynpickering:
Fixes version of pre-built Euro-Calliope being referenced in the docs/README. Currently users are directed to v1.0 and have to notice the Zenodo banner RE a newer version being available.

This is a patch on main and we should probably then move the v1.1 tag to this commit so that the stable version of the docs updates accordingly.

## Checklist

- [ ] CHANGELOG updated
- [ ] Documentation updated

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -3,9 +3,11 @@
 # See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
 
 version: 2
-
+build:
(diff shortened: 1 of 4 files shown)


brynpickering:
Since this isn't related to the workflow I think it's OK to ignore the fact that the tests fail (due to dependency issues).

